PR #3, opened 2022-09-21
Flyway changes

jennyglassmyer:
Adjusts flyway migrations stopping build of the server


talentbeyondboundaries (commented):
The only difference here is that a blank line has been removed.

We don't need to change V1_168

talentbeyondboundaries (commented):
I don't think we want the status to be null. The proper change for this is when we add a user - particularly the default SystemAdmin user we need to set status.

I have updated SystemAdminConfiguration to set the status of the auto created SystemAdminUser, so I think we can leave status being defined as not null.

That is the proper fix - not to allow status to be null.

talentbeyondboundaries (commented):
I have also removed all the automatically created users. 

They are not needed.

talentbeyondboundaries (commented):
I can't remember why created_by needed to be null so I am leaving it as not null.
It owuld be better to fix any code that does not supply a created_by

talentbeyondboundaries (commented):
Rather than comment these out, it is better to delete them.

We can always see previous values if we ever need to through version control.

talentbeyondboundaries:
I have applied all these changes to the latest code. The code that you are working from is quite old.

So nothing for you to do.

talentbeyondboundaries (commented):
Shoudl be able to V1_210 in - I don't see why it should fail.

And partner abbreviations should not be null.

camerojo:
Closing this pull request. I have manually updated selected Flyway changes as indicated in above comments

PR #4, opened 2022-09-21
Dependency version & SystemAdminConfiguration changes

jennyglassmyer:
Updates version of gradle wrapper, lombok, node, jasmine-core, and karma to be compatible with M1 macs. Also changes status of default SystemAdmin user created on startup in dev to active and removed MFA requirement for this user to streamline dev process.


camerojo:
As discussed I have done a complete update of all versions as far as we can go. Hopefully that will address some of the issues you encountered.
So this pull request is no longer needed.

PR #6, opened 2022-09-29
Fix DB migration script errors found whilst starting from an empty DB

bplump:
These changes address some errors I encountered when trying to start the app with an empty database.

### DIFF
--- a/server/src/main/resources/db/migration/V1_167__add_more_standard_tasks.sql
+++ b/server/src/main/resources/db/migration/V1_167__add_more_standard_tasks.sql
@@ -53,7 +53,7 @@ values ('Residence Attestation','UploadTask',
 insert into task (name, task_type, description, days_to_complete, upload_subfolder_name,
                   upload_type, uploadable_file_types,
                   created_by, created_date)
-values ('Residence Attestation','UploadTask', 'Translation of original document',
+values ('Residence Attestation Translation','UploadTask', 'Translation of original document',
         14, 'Address', 'residenceAttestTrans', null,
         (select id from users where username = 'SystemAdmin'), now());
 
--- a/server/src/main/resources/db/migration/V1_209__add_partners_abbreviations.sql
+++ b/server/src/main/resources/db/migration/V1_209__add_partners_abbreviations.sql
@@ -14,10 +14,10 @@
  * along with this program. If not, see https://www.gnu.org/licenses/.
  */
 
-update partner set abbreviation = 'IOM' where 'name' = 'IOM';
-update partner set abbreviation = 'HIAS' where 'name' = 'HIAS';
-update partner set abbreviation = 'UNHCR' where 'name' = 'UNHCR';
-update partner set abbreviation = 'RP' where 'name' = 'Refuge Point';
-update partner set abbreviation = 'Dignity' where 'name' = 'Dignity for Children';
-update partner set abbreviation = 'CRS' where 'name' = 'Catholic Relief Services';
-update partner set abbreviation = 'TBB' where 'name' = 'Talent Beyond Boundaries';
+update partner set abbreviation = 'IOM' where name = 'IOM';
+update partner set abbreviation = 'HIAS' where name = 'HIAS';
+update partner set abbreviation = 'UNHCR' where name = 'UNHCR';
+update partner set abbreviation = 'RP' where name = 'Refuge Point';
+update partner set abbreviation = 'Dignity' where name = 'Dignity for Children';
+update partner set abbreviation = 'CRS' where name = 'Catholic Relief Services';
+update partner set abbreviation = 'TBB' where name = 'Talent Beyond Boundaries';
--- a/server/src/main/resources/db/migration/V1__init.sql
+++ b/server/src/main/resources/db/migration/V1__init.sql
@@ -201,3 +201,7 @@ comment                 text,
 created_by              bigint references users,
 created_date            timestamptz
 );
+
+-- Create the default system admin user which is required by later migration scripts
+-- Both the id (1) and username (SystemAdmin) are referenced
+insert into users (id, username, first_name, last_name, email, role, status) values (1, 'SystemAdmin', 'System', 'Admin', 'systemadmin@example.com', 'systemadmin', 'active');


server/src/main/resources/db/migration/V1_167__add_more_standard_tasks.sql
```diff
@@ -53,7 +53,7 @@ values ('Residence Attestation','UploadTask',
 insert into task (name, task_type, description, days_to_complete, upload_subfolder_name,
                   upload_type, uploadable_file_types,
                   created_by, created_date)
-values ('Residence Attestation','UploadTask', 'Translation of original document',
+values ('Residence Attestation Translation','UploadTask', 'Translation of original document',
```
bplump, line 56: Migration `V1_184__task_display_name.sql` fails because it's adding a unique constraint on the `name` column but this task has the same name as the one above.

server/src/main/resources/db/migration/V1_209__add_partners_abbreviations.sql
```diff
@@ -14,10 +14,10 @@
...
-update partner set abbreviation = 'Dignity' where 'name' = 'Dignity for Children';
-update partner set abbreviation = 'CRS' where 'name' = 'Catholic Relief Services';
-update partner set abbreviation = 'TBB' where 'name' = 'Talent Beyond Boundaries';
+update partner set abbreviation = 'IOM' where name = 'IOM';
+update partner set abbreviation = 'HIAS' where name = 'HIAS';
+update partner set abbreviation = 'UNHCR' where name = 'UNHCR';
+update partner set abbreviation = 'RP' where name = 'Refuge Point';
+update partner set abbreviation = 'Dignity' where name = 'Dignity for Children';
+update partner set abbreviation = 'CRS' where name = 'Catholic Relief Services';
+update partner set abbreviation = 'TBB' where name = 'Talent Beyond Boundaries';
```
bplump, line 23: These updates didn't actually set the corresponding abbreviation for me because Postgres says no row matches. (I'm not sure how/why Postgres accepts this as valid SQL if it thinks the quoted column name doesn't exist but the fix was easy enough... I'm using Postgres 14, so there's a chance my version of Postgres is less-tolerant than the version that originally ran the migration)

Without this fix, migration `V1_210__set_partners_abbreviations_not_null.sql` fails because it's trying to make the `abbreviation` column NON NULL and the rows corresponding to these statements don't have abbreviations.

server/src/main/resources/db/migration/V1__init.sql
```diff
@@ -201,3 +201,7 @@ comment                 text,
 created_by              bigint references users,
 created_date            timestamptz
 );
+
+-- Create the default system admin user which is required by later migration scripts
+-- Both the id (1) and username (SystemAdmin) are referenced
+insert into users (id, username, first_name, last_name, email, role, status) values (1, 'SystemAdmin', 'System', 'Admin', 'systemadmin@example.com', 'systemadmin', 'active');
```
bplump, line 207: I'm unsure about this change because I've not actually spun up the application to see how it works with a user with no password...

But this is required for (at least) two migrations:
* `V1_6__translate.sql` references a user with ID=1
* `V1_166__add_standard_tasks.sql` references a user with name='SystemAdmin'

bplump: Hmm... I see there's a `SystemAdminConfiguration` to create the sysadmin user (which seems preferable).

An alternative to this is to create a bootstrap user here (id=0?) and update the migrations to reference that user? Any preferences?

camerojo: Can we make this a separate pull request? 

I prefer the bootstrap user (which was added after the original code was written). That is what SystemAdminConfiguration is supposed to do. Then code can assume the existence of a user called SystemAdmin - which I think is preferable to assuming a particular id. So V1_6 needs to change to work like V1_166

server/src/main/resources/db/migration/V1_209__add_partners_abbreviations.sql
camerojo: This looks good. Nasty that this got through. Thanks for the pick up.

server/src/main/resources/db/migration/V1_167__add_more_standard_tasks.sql
camerojo: Good pick up - many thanks

camerojo (changes requested):
I have never approved a pull request before - so forgive me!

The first two changes are fine and should be applied.

But, as I mentioned in my comment, instead of the last change I would prefer to make the bootstrap user approach work properly.

server/src/main/resources/db/migration/V1__init.sql
bplump: Yeah, I'm not very happy with what I've done here but I'm also a bit stuck...

The initial DB scripts are run during bean creation by Spring's `FlywayMigrationInitializer.afterPropertiesSet`. `SystemAdminConfiguration.autoCreateSystemAdmin` runs after that... so the `SystemAdmin` user won't get created until after all the scripts are run.

Unfortunately, a number of the scripts want a user in the system so they can set `created_by` columns. Take a look at [the two insertions in `V1_6__translate.sql`](https://github.com/talentbeyondboundaries/tbbtalentv2/blob/master/server/src/main/resources/db/migration/V1_6__translate.sql#L42-L43). I tried leaving `created_by` as `NULL`, but I'm not sure how that might impact the application. And it only works for the `system_language` table because there's no constraint on the column. 

There's no way around it for `V1_166__add_standard_tasks.sql` because [`V1_164__create_task_and_taskassignment_tables.sql` specifies the  `created_by` column is `not null`](https://github.com/talentbeyondboundaries/tbbtalentv2/blob/master/server/src/main/resources/db/migration/V1_164__create_task_and_taskassignment_tables.sql#L21). 

Creating a dummy/bootstrap user is the only way I could get around all the problems above. 

What do new developers on the application do now? Do they have a DB dump to work from?

camerojo: Ahhhh - yes. Sorry about this. Yes - to date - new developers (really only one) have worked with a DB dump. But I do want to get this fixed.

I am assuming that you are Brent from Atlassian - Lauren Button mentioned your name. I can give you a DB dump but would also very much welcome the opportunity to demo the software to you. As you can see we are very crunched for resources. I used to work on this with my daughter - but she is taking maternity leave - so it is just me for now and I am a bit stretched.

bplump: Yup, that's me! I'd love to see the application running in IRL whenever you've got some time. I've only imagined how it works based on the a bunch of static screenshots that Lauren showed me.

I'm pretty impressed with what you've built! Hopefully, I can get it up and running soon on my machine. 

After I made these DB changes, I got stuck with the requirement for Google Drive. Whilst I waited for Atlassian to approve an account, I wrote a hacky little local filesystem stub. Then I got stuck with SalesForce... writing a stub for that was beyond me (and I ran out of time).

camerojo: IRL? But delighted to demo it. I suspect that your time is more structured than mine - so give me a couple of times that suit you and I will probably be able to accommodate

camerojo: I just pushed a release that hopefully helps. The V1__init Flyway has been updated (and repaired) to insert the SystemAdmin user into the database (as user id = 1 hopefully). You should be able to log in to it with password "password".

camerojo: As for connections to GoogleDrive and Salesforce, we have test connections for those. There are some test Google drives on the TBB account. The Salesforce connection is to the prod Salesforce (not great, I know!) where we use some test data on prod to play with.

The config for connecting to the above are in a secret file as described in the extract from the Readme copied below.

I could send you a copy of tbb_secrets.txt in a DM on Slack - or any other way you like.

### Run the server ###

- Some secret information such as passwords and private keys are set in 
  environment variables - including programmatic access to TBB's Amazon AWS, 
  Google and Salesforce accounts. If these environment variables are not set
  the application should still run in your development environment, but it may
  not have access to these integrations. Contact TBB if you need access to these
  "secrets". They are stored in a tbb_secrets.txt file which you can hook into
  your start up to set the relevant environment variables. 
  For example add "source ~/tbb_secrets.txt" to .bash_profile or .zshenv
  depending on whether you are running bash or zsh.

camerojo:
Closing this - we have addressed these issues in other commits.

PR #18, opened 2023-05-15
Unit tests for AuthAdminApi

sadatmalik:
- unit tests for AuthAdminApi
- minor code refactoring and some java doc


server/src/main/java/org/tbbtalent/server/exception/PasswordExpiredException.java
```diff
@@ -18,8 +18,14 @@
```
sadatmalik, line 18: Would be great to have unit tests for these!

sadatmalik: Discussed with JC - we'll add as we go along

sadatmalik:
Code walk through with JC - approved to merge into staging.

PR #28, opened 2023-05-18
unit tests for branding admin api

### LINKED ISSUE #19: Write unit tests for branding admin api

sadatmalik:
- BrandingInfo - unit tests and refactor to use @Builder annotation
- BrandingAdminApi - unit tests and refactor away some boilerplate code
- BrandingServiceImpl - unit tests, some java-doc and minor method refactoring

### DIFF
--- a/server/src/main/java/org/tbbtalent/server/api/admin/BrandingAdminApi.java
+++ b/server/src/main/java/org/tbbtalent/server/api/admin/BrandingAdminApi.java
@@ -18,12 +18,14 @@ package org.tbbtalent.server.api.admin;
 
 import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
-import org.springframework.http.ResponseEntity;
 import org.springframework.web.bind.annotation.GetMapping;
 import org.springframework.web.bind.annotation.RequestMapping;
 import org.springframework.web.bind.annotation.RestController;
 import org.tbbtalent.server.model.db.BrandingInfo;
 import org.tbbtalent.server.service.db.BrandingService;
+import org.tbbtalent.server.util.dto.DtoBuilder;
+
+import java.util.Map;
 
 /**
  * Access to branding information for the admin portal.
@@ -41,9 +43,18 @@ public class BrandingAdminApi {
      * @return branding information
      */
     @GetMapping()
-    public ResponseEntity<BrandingInfo> getBrandingInfo() {
+    public Map<String, Object> getBrandingInfo() {
+
         BrandingInfo info = brandingService.getBrandingInfo(null);
-        return ResponseEntity.ok(info);
+        return brandingInfoDto().build(info);
+    }
+
+    DtoBuilder brandingInfoDto() {
+        return new DtoBuilder()
+                .add("logo")
+                .add("partnerName")
+                .add("websiteUrl")
+                ;
     }
 
 }
--- a/server/src/main/java/org/tbbtalent/server/api/admin/BrandingAdminApi.java
+++ b/server/src/main/java/org/tbbtalent/server/api/admin/BrandingAdminApi.java
@@ -16,48 +16,34 @@
 
 package org.tbbtalent.server.api.admin;
 
-import java.util.Map;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-import org.springframework.beans.factory.annotation.Autowired;
+import lombok.RequiredArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
+import org.springframework.http.ResponseEntity;
 import org.springframework.web.bind.annotation.GetMapping;
 import org.springframework.web.bind.annotation.RequestMapping;
 import org.springframework.web.bind.annotation.RestController;
 import org.tbbtalent.server.model.db.BrandingInfo;
 import org.tbbtalent.server.service.db.BrandingService;
-import org.tbbtalent.server.util.dto.DtoBuilder;
 
 /**
  * Access to branding information for the admin portal.
  */
 @RestController()
 @RequestMapping("/api/admin/branding")
+@Slf4j
+@RequiredArgsConstructor
 public class BrandingAdminApi {
-    private static final Logger log = LoggerFactory.getLogger(BrandingAdminApi.class);
 
     private final BrandingService brandingService;
-
-    @Autowired
-    public BrandingAdminApi(BrandingService brandingService) {
-        this.brandingService = brandingService;
-    }
 
     /**
      * Retrieve the branding information for the admin portal.
      * @return branding information
      */
     @GetMapping()
-    public Map<String, Object> getBrandingInfo() {
-
+    public ResponseEntity<BrandingInfo> getBrandingInfo() {
         BrandingInfo info = brandingService.getBrandingInfo(null);
-        return brandingInfoDto().build(info);
+        return ResponseEntity.ok(info);
     }
 
-    DtoBuilder brandingInfoDto() {
-        return new DtoBuilder()
-                .add("logo")
-                .add("partnerName")
-                .add("websiteUrl")
-                ;
-    }
 }
--- a/server/src/main/java/org/tbbtalent/server/api/admin/BrandingAdminApi.java
+++ b/server/src/main/java/org/tbbtalent/server/api/admin/BrandingAdminApi.java
@@ -38,10 +38,6 @@ public class BrandingAdminApi {
 
     private final BrandingService brandingService;
 
-    /**
-     * Retrieve the branding information for the admin portal.
-     * @return branding information
-     */
     @GetMapping()
     public Map<String, Object> getBrandingInfo() {
 
--- a/server/src/main/java/org/tbbtalent/server/model/db/BrandingInfo.java
+++ b/server/src/main/java/org/tbbtalent/server/model/db/BrandingInfo.java
@@ -16,9 +16,7 @@
 
 package org.tbbtalent.server.model.db;
 
-import lombok.Getter;
-import lombok.Setter;
-import lombok.ToString;
+import lombok.*;
 import org.springframework.lang.Nullable;
 
 /**
@@ -29,6 +27,7 @@ import org.springframework.lang.Nullable;
 @Getter
 @Setter
 @ToString
+@Builder
 public class BrandingInfo {
 
     /**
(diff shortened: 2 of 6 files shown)


server/src/main/java/org/tbbtalent/server/api/admin/BrandingAdminApi.java
```diff
@@ -16,48 +16,34 @@
...
-    public BrandingAdminApi(BrandingService brandingService) {
-        this.brandingService = brandingService;
-    }
-
     /**
      * Retrieve the branding information for the admin portal.
      * @return branding information
      */
     @GetMapping()
-    public Map<String, Object> getBrandingInfo() {
```
camerojo, line 50: This method of returning  results from API calls is all through the code and would be a big job to replace. 
We should talk about the pros and cons.

The advantage of the Map<String, Object> and DtoBuilder approach is that we can define subsets of server objects that are passed back to the browser. Whether or not this was a good design approach - it is what we currently have and we would need to think seriously about changing it. There are definitely places in the code which depend on the flexibility of DtoBuilder to work.

server/src/test/java/org/tbbtalent/server/api/admin/BrandingAdminApiTest.java
```diff
@@ -0,0 +1,86 @@
...
+import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.content;
+import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.jsonPath;
+import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.status;
+
+/**
+ * @author sadatmalik
+ */
+@WebMvcTest(BrandingAdminApi.class)
+@AutoConfigureMockMvc
+class BrandingAdminApiTest extends ApiTestBase {
```
camerojo, line 46: Where does ApiTestBase come from and what does it do? I can't find it in any import.

server/src/main/java/org/tbbtalent/server/model/db/BrandingInfo.java
```diff
@@ -29,6 +27,7 @@
 @Getter
 @Setter
 @ToString
+@Builder
```
camerojo: Lets chat about the pros and cons of the Builder pattern.

For example, https://softwareengineering.stackexchange.com/questions/419403/is-it-a-good-or-a-bad-idea-to-use-the-builder-pattern-everywhere

Also is it worth the effort moving everything across to it - given that the existing code doesn't use the pattern extensively.

server/src/main/java/org/tbbtalent/server/service/db/impl/BrandingServiceImpl.java
```diff
@@ -34,47 +36,51 @@
...
+     * If no partner is found gets the default source partner.
+     *
+     * @param partnerAbbreviation Optional partner abbreviation
+     * @return branding info
+     */
     @Override
     @NonNull
     public BrandingInfo getBrandingInfo(@Nullable String partnerAbbreviation) {
 
         User user = userService.getLoggedInUser();
```
camerojo: Why is this better? I think it is a matter of personal preference. We lost some comments in the change here - highlighting for example that the different processing related to whether or not a user is logged in.

server/src/main/java/org/tbbtalent/server/api/admin/BrandingAdminApi.java
sadatmalik: Looking at e.g. the candidate api i see that this would be quite extensive work to replace throughout the code base. Ok let's talk it through. I think advantages might be better type checking, lighter code base, and using Jackson's natural support.

server/src/test/java/org/tbbtalent/server/api/admin/BrandingAdminApiTest.java
sadatmalik: ApiTestBase is in the same package as BrandingAdminApiTest (it was merged with the AuthAdminApi unit tests). It's just a container for some beans we need to mock out in all admin Api unit test classes.

server/src/main/java/org/tbbtalent/server/service/db/impl/BrandingServiceImpl.java
sadatmalik: I lifted the comments out as JavaDoc - so don't think they're lost? A learned habit I guess through working with SonarQube which flags this up.
I wouldn't say the code change is better but maybe it is cleaner? That's what i was aiming at anyway! You're right though it's personal preference - this one's another habit learned from SonarQube which encourages the ternary if/else operator where possible.

server/src/main/java/org/tbbtalent/server/model/db/BrandingInfo.java
sadatmalik: Sure let's chat. I brought it in here primarily for use in the unit test. I think it's a neat pattern to introduce where we can easily do so while we are going through writing unit tests. The Lombok annotation makes it very easy but I wouldn't say it's worth the effort using it everywhere.

server/src/main/java/org/tbbtalent/server/api/admin/BrandingAdminApi.java
camerojo: Fair enough. One of the trickier ways we use DTO's is when we need to convert a collection of Candidates where the conversion of each candidate in the collection is different based on the user requesting the data. Partners get to see full details of candidates assigned to them, but only partial details of candidates who are not.

server/src/test/java/org/tbbtalent/server/api/admin/BrandingAdminApiTest.java
camerojo: Mmmm - for some reason my Intellij could not find the source. It would be good to track down why.

sadatmalik: That's odd. The PR with ApiTestBase is merged into staging (not yet master). Should get picked up if you pull from staging?

server/src/main/java/org/tbbtalent/server/api/admin/BrandingAdminApi.java
```diff
@@ -16,48 +16,34 @@
...
-
+    public ResponseEntity<BrandingInfo> getBrandingInfo() {
         BrandingInfo info = brandingService.getBrandingInfo(null);
-        return brandingInfoDto().build(info);
+        return ResponseEntity.ok(info);
     }
 
-    DtoBuilder brandingInfoDto() {
```
camerojo: Note that this call to getBrandingInfo is for admin portal - and admin portal using default branding - ie is not partner specific - note the null passed to the getBrandingInfo call.

Partner based branding is only supplied to the candidate portal. See the implementation of getBrandingInfo when a non null partner is supplied. In particular, partner based branding includes support for different landing pages.

That is why - for the admin portal - the DTO of brandingInfo does not pass back the registrationLandingPage. There is no concept of a landing page for admin portal logins.

I think that this is what was causing you some confusion in earlier discussions. Sorry I did not pick this up before.

server/src/main/java/org/tbbtalent/server/service/db/impl/BrandingServiceImpl.java
camerojo: Note that this call to getBrandingInfo is for admin portal - and admin portal uses default branding - ie is not partner specific - note the null passed to the getBrandingInfo call.

Partner based branding is only supplied to the candidate portal. See the implementation of getBrandingInfo when a non null partner is supplied. In particular, partner based branding includes support for different landing pages.

That is why - for the admin portal - the DTO of brandingInfo does not pass back the registrationLandingPage. There is no concept of a landing page for admin portal logins.

I think that this is what was causing you some confusion in earlier discussions. Sorry I did not pick this up before.

sadatmalik: .

server/src/main/java/org/tbbtalent/server/api/admin/BrandingAdminApi.java
sadatmalik: Reverting these changes - will re-introduce in a separate PR as agreed in refinement.

server/src/main/java/org/tbbtalent/server/service/db/impl/BrandingServiceImpl.java
sadatmalik: Reverting these changes in this PR - may re-introduce separately if agreed in refinement.

server/src/main/java/org/tbbtalent/server/service/db/impl/BrandingServiceImpl.java
```diff
@@ -34,15 +36,17 @@
...
     private final PartnerService partnerService;
     private final UserService userService;
 
-    public BrandingServiceImpl(PartnerService partnerService, UserService userService) {
-        this.partnerService = partnerService;
-        this.userService = userService;
-    }
-
+    /**
+     * Returns the branding information for a partner.
```
camerojo, line 45: We don't duplicate JavaDoc on method implementations that is already present on the method as described in the interface. That leads to duplication which easily gets out of sync.

I would remove these JavaDoc comments

server/src/main/java/org/tbbtalent/server/service/db/impl/BrandingServiceImpl.java
```diff
@@ -70,8 +74,8 @@ public BrandingInfo getBrandingInfo(@Nullable String partnerAbbreviation) {
     private @NotNull BrandingInfo extractBrandingInfoFromPartner(@NonNull Partner partner) {
         BrandingInfo info = new BrandingInfo();
         info.setLogo(partner.getLogo());
-        if (partner instanceof SourcePartner) {
-            info.setLandingPage(((SourcePartner)partner).getRegistrationLandingPage());
+        if (partner instanceof SourcePartner sourcePartner) {
+            info.setLandingPage((sourcePartner).getRegistrationLandingPage());
```
camerojo: Nice! I didn't know this syntax.

But do you need the (sourcePartner). This would work, right?

```suggestion
            info.setLandingPage(sourcePartner.getRegistrationLandingPage());
```

camerojo (dismissed):

server/src/main/java/org/tbbtalent/server/service/db/impl/BrandingServiceImpl.java
sadatmalik: Yep - makes sense - removed.

camerojo (approved):

PR #31, opened 2023-05-19
Cleaning up javadoc in user.java model

samschlicht:


tbbadmin:
Paired up with Sam on this one. @param makes sense in the context of a method's Javadoc (but don't make sense on class properties which are not params). The doc comments themselves are looking great.

PR #32, opened 2023-05-21
29 deleting the contents of a string field value in search does not clear the search field

### LINKED ISSUE #29: Deleting the contents of a string field value in Search does not clear the search field
To reproduce this issue:

1. Click New Search
2. In Referrer type "Alight" - click Apply - no matches
3. Delete "Alight" in Referrer
4. In Partner select Alight - click Apply - should have matches but don't see any

If you now click "Clear Search" - reselect Partner "Alight" and click Apply - you will see matches.

Suspect that step "3" above doesn't set the referrer to empty (null) but trims it down to a blank string (i.e. "")

samschlicht:
This fixes the bug — curious to know if it's a sensible way to do it!

### DIFF
--- a/ui/admin-portal/src/app/components/search/define-search/define-search.component.ts
+++ b/ui/admin-portal/src/app/components/search/define-search/define-search.component.ts
@@ -324,6 +324,11 @@ export class DefineSearchComponent implements OnInit, OnChanges, OnDestroy {
     request.sortFields = [this.sortField];
     request.sortDirection = this.sortDirection;
 
+    // Converts empty string to null for Referrer input - fixes bug whereby search included "" criterion
+    if (request.regoReferrerParam === "") {
+      request.regoReferrerParam = null;
+    }
+
     //Note that just changing searchRequest triggers the display of the results
     //See the html of this component, for which <app-show-candidates takes
     //searchRequest as an input.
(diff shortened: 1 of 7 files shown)


camerojo (commented):
This PR mixes in a bunch of stuff which has nothing to do with the issue.

I think you have dragged in some of Sadat's tidying up. It would be good to understand how this happened.

ui/admin-portal/src/app/components/search/define-search/define-search.component.ts
```diff
@@ -324,6 +324,11 @@ export class DefineSearchComponent implements OnInit, OnChanges, OnDestroy {
     request.sortFields = [this.sortField];
     request.sortDirection = this.sortDirection;
 
+    // Converts empty string to null for Referrer input - fixes bug whereby search included "" criterion
```
camerojo, line 327: I think that this will fix the problem - as you say - but probably not the best way to do it.

I think it would be better to fix the issue on the Spring server - to handle the special case where empty strings are passed down. In fact, we do already have some code there handling that case for certain fields - just not this one.

We can go though it together in our call tomorrow. If you have time, take a look at CandidateSpecification.java. That is where the candidate search parameters are processed. The fix is just as simple as your Angular based solution below - but better located on the server I think.

PR #36, opened 2023-05-25
Server no longer includes regoReferrerParam in a search request if populated by empty string or null

### LINKED ISSUE #29: Deleting the contents of a string field value in Search does not clear the search field

samschlicht:
Server no longer includes regoReferrerParam in a search request if populated by empty string or null.

### DIFF
--- a/server/src/main/java/org/tbbtalent/server/repository/db/CandidateSpecification.java
+++ b/server/src/main/java/org/tbbtalent/server/repository/db/CandidateSpecification.java
@@ -286,7 +286,7 @@ public class CandidateSpecification {
             }
 
             // REFERRER
-            if (request.getRegoReferrerParam() != null) {
+            if (request.getRegoReferrerParam() != null && !request.getRegoReferrerParam().equals("")) {
                 conjunction.getExpressions().add(
                         builder.equal(candidate.get("regoReferrerParam"), request.getRegoReferrerParam())
                 );


camerojo:
I notice that this PR is committing direct to master - so we still have that problem.

server/src/main/java/org/tbbtalent/server/repository/db/CandidateSpecification.java
```diff
@@ -286,7 +286,7 @@ public static Specification<Candidate> buildSearchQuery(
             }
 
             // REFERRER
-            if (request.getRegoReferrerParam() != null) {
+            if (request.getRegoReferrerParam() != null && !request.getRegoReferrerParam().equals("")) {
```
camerojo, line 289: Even better would be !request.getRegoReferrerParam().trim().equals("")

That will work if there are all spaces - which will also be invisible to user

camerojo: Or maybe simpler request.getRegoReferrerParam().trim().length() != 0

PR #43, opened 2023-05-26
35 we need more flexible mapping of salesforce json api responses to map to java objects

### LINKED ISSUE #35: We need more flexible mapping of Salesforce json API responses to map to Java objects
Our current code uses the default of mapping json field names directly to Java properties on the receiving Java object with the same field name.
 
This does not work when the json field to be mapped uses dot notation - for example "Account.Description". This is not an allowed Java property name - so the default mapping won't work.

@sadatmalik felt that it should be possible to control the mapping using field annotations. Can we confirm that this is possible, and if so implement it in our Spring SalesforceService code.

See, for example, org/tbbtalent/server/model/sf/Opportunity.java and the code that uses it in org/tbbtalent/server/service/db/impl/SalesforceServiceImpl.java

sadatmalik:
- Use JsonSetters to map SF field names to Java field names in Opportunity, Contact and SalesforceObjectBase
- Create unit tests to verify for json string to Opportunity and Contact mappings
- Update all related classes to use the correct getter methods
- Adding Account.Description to accountDescription mapping in Opportunity

### DIFF
--- a/server/src/main/java/org/tbbtalent/server/api/admin/SystemAdminApi.java
+++ b/server/src/main/java/org/tbbtalent/server/api/admin/SystemAdminApi.java
@@ -554,7 +554,7 @@ public class SystemAdminApi {
         log.info("Updating " + contacts.size() + " candidates");
         int count = 0;
         for (Contact contact : contacts) {
-            final String candidateNumber = contact.getTBBid__c().toString();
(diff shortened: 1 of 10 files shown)


camerojo (approved):
Many thanks @sadatmalik  - fyi @cazcam34

PR #47, opened 2023-05-28
29 second attempt deleting the contents of a string field value in search does not clear the search field

samschlicht:


camerojo (approved):
Nice one @samschlicht !

PR #53, opened 2023-06-05
I've rewritten this section of the code, which has fixed the bug

### LINKED ISSUE #52: When creating a new user, 'approver_id' is being set to 6 in the backend when sent null value from the frontend
The HTTP POST request has a null value for approverID but is then 6 on the DB — this isn't happening in development, only production.

samschlicht:
I found the issue, which was with the code I replicated from the partner function in UserServiceImpl.java, specifically this part:

`            if (currentApprover == null) {
                newApprover = creatingUser.getApprover();
            }`

So, if the user we're creating doesn't already have an approver, then it uses the "creating user's" approver — which in my case, in production, was set to Caroline (ID: 6). But in development, the "creating user" is SystemAdmin, who doesn't have an approver — hence the discrepancy, and my not noticing this bug in development. 

This section of code was appropriate for partners but not for approvers — I shouldn't have replicated it.

I've rewritten it from scratch, which is what I should have done in the first place! All working as before, except without this bug.

### DIFF
--- a/server/src/main/java/org/tbbtalent/server/service/db/impl/UserServiceImpl.java
+++ b/server/src/main/java/org/tbbtalent/server/service/db/impl/UserServiceImpl.java
@@ -250,31 +250,20 @@ public class UserServiceImpl implements UserService {
             user.setPartner((PartnerImpl) newSourcePartner);
         }
 
-        //Possibly update the user's approver
+//      Possibly update the user's approver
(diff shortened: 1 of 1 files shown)


camerojo:
I did need to make one change - Intellij highlighted a possible Null pointer exception - if creatingUser was null - which it can be according to the @Nullable on the method call. Always best to check those Inteelij hints - they are very useful.

PR #57, opened 2023-06-05
Add support for candidates and users to use the Rich Text

timhill-iress:
Add support for candidates and users to use the Rich Text (HTML) for  work experience

I've created a short Zoom video of me running through these changes which I will share on slack,

### DIFF
--- a/server/src/main/java/org/tbbtalent/server/model/es/CandidateEs.java
+++ b/server/src/main/java/org/tbbtalent/server/model/es/CandidateEs.java
@@ -47,6 +47,7 @@ import org.tbbtalent.server.model.db.MaritalStatus;
 import org.tbbtalent.server.model.db.ResidenceStatus;
 import org.tbbtalent.server.model.db.UnhcrStatus;
 import org.tbbtalent.server.request.PagedSearchRequest;
+import org.tbbtalent.server.util.html.TextExtracter;
 
 /**
  * This defines the fields which are stored in Elasticsearch "documents"
@@ -181,12 +182,7 @@ public class CandidateEs {
     public CandidateEs() {
     }
 
-    public CandidateEs(Candidate candidate) {
-        this();
-        copy(candidate);
-    }
-
-    public void copy(Candidate candidate) {
+    public void copy(Candidate candidate, TextExtracter textExtracter) {
 
         this.additionalInfo = candidate.getAdditionalInfo();
         this.firstName = candidate.getUser() == null ? null
@@ -284,12 +280,12 @@ public class CandidateEs {
                 String text = null;
                 if (role == null) {
                     if (description != null) {
-                        text = description;
+                        text = textExtracter.ExtractText(description);
                     }
                 } else {
                     text = role;
                     if (description != null) {
-                        text += " " + description;
+                        text += " " + textExtracter.ExtractText(description);
                     }
                 }
 
--- a/server/src/main/java/org/tbbtalent/server/service/db/impl/CandidateServiceImpl.java
+++ b/server/src/main/java/org/tbbtalent/server/service/db/impl/CandidateServiceImpl.java
@@ -173,6 +173,7 @@ import org.tbbtalent.server.service.db.util.PdfHelper;
 import org.tbbtalent.server.util.BeanHelper;
 import org.tbbtalent.server.util.filesystem.GoogleFileSystemDrive;
 import org.tbbtalent.server.util.filesystem.GoogleFileSystemFolder;
+import org.tbbtalent.server.util.html.TextExtracter;
 
 
 /**
@@ -252,6 +253,7 @@ public class CandidateServiceImpl implements CandidateService {
     private final TaskService taskService;
     private final EmailHelper emailHelper;
     private final PdfHelper pdfHelper;
+    private final TextExtracter textExtracter;
 
     @Autowired
     public CandidateServiceImpl(UserRepository userRepository,
@@ -280,7 +282,7 @@ public class CandidateServiceImpl implements CandidateService {
         CandidateExamRepository candidateExamRepository,
         RootRequestService rootRequestService, TaskAssignmentRepository taskAssignmentRepository,
         TaskService taskService, EmailHelper emailHelper,
-        PdfHelper pdfHelper) {
+        PdfHelper pdfHelper, TextExtracter textExtracter) {
         this.userRepository = userRepository;
         this.userService = userService;
         this.candidateRepository = candidateRepository;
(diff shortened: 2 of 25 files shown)


server/src/main/java/org/tbbtalent/server/model/es/CandidateEs.java
camerojo, line 1: All very neat with the TextExtracters! I guess the alternative is what you had before - where you configured ElasticSearch to extract the text.

What do you think are the pros and cons? I would like to get our new CTO, @sadatmalik  involved in the discussion.

server/src/main/java/org/tbbtalent/server/service/db/impl/CandidateServiceImpl.java
camerojo, line 1: I wonder if letting ElasticSearch do the text extraction is less invasive - ie impactful on other parts of the code.

server/src/main/java/org/tbbtalent/server/util/html/JsoupTextExtracterImpl.java
```diff
@@ -0,0 +1,40 @@
...
+@Service
+public class JsoupTextExtracterImpl implements TextExtracter {
+    /**
+     * Given an HTML string extract just the text
+     *
+     * @param html an HTML string
+     * @return the text from the HTML. Tags replaced with space
+     */
+    @Override
+    public String ExtractText(String html) {
```
sadatmalik, line 37: probably just a typo - but best use camel casing for method name here (i.e. extractText)

server/src/main/java/org/tbbtalent/server/util/html/TextExtracter.java
```diff
@@ -0,0 +1,31 @@
...
+ *
+ * @author Tim Hill
+ */
+public interface TextExtracter {
+    /**
+     * Given an HTML string extract just the text
+     * @param html an HTML string
+     * @return the text from the HTML. Tags replaced with space
+     */
+    String ExtractText(String html);
```
sadatmalik, line 30: camel case method names? (i.e. extractText)

ui/admin-portal/src/app/components/candidates/view/occupation/experience/edit/edit-candidate-job-experience.component.html
```diff
@@ -86,7 +86,7 @@ <h4 class="modal-title">Edit Job Experience</h4>
       <!-- DESCRIPTION -->
       <div class="mb-3">
         <label class="form-label" for="description">Description *</label>
-        <textarea type="text" class="md-textarea-auto form-control" rows="8" id="description" placeholder="" [formControlName]="'description'"></textarea>
+        <ngx-wig id="description" placeholder="" [formControlName]="'description'"></ngx-wig>
```
sadatmalik, line 89: Nice

ui/admin-portal/src/app/components/candidates/view/occupation/experience/view-candidate-job-experience.component.html
```diff
@@ -63,7 +63,7 @@
 
         <div class="col-10">
           <h6 class="card-title">{{experience.role}}</h6>
-          <p *ngIf="experience.description">{{experience.description}}</p>
+          <p *ngIf="experience.description" [ngClass]="isHtml(experience?.description)?'html':'text'" [innerHtml]="experience.description"></p>
```
sadatmalik, line 66: wonder if there's any harm defaulting to 'html' in all cases here?

sadatmalik (commented):
Hi Tim - looks great - thanks so much. I've made a few minor comments.

server/src/main/java/org/tbbtalent/server/service/db/impl/CandidateServiceImpl.java
timhill-iress: Hi John, Yes, I think using the ElasticSearch HTML strip filter is less invasive, on the flip side we need to drop and rebuild the index. Definitely a conversation point on the pros and cons of the two approaches.

server/src/main/java/org/tbbtalent/server/util/html/JsoupTextExtracterImpl.java
timhill-iress: lol, I think you've spotted I'm not a real Java developer. Let's chat about the different approaches with the html strip character filter https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-htmlstrip-charfilter.html and Jsoup text extract before I look to resolve this.

ui/admin-portal/src/app/components/candidates/view/occupation/experience/view-candidate-job-experience.component.html
timhill-iress: I found this issue in several places, hence the repeated code in the public, admin and candidate portals. Maybe the answer is to convert all the textual descriptions to HTML description as part of an upgrade then we don't need to worry about the two formats (text, and html). It would certainly reduce the amount of change and also reduce the complexity of dealing with both, which might get messier over time. Let's chat about this when we zoom.

sadatmalik: thanks so much Tim - look forward to chatting with you soon!

server/src/main/java/org/tbbtalent/server/service/db/impl/CandidateServiceImpl.java
```diff
@@ -325,7 +328,8 @@ public int populateElasticCandidates(
         for (Candidate candidate : candidates) {
             try {
                 if (createElastic) {
-                    CandidateEs ces = new CandidateEs(candidate);
+                    CandidateEs ces = new CandidateEs();
+                    ces.copy(candidate, textExtracter);
```
sadatmalik, line 332: Maybe good to consider a factory method here

server/src/main/java/org/tbbtalent/server/model/es/CandidateEs.java
timhill-iress: Discussed on Wednesday.
Decision was to go with this method for the short term, and then move to the ElasticSearch alternative later, when the Elastic Search indexes will need to be rebuild for other changes.

server/src/main/java/org/tbbtalent/server/util/html/JsoupTextExtracterImpl.java
timhill-iress: Corrected.

camerojo (approved):
Let's get this approach into staging and start playing with it.

We might still revert to using Elastic Search to do more work itself - but let's get this version in. We know we need JSoup anyway in order to santize the html

PR #65, opened 2023-06-08
Added the SF stage to server, candidate and admin portals

### LINKED ISSUE #62: Need to add support in TC for new SF candidate opp stage
The stage "Job offer retracted by employer" has been added to Salesforce but there is no support in the TC. Needs to be added.

samschlicht:


camerojo (changes requested):
Hi Sam, can I suggest just one change - chaging the name of the enum to something short and which fits better with the other enums.

Can we change it to just jobOfferRetracted. I don't think that the "ByEmployer" is needed. It is implicit, as in the "noJobOffer" enum. We don't say, for example, noJobOfferFromEmployer.

camerojo:
I know it is your last day - so I am happy to make that minor change if you like @samschlicht

camerojo:
Also - it looks like you didn't merge from staging before making your change, so you changed an old version of the file.

PR #84, opened 2023-06-12
Add Dialog to tailor the CV produced

timhill-iress:


server/src/main/java/org/tbbtalent/server/api/admin/CandidateAdminApi.java
```diff
@@ -320,10 +323,13 @@ public void resolveOutstandingTasks(@RequestBody ResolveTaskAssignmentsRequest r
...
-        return token;
-    }
+    @GetMapping(value = "/token/{cn}", produces = MediaType.TEXT_PLAIN_VALUE)
+    public String generateToken(@PathVariable("cn") String candidateNumber,
+                                @RequestParam(defaultValue = "false") boolean restrictCandidateOccupations,
+                                @RequestParam(defaultValue = "") List<Long> candidateOccupationIds) {
+             CvClaims cvClaims = new CvClaims(candidateNumber, restrictCandidateOccupations, candidateOccupationIds);
+             String token = this.candidateTokenProvider.generateCvToken(cvClaims, 365L);
+             return token;
+        }
```
camerojo, line 333: Let's get this code out there, but I am thinking that we might need to move away from this token based url approach. This was my design originally but I think that there is a better way to do obfuscated urls.

The better way is to create obfuscated urls using some unique string generated from a globally unique id and store the link in a new database table of generated candidate CVs and their generated links. The database would store all the stuff that is currently encoded in the token - ie candidate number, link expiry - plus your new data around restricted candidate occupations. Then when the TC receives such a url, it looks up the link in the database to check expiry and regenerate the CV from the data associated with the link.

We end up with shorter - but still obfuscated - urls (which could include candidate number and maybe the job number in plain text). There are other advantages in recording these autogenerated CV links in the TC database.

server/src/main/java/org/tbbtalent/server/api/pub/CvPublicApi.java
```diff
@@ -50,122 +56,144 @@ public CvPublicApi(CandidateService candidateService,
     public Map<String, Object> decodeCvRequest(@PathVariable("token") String token) {
 
         try {
-            String candidateNumber = candidateTokenProvider.getCandidateNumberFromToken(token);
-            Candidate candidate = candidateService.findByCandidateNumber(candidateNumber);
+            CvClaims claims = candidateTokenProvider.getCvClaimsFromToken(token);
+            Candidate candidate = candidateService.findByCandidateNumber(claims.candidateNumber());
```
camerojo, line 60: In the non token based approach mentioned earlier, this code would just look up the database from data in the url, returning the candidate number and occupation ids. Then build the CV using your existing CvBuilder

ui/admin-portal/src/app/components/candidates/view/view-candidate.component.ts
```diff
@@ -218,8 +219,15 @@ export class ViewCandidateComponent extends MainSidePanelBase implements OnInit
...
-    return 'https://tctalent.org/public-portal/cv/' + this.token;
+    const isDevSetup = document.location.port == '4201';
+    let origin = document.location.hostname;
+    let path = '/public-portal/cv/';
+    let protocol = document.location.protocol;
+    if (isDevSetup) {
+      origin = `${document.location.hostname}:4202`;
+      path = '/cv/'
+    }
+    return `${protocol}//${origin}${path}${this.token}`;
```
talentbeyondboundaries, line 230: Nice one!!

talentbeyondboundaries (approved):
Looks good - but see my comments about moving away from token based url's

Issue #91, opened 2023-06-14
"Recruiter portal" - Changes to support non TBB Recruiter and Employer as recruiter


camerojo:
As a professional recruiter, or as an employer
I want direct access to the TC
so that I can independently search for and engage talent

camerojo:
Employer needs to be able to create SF employer opp directly, rather than using a pre-existing SF opp. Just needs to specify role.

No need to paste in existing opp url.

Employer (partner) creating New Job uses their sfAccountId to create opp.

Recruiter partner needs to also create SF account for associated employer. Need to accept possibility of duplicate employers - distinguishing them with associated recruiter partner on SF account record. Then we might have multiple accounts for the same employer but each with a different recruiter partner. They could be linked together through a parent account link.

camerojo:
- [ ] Need to change isJobOurs in Angular AuthService - see todo. Currently assumes that all jobs are owned by TBB

camerojo:
Employer/Recruiter constraints:

- Can only change stage after CV Review
- Until CV Review they do not see candidate name and other personal details
- [ ] On server need to filter out candidate names for job creators viewing submission lists of candidates whose opps are not past the CVReview stage

camerojo:
Moved these comments to the Epic description and closed

PR #98, opened 2023-06-16
69 additional enhancements for candidate opportunities on tc

### LINKED ISSUE #69: Additional enhancements for Candidate Opportunities on TC
- [x] #104
- [x] #105
- [x] #75
- [x] Add default next step for autocreated prospects
- [x] Adding candidates to SavedList should avoid all additional processing when as candidate is already in the list. See SavedListServiceImpl.addCandidateToList
- [x] Change job icon and opps -> cases
- [x] Live jobs should filter out jobs that are not later than candidate search or that are later than VisaPreparation. Suggest to Louna that she moves Iress opps to Candidate Search. Not prospect.
- [x] Need Show inactive opps option on Opps tab - default should filter out relocated opps
- [x] Opps display should have filter by stage (Zeynep)
- [x] Need logic for who can edit an opp - see ViewCandidateOppComponent. Basically if user is Partner managing candidate or Partner who created job - see todos in AuthService
- [x] #74
- [x] Check for duplicate bug? Adding opp when adding to submission list - check that candidate not already there
- [x] Modify restriction to updating job opp stage - created by if not contactBy (plus system admin)
- [x] #46 
- [x] Easy access to Country and Submission List could be useful - display with opp
- [x] Copy Created from SF
- [x] Need "working" indicator while updates happening on opps (SF is slow)
- [x] Sort of NextStep in opps should be by updated. Hover over NextStep should show date. 
- [x] SF links from opps - for TBB staff only
- [x] Last Change Date - can it be mapped to UpdatedDate. Important to have to see stale opps. Should display in opp detail along with created and update plus who.
- [x] Get rid of LastChangeDate
- [x] Details on selected opp should lock to the top of window like selected candidate card in candidate source

camerojo:
My first pull request! Looking for reviewers.

I know that this is a bit of a monster - sorry - still getting used to pull requests. Will try and make sure that future pull requests are more manageable.

Hoping to push into staging on AWS - then show to Mariam and Zeynep. It delivers some of their requests off the back of the most recent production release.

### DIFF
--- a/server/src/main/java/org/tbbtalent/server/model/db/SalesforceJobOpp.java
+++ b/server/src/main/java/org/tbbtalent/server/model/db/SalesforceJobOpp.java
@@ -66,13 +66,7 @@ public class SalesforceJobOpp extends AbstractAuditableDomainObject<Long> {
     @Column(name = "sf_job_opp_id")
     private String sfId;
 
-    /**
-     * True if the job is currently accepting processing by other than its creator.
-     * Setting it false will make the job hidden on the TC front end to all but the creator.
-     * The first time accepting is set true for a job effectively "publishes" the job so that
-     * others can see it and process it.
-     */
-    private boolean accepting;
+    //TODO JC accepting DB field is no longer used and can be removed
 
     /**
      * Salesforce id of account (ie employer) associated with opportunity
--- a/ui/admin-portal/src/app/components/candidates/home.component.html
+++ b/ui/admin-portal/src/app/components/candidates/home.component.html
@@ -49,7 +49,7 @@
   </ng-container>
 
   <ng-container *ngIf="ownsOpps()" ngbNavItem="MyOpps">
-    <a ngbNavLink title="Opps created by my partner organization">{{loggedInPartner.abbreviation}} Opps</a>
+    <a ngbNavLink title="Candidate job opportunity cases managed by my partner organization">{{loggedInPartner.abbreviation}} Cases</a>
     <ng-template ngbNavContent>
       <app-candidate-opps-with-detail
         [searchBy]="SearchOppsBy.mine"
--- a/ui/admin-portal/src/app/components/job/jobs/jobs.component.html
+++ b/ui/admin-portal/src/app/components/job/jobs/jobs.component.html
@@ -68,7 +68,7 @@
         <td title="{{job.country}}">
           {{job.name| truncate: 36}}
           <a [routerLink]="['/job',job.id]">
-            <i class="fa-solid fa-plane-departure" title="Job"></i>
+            <i class="fa-solid fa-briefcase" title="Job"></i>
           </a>
         </td>
         <td title="Contact {{fullUserName(job.contactUser)}}">
--- a/ui/admin-portal/src/app/services/auth.service.ts
+++ b/ui/admin-portal/src/app/services/auth.service.ts
@@ -149,11 +149,30 @@ export class AuthService {
   }
 
   isCandidateOurs(candidate: ShortCandidate): boolean {
-    return true; //todo
+    let ours = false;
+    const loggedInUser = this.getLoggedInUser()
+    //Must be logged in
+    if (loggedInUser) {
+      ours = candidate.user?.partner?.id === loggedInUser.partner.id;
+    }
+    return ours;
   }
 
   isJobOurs(job: ShortJob): boolean {
-    return true; //todo
+
+    //For now all jobs belong to just the default partner.
+    return this.isDefaultDestinationPartner();
+
+    //todo Eventually when we have proper recruiter partner support, the code will look like this:
+    /*
+      let ours = false;
+      const loggedInUser = this.getLoggedInUser()
+      //Must be logged in
+      if (loggedInUser) {
+        ours = job.recruiterPartner?.id === loggedInUser.partner.id;
+      }
+      return ours;
+    */
   }
 
   /**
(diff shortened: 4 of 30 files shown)


server/src/main/java/org/tbbtalent/server/model/db/SalesforceJobOpp.java
```diff
@@ -66,13 +66,7 @@ public class SalesforceJobOpp extends AbstractAuditableDomainObject<Long> {
...
     private String sfId;
 
-    /**
-     * True if the job is currently accepting processing by other than its creator.
-     * Setting it false will make the job hidden on the TC front end to all but the creator.
-     * The first time accepting is set true for a job effectively "publishes" the job so that
-     * others can see it and process it.
-     */
-    private boolean accepting;
+    //TODO JC accepting DB field is no longer used and can be removed
```
sadatmalik, line 69: If we create a new issue for this on the board (probably as a bit of tech-debt?) then the TODO here can be removed.
The idea being that TODOs in the code are easier to lose track of whereas we'll have eyes on the board and can prioritise as we go along.

ui/admin-portal/src/app/components/candidates/home.component.html
```diff
@@ -49,7 +49,7 @@
   </ng-container>
 
   <ng-container *ngIf="ownsOpps()" ngbNavItem="MyOpps">
-    <a ngbNavLink title="Opps created by my partner organization">{{loggedInPartner.abbreviation}} Opps</a>
+    <a ngbNavLink title="Candidate job opportunity cases managed by my partner organization">{{loggedInPartner.abbreviation}} Cases</a>
```
sadatmalik, line 52: Zeynep will be pleased! :)

ui/admin-portal/src/app/components/job/jobs/jobs.component.html
```diff
@@ -68,7 +68,7 @@
         <td title="{{job.country}}">
           {{job.name| truncate: 36}}
           <a [routerLink]="['/job',job.id]">
-            <i class="fa-solid fa-plane-departure" title="Job"></i>
+            <i class="fa-solid fa-briefcase" title="Job"></i>
```
sadatmalik, line 71: Nice UX touch up :)

ui/admin-portal/src/app/services/auth.service.ts
```diff
@@ -149,11 +149,30 @@ export class AuthService {
   }
 
   isCandidateOurs(candidate: ShortCandidate): boolean {
-    return true; //todo
+    let ours = false;
```
sadatmalik, line 152:
```suggestion
    let ours: boolean = false;
```

ui/admin-portal/src/app/services/auth.service.ts
```diff
@@ -149,11 +149,30 @@ export class AuthService {
...
+    const loggedInUser = this.getLoggedInUser()
```
sadatmalik, line 153:
```suggestion
    const loggedInUser: User = this.getLoggedInUser()
```

ui/admin-portal/src/app/services/auth.service.ts
```diff
@@ -149,11 +149,30 @@ export class AuthService {
...
+    return ours;
   }
 
   isJobOurs(job: ShortJob): boolean {
-    return true; //todo
+
+    //For now all jobs belong to just the default partner.
+    return this.isDefaultDestinationPartner();
+
+    //todo Eventually when we have proper recruiter partner support, the code will look like this:
```
sadatmalik, line 166: I can definitely see the benefit of putting this TODO here. I personally lean towards having an issue on the board under the appropriate epic (e.g. an epic for recruiter partner support) for this sort of thing. Mostly so it doesn't get missed or forgotten. Let's discuss in team refinement what works best going forward.

sadatmalik (approved):
Some comments and suggestions from me but nothing major and no blockers.

Marking as approved so that you can move forward to staging and  demo to Mariam and Zeynep. If it's okay to record that session that would be good.

Laslty, if we get a chance it might be worth taking 10 minutes to do a code walk through in team refinement next week.

server/src/main/java/org/tbbtalent/server/model/db/SalesforceJobOpp.java
camerojo: Good point. I like the TODO's as well sometimes - as a quick reminder - but agree that issues are the way to go. 

By the way, I also have an item, still to be completed in this mega issue, which refers directly to this todo.

ui/admin-portal/src/app/services/auth.service.ts
camerojo: In fact this is completely typesafe in Typescript, and recommended practice. I only add an explicit type occasionally by way of documentation where I think it is informative. 

Adding a line like ours = 123 after the declaration without the boolean, will result in a compile error.

A useful way of playing around with Typescript is the "playground" - here https://www.typescriptlang.org/play/index.html

ui/admin-portal/src/app/services/auth.service.ts
camerojo: Same comment as above - really no need for explicit type

ui/admin-portal/src/app/services/auth.service.ts
camerojo: Definitely agree. I have in fact added an item in an existing issue which refers directly to this TODO.

See #91

ui/admin-portal/src/app/services/auth.service.ts
sadatmalik: I wasn't sure if this might be overkill when I suggested it but thought it would be good to get your view. Good for me to know for future that it's recommended practice - thanks John.

PR #117, opened 2023-06-21
This seems to be all that's required! As long as we're agreed we want…

### LINKED ISSUE #79: Can not delete work experience

samschlicht:
… sourcepartneradmins to have this permission, then it's done. (Noting that I've checked it works for TBB and CRS, since Suci works for the latter. 🙂)


camerojo (approved):
Nice one Sam!

PR #118, opened 2023-06-21
50 allow other languages in elastic search

### LINKED ISSUE #50: Allow other languages in Elastic Search
As a source partner user
I want to be able to use Other Language search criteria in elastic searches
Because I am seeing more Canadian opportunities in French speaking provinces

Mariam -
Please note that we're receiving more opportunities in French-speaking provinces in Canada and would greatly appreciate your assistance in optimizing the TC search features to streamline our candidate sorting process and make it easier for us to find candidates fluent in French.
Currently, the 'Other Language' field is available for filtering French-speaking candidates. However, this field is inaccessible when using the elastic search feature, which is essential for efficient sorting. Consequently, it becomes challenging to effectively filter candidates based on both their skills and French language proficiency.
Could we explore the possibility of unlocking the 'Other Language' field when utilizing the elastic search feature? This adjustment would enable us to sort candidates based on keywords and foreign language proficiency, providing greater flexibility as we encounter different language requirements in the future, including Dutch for some of the opportunities we are now receiving in Belgium and potential others as we scale in the future.

image

sadatmalik:
Enable other languages in elastic searches


camerojo (dismissed):
Nice stuff!!!

sadatmalik:
Pause on merging with staging as this change will affect the ES index dynamic mappings - better to wait until ES staging is provisioned.

camerojo (approved):
Go Sadat!

PR #151, opened 2023-07-10
123 enable all fields for elastic searching

### LINKED ISSUE #123: Enable all fields for elastic searching
As a source partner user
I want to be able to use all fields when elastic searching
To better the accuracy of my candidate search lists

sadatmalik:
enables the following fields for elastic searches:
- Occupation
- Education Level 
- Education Major


server/src/main/java/org/tbbtalent/server/model/es/CandidateEs.java
```diff
@@ -156,7 +155,6 @@ public class CandidateEs {
 
     private String partner;
 
-    @Field(type = FieldType.Text)
```
camerojo, line 175: Wondering why this annotation is no longer needed. It is used elsewhere.

camerojo (dismissed):
Looks good - just one comment - but OK to merge.

server/src/main/java/org/tbbtalent/server/model/es/CandidateEs.java
sadatmalik: This annotation forces the elasticsearch index mapping to full text searches only. 

We are removing it because we want this field to be indexed for both full text and keyword search - which is the default elastic search behaviour if no annotation is provided. 

Keyword mappings are required for elastic search term queries which we want to use for this field (as well as having it included in full text searches).

camerojo (approved):
Looks good! Users will like this!

PR #173, opened 2023-07-14
112-cannot-delete-list

### LINKED ISSUE #112: Cannot delete list.
See https://refugeejobsmarket.slack.com/archives/C42S4AJ73/p1687280507762239

samschlicht:
This only required a small change to the entity relationship between the saved_list and export_column, to avoid violating a foreign key constraint which was causing the error.

### DIFF
--- a/server/src/main/java/org/tbbtalent/server/model/db/SavedList.java
+++ b/server/src/main/java/org/tbbtalent/server/model/db/SavedList.java
@@ -202,7 +202,7 @@ public class SavedList extends AbstractCandidateSource {
     @ManyToMany(fetch = FetchType.LAZY, mappedBy = "sharedLists", cascade = CascadeType.MERGE)
     private Set<User> users = new HashSet<>();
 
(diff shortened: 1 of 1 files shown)


server/src/main/java/org/tbbtalent/server/model/db/SavedList.java
```diff
@@ -202,7 +202,7 @@ public class SavedList extends AbstractCandidateSource {
     @ManyToMany(fetch = FetchType.LAZY, mappedBy = "sharedLists", cascade = CascadeType.MERGE)
     private Set<User> users = new HashSet<>();
 
-    @OneToMany(fetch = FetchType.LAZY, mappedBy = "savedList", cascade = CascadeType.MERGE)
+    @OneToMany(fetch = FetchType.LAZY, mappedBy = "savedList", cascade = CascadeType.ALL)
```
camerojo, line 205: Makes sense! Well done!

Sometimes I think that ALL can be problematic - but this looks straightforward.

camerojo (approved):
It might be good if you could chat about what you learned about ALL.

PR #181, opened 2023-07-18
165 continue work on defining destination partner

### LINKED ISSUE #154: Flatten Partner classes - strict inheritance is too limiting and unnecessary
Uses  branch 165

TBB is both a source partner and a destination partner - which is not possible in a single inheritance model.

- [x] Check where problem is assigning tbb as recruiter (rename to job creator)

### LINKED ISSUE #165: Continue work on defining destination partner
- [x] Undo new TBB destination partner
- [x] Add sourcePartner and jobCreator booleans to Partner
- [x] Update PartnerImpl for sourcePartner and jobCreator
- [x] Populate sourcePartner and jobCreator fields for existing partners - TBB is both
- [x] Change PartnerImpl to rename defaultDestinationPartner to defaultJobCreator
- [x] #154

camerojo:
This pull request covers the following issues:

- #154 
- #165


.idea/.name
```diff
@@ -0,0 +1 @@
+tbbtalentv2
```
sadatmalik, line 1: Should the files in /.idea be kept out of the repository?

server/src/main/java/org/tbbtalent/server/api/admin/PartnerAndRoleBasedDtoPropertyFilter.java
```diff
@@ -108,10 +106,12 @@ public boolean ignoreProperty(Object o, String property) {
...
     private boolean isDefaultPartner(Partner partner) {
         boolean res = false;
-        if (partner instanceof SourcePartner) {
-            res = partner.isDefaultSourcePartner();
-        } else if (partner instanceof DefaultDestinationPartner) {
-            res = partner.isDefaultDestinationPartner();
+        if (partner != null) {
+            if (partner.isSourcePartner()) {
+                res = partner.isDefaultSourcePartner();
+            } else if (partner.isJobCreator()) {
```
sadatmalik, line 112: Using JobCreator throughout is a great choice - very effectively describes what we're talking about. A great example of good naming!

server/src/main/java/org/tbbtalent/server/model/db/partner/Partner.java
```diff
@@ -28,7 +28,7 @@
  *
  * @author John Cameron
  */
-public interface Partner {
+public interface Partner extends SourcePartner {
```
sadatmalik, line 31: Maybe a bit confusing this. Seems more natural the other way around where a SourcePartner is a more specific type of Partner?

server/src/main/java/org/tbbtalent/server/model/db/partner/Partner.java
```diff
@@ -48,14 +48,20 @@ public interface Partner {
     boolean isDefaultSourcePartner();
     void setDefaultSourcePartner(boolean b);
 
+    boolean isJobCreator();
+    void setJobCreator(boolean b);
+
+    boolean isSourcePartner();
```
sadatmalik, line 54: If the Partner interface now extends SourcePartner then wouldn't this always be true?

server/src/main/java/org/tbbtalent/server/model/db/PartnerDtoHelper.java
```diff
@@ -41,9 +42,9 @@ static private class PartnerDtoPropertyFilter implements DtoPropertyFilter {
                 "autoAssignable", "defaultPartnerRef"));
 
         public boolean ignoreProperty(Object o, String property) {
-            //Ignore properties which do not exist on type of partner
-            boolean ignore =
-                sourcePartnerOnlyProperties.contains(property) && ! (o instanceof SourcePartner);
+            Partner partner = (Partner) o;
```
sadatmalik, line 45: Wonder if we need a check here, i.e. could there be any cases where the cast fails?

server/src/main/java/org/tbbtalent/server/model/db/PartnerImpl.java
```diff
@@ -66,8 +50,15 @@ public class PartnerImpl extends AbstractDomainObject<Long>
     @Transient
     private Long contextJobId;
 
+    @Column(name="default_destination_partner")
```
sadatmalik, line 53: Is it worth renaming the column while we're at - default_job_creator ?

server/src/main/java/org/tbbtalent/server/service/db/impl/CandidateServiceImpl.java
```diff
@@ -61,120 +34,50 @@
...
-import org.tbbtalent.server.model.db.SavedSearch;
-import org.tbbtalent.server.model.db.SearchJoin;
-import org.tbbtalent.server.model.db.Status;
-import org.tbbtalent.server.model.db.SurveyType;
-import org.tbbtalent.server.model.db.TaskAssignmentImpl;
-import org.tbbtalent.server.model.db.UnhcrStatus;
-import org.tbbtalent.server.model.db.UploadTaskImpl;
-import org.tbbtalent.server.model.db.User;
-import org.tbbtalent.server.model.db.YesNoUnsure;
+import org.tbbtalent.server.exception.*;
```
sadatmalik, line 37: I lean towards not using wildcard imports i.e. switching that off in intellij

server/src/main/java/org/tbbtalent/server/service/db/impl/CandidateServiceImpl.java
```diff
@@ -1083,8 +986,8 @@ public LoginRequest register(RegisterCandidateRequest request, HttpServletReques
         }
 
         //Assign partner based on the partner abbreviation, if any
-        SourcePartner sourcePartner = (SourcePartner) partnerService.getPartnerFromAbbreviation(partnerAbbreviation);
-        if (sourcePartner == null) {
+        Partner sourcePartner = partnerService.getPartnerFromAbbreviation(partnerAbbreviation);
+        if (sourcePartner == null || !sourcePartner.isSourcePartner()) {
```
sadatmalik, line 990: This is a bit confusing !sourcePartner.isSourcePartner()

server/src/main/resources/db/migration/V1_269__add_partner_source_partner_and_job_creator_fields.sql
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022 Talent Beyond Boundaries.
+ * Copyright (c) 2023 Talent Beyond Boundaries.
```
sadatmalik, line 2: Cool - we should really do this throughout. I've opened a ticket for it.

ui/admin-portal/src/app/components/settings/settings.component.html
```diff
@@ -110,7 +110,7 @@
 
   </ng-container>
 
-  <ng-container *ngIf="operatingPartnerSourcePartnerAdminOrGreater()">
+  <ng-container *ngIf="canSeeExternalLinksAndTasks()">
```
sadatmalik, line 113: Neat

.idea/.name
camerojo: It is useful to standardize on our configuration - ref your comment about imports later.

We haven't got it properly working yet - but I think it will be useful for us to share certain .idea config files.

server/src/main/java/org/tbbtalent/server/model/db/partner/Partner.java
camerojo: Yes - it was before. This was a bit of a hack just to get things working.

In fact I will be getting rid of SourcePartner - and all those separate interfaces now - they don't serve any purpose.

There is no longer any class hierarchy - there is just "Partner" which contains all fields collapsed into a single class. 

When I remove SourcePartner and the other interfaces - which are no longer used anywhere - this confusing bit of code will disappear. There will just be a Partner interface which includes all fields.

server/src/main/java/org/tbbtalent/server/model/db/partner/Partner.java
camerojo: As above. This code will undergo quite a few more changes yet on the road to full JobCreator support.

server/src/main/java/org/tbbtalent/server/model/db/PartnerDtoHelper.java
camerojo: No, this should be safe - PartnerDtoHelper is expressly designed to process partner objects.

ignoreProperty is a generic interface method of DtoPropertyFilter

I suppose I could make DtoPropertyFilter a generic interface. 

eg
`
public interface DtoPropertyFilter<T> {

    boolean ignoreProperty(T o, String property);

    @Nullable
    default T getProperty(T o, String property) {
        T value;
        try {
            value = PropertyUtils.getProperty(o, property);
        } catch (Exception e) {
            value = null;
        }
        return value;
    }
}
`

server/src/main/java/org/tbbtalent/server/model/db/PartnerImpl.java
camerojo: From previous experience we have found that it makes for a smoother migration of database changes involving renaming or deleting database fields if we do it in two releases:
1. Remove all code access to the field being deleted/renamed
2. Do the database change

It just means that you avoid a whole lot of nasty database related errors when converting over to a new release. During the release process you go through a period where you have multiple instances of the code running - some running the new software and some running the old software.

If the new software starts up and renames/deletes database columns that are still accessed by the old code, you can get nasty errors occurring for users still connected  to instances of the old software.

server/src/main/java/org/tbbtalent/server/service/db/impl/CandidateServiceImpl.java
camerojo: Me too. This was done on another machine where I haven't configured Intellij the way we want it.

This is a good example of why it would be good get our standard Intellij configuration into the repo. As I say, I have tried to get that working, but failed so far.

server/src/main/java/org/tbbtalent/server/service/db/impl/CandidateServiceImpl.java
camerojo: Yes - I suppose so. But it is nice to have the variable named "sourcePartner" in the following code once its validity has been checked.

Maybe a comment would help like:
`// Check that returned partner really is a source partner`

server/src/main/resources/db/migration/V1_269__add_partner_source_partner_and_job_creator_fields.sql
camerojo: Intellij has support for autogenerating and updating this. I didn't put that change in - Intellij did.

camerojo:
BEWARE!!

After merging this PR and rebuilding and starting the Spring server in Intellij, you will probably get a strange error (like "missing partnerimp..._dtype"). It is a hangover from Hibernate - which must store some cached information inside Intellij. You need to clear the Intellij file system cache (File | Invalidate Caches) and restart. The rerun the server - no more error!

Screen Shot 2023-07-19 at 3 40 13 pm

server/src/main/java/org/tbbtalent/server/model/db/PartnerImpl.java
sadatmalik: Ok - yes - makes sense.

server/src/main/java/org/tbbtalent/server/service/db/impl/CandidateServiceImpl.java
sadatmalik: Ok thanks John - can pick up on the Intellij shared config separately - good idea.

sadatmalik (approved):
Reviewed and approved - SM

PR #184, opened 2023-07-19
180 my cases only does not display cases for default partner contact

### LINKED ISSUE #180: MyCasesOnly does not display cases for default partner contact
Current my cases only filtering only uses the PartnerJob table

Cases that should belong to the default contact for the partner are not picked up.

camerojo:
Merged PR with new staging


camerojo:
Now that #165 is merged into staging, I will merge this with the new staging. That will mean that only the single actual change will appear in this PR.

sadatmalik (approved):
Ok

PR #187, opened 2023-07-19
179 inverting selections on a list which is not showing closed opps selects all closed opps

### LINKED ISSUE #179: Inverting selections on a list which is not showing closed opps, selects all closed opps
This is a problem because the selection is not visible (because closed opps aren't displaying), so any operation (eg changing stage) on the currently selected opps unexpected also modifies the closed opps.

It would be more intuitive if inverting selection only changed selection for currently displayed opps - ie ignored closed opps if they are not being displayed.

This is reported here: https://refugeejobsmarket.slack.com/archives/C42S4AJ73/p1689628028352509 by Saleem

camerojo:


server/src/main/java/org/tbbtalent/server/api/admin/SavedListCandidateAdminApi.java
```diff
@@ -149,6 +149,20 @@ public void replace(long savedListId, @Valid UpdateExplicitSavedListContentsRequ
         savedListService.mergeSavedList(savedListId, request);
     }
 
+    @Override
+    public  @NotNull List<Map<String, Object>> search(
+        long savedListId, @Valid SavedListGetRequest request) throws NoSuchObjectException {
+        SavedList savedList = savedListService.get(savedListId);
+
+        List<Candidate> candidates = this.candidateService
```
sadatmalik, line 157:
```suggestion
        List<Candidate> candidates = candidateService
```

sadatmalik (approved):
Neat solution - looks great.

PR #188, opened 2023-07-20
107 we need a strong visual indication that the aws test system httptctalent testorg is preview

### LINKED ISSUE #107: We need a strong visual indication that the AWS test system (http://tctalent-test.org/) is PREVIEW

cazcam34:
Also added colour change for local env as requested by @camerojo


camerojo (approved):
Looks good. Approved

PR #226, opened 2023-07-25
225 point to staging registry by default

### LINKED ISSUE #225: Update Gradle build - add environment properties
- [x] Decouple jib from build
- [x] Control both environments with properties (do not default to Prod)

sadatmalik:
- Use explicit property to control jib deployment environments
- Do not trigger jib on build

### DIFF
--- a/server/build.gradle
+++ b/server/build.gradle
@@ -125,7 +125,7 @@ jib {
     to {
         if (project.hasProperty("test-tc-system")) {
             image = "231168606641.dkr.ecr.us-east-1.amazonaws.com/test-ecs"
-        } else {
+        } else if (project.hasProperty("prod-tc-system")) {
             image = "968457613372.dkr.ecr.us-east-1.amazonaws.com/tbbtalentv2"
         }
         credHelper = 'ecr-login'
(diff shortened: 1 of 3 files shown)


server/build.gradle
```diff
@@ -125,15 +125,13 @@ jib {
...
             image = "231168606641.dkr.ecr.us-east-1.amazonaws.com/test-ecs"
-        } else {
+        } else if (project.hasProperty("prod-tc-system")) {
             image = "968457613372.dkr.ecr.us-east-1.amazonaws.com/tbbtalentv2"
         }
         credHelper = 'ecr-login'
     }
 }
 
-tasks.build.dependsOn tasks.jib
```
camerojo, line 135: Will jib run if this is not present?

Yes it will because the above dependency is now in the gradle build file. I don't think it was there previously.

camerojo (approved):

server/build.gradle
sadatmalik: Jib will not run automatically when executing "gradle build"

It has to be called explicitly now - "gradle jib" - this prevents it from running when not required for local builds.

Of course for Git Action builds we want Jib to always run - therefore it must be present as an explicit step. The below PR adds the required step to the Git Actions:

https://github.com/Talent-Catalog/talentcatalog/pull/250

PR #230, opened 2023-07-26
Display (truncated) opportunity stage, added title to show full non t…

### LINKED ISSUE #193: Stage column in TBB Jobs and LIve Job view display issue
Stage values are not mapping (see TBB Cases - i.e this is how they should render)

cazcam34:
…runcated stage.

### DIFF
--- a/ui/admin-portal/src/app/components/job/jobs/jobs.component.html
+++ b/ui/admin-portal/src/app/components/job/jobs/jobs.component.html
@@ -91,10 +91,10 @@
               <i class="fa-solid fa-briefcase" title="Job"></i>
             </a>
           </td>
-          <td title="Contact {{fullUserName(opp.contactUser)}}">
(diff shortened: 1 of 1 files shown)


ui/admin-portal/src/app/components/job/jobs/jobs.component.html
```diff
@@ -91,10 +91,10 @@
               <i class="fa-solid fa-briefcase" title="Job"></i>
             </a>
           </td>
-          <td title="Contact {{fullUserName(opp.contactUser)}}">
```
cazcam34, line 94: Added a tooltip to display the full (non truncated) stage. Moved the contact tooltip from the stage to the created date column. Was there a reason it was on the stage? Also what is the purpose of this, as i didn't there was no contact user on my test job opp so it just said 'contact'.

camerojo (approved):

PR #234, opened 2023-07-28
99 option to copy candidate opp next step and due date up to parent job opp

samschlicht:
Screenshot 2023-07-28 at 5 13 49 pm
Changed approach after discussion w JC - no need for new backend methods. This is now working and seems pretty straightforward, after a lot of discovery. 😅


ui/admin-portal/src/app/components/opportunity/edit-opp/edit-opp.component.ts
```diff
@@ -62,7 +80,42 @@ export class EditOppComponent implements OnInit {
...
+        this.activeModal.close(info)
+    }
   }
 
+  /**
+   * If opp is the only open case of its parent job, sets isOnlyOpenCaseOfParentJob to true — the job has a ShortSavedList property but checking for open cases requires the full SavedList, so this is retrieved before proceeding.
+   * @param opp CandidateOpportunity/case whose progress is being updated
+   *
+   */
+  setIsOnlyOpenCaseOfParentJob(opp: CandidateOpportunity) {
```
camerojo: The naming of this method should really change to simply "isOnlyOpenCaseOfParentJob". There is no need for the "set".

See for example https://en.wikipedia.org/wiki/JavaBeans for the naming conventions - using the "get", "set" and "is prefixes.

This is not really a "bean" - but it does some processing and then returns a boolean value, so a simple prefix of "is" would be logical.

camerojo (changes requested):
See simple suggested changed name of method.

ui/admin-portal/src/app/components/opportunity/edit-opp/edit-opp.component.ts
```diff
@@ -62,7 +80,42 @@ export class EditOppComponent implements OnInit {
...
+        this.activeModal.close(info)
+    }
   }
 
+  /**
+   * If opp is the only open case of its parent job, sets isOnlyOpenCaseOfParentJob to true — the job has a ShortSavedList property but checking for open cases requires the full SavedList, so this is retrieved before proceeding.
+   * @param opp CandidateOpportunity/case whose progress is being updated
+   *
+   */
+  isOnlyOpenCaseOfParentJobCheck(opp: CandidateOpportunity) {
```
camerojo: Sorry to be a pain, but it would be good to get this right.

If you have a method starting with "is" people will naturally assume that the method will return true or false.

So this is confusing. A better name for the method would be "checkOpenCases" or something like that.

camerojo (changes requested):
Sorry - just one trivial change - but I hope you see why.

camerojo (approved):

PR #259, opened 2023-08-09
Remove temporary alert for two stage job creation process

### LINKED ISSUE #45: Remove "Red Alert" notes when someone creates a new job
These notes were added when we moved from a single stage job creation to a two stage job creation.

People should be used to that now, so we can remove the alert.

camerojo:


samschlicht (approved):
V Minor front-end template change. I approve!

PR #291, opened 2023-09-01
289 standardise the final assessment section of visa checks

### LINKED ISSUE #289: Standardise the final assessment section of visa checks
_As a staff member filling out a candidate's visa intake,
I would like to be able to add a final assessment based on the visas on the country I am intaking,
So that I can determine which specific visas the candidate is eligible for._

Currently the questions are hard coded as Australian visas and stored as AU specific fields in database e.g. "Eligible for 494 visa?" Yes/No dropdown. Stored as eligible494 in database on visaJobCheck table.

We need a better way to collect the data for different country visa checks. One thought is a multiselect dropdown field on a question "Visas the candidate is eligible for?", which lists all the country specific visa options. Then a single notes section where more details on each visa eligibility can be added. On the database this would look like two fields "eligibleVisas" & "eligbleVisasNotes". HOWEVER this would mean we need to store all the visa pathways for each country, which is data that could change and may need to be kept up to date manually. 

Alternatively a free text field (e.g. Q: Visas the candidate is eligible for?) and a link to an up to date government site listing the visas could be better, so the final assessment section can just be the 'Put Candidate forward to employer' and 'overall visa notes' questions.

cazcam34:
For canada visa check I have added the final assessment fields. This is a potential example of a standardised visa check, where we have a free text field for eligible visas (see pathway eligibility section) and then the final assessment is just two questions (put forward and final notes). If I can get this into staging and the test environment, I can get Kat and Saleems feedback on this final section.

### DIFF
--- a/server/src/main/java/org/tbbtalent/server/api/admin/CandidateVisaCheckAdminApi.java
+++ b/server/src/main/java/org/tbbtalent/server/api/admin/CandidateVisaCheckAdminApi.java
@@ -80,6 +80,7 @@ public class CandidateVisaCheckAdminApi
                 .add("id")
                 .add("country", countryDto())
                 .add("assessmentNotes")
+                .add("candidateVisaJobChecks", candidateVisaJobDto())
                 ;
     }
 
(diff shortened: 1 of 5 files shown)


cazcam34:
Some issues with local storage of the radio buttons of visa checks and visa job checks. Useful if changing between tabs on the same candidate - however if changing candidates for example, and one candidate stored the last index as index = 2, but the new candidate doesn't have that many items in their array the radio button appears as blank as there is nothing to select. Perhaps worth removing the index memory storage.

camerojo (approved):

PR #298, opened 2023-09-08
Fix bug where education major was not being detected in text searches

### LINKED ISSUE #192: Normalise education major drop down values
As an admin user
I want the selectable drop down values for education major to match the stored database values for this field
So that I get more hits when I include this field in my search criteria

camerojo:
Also added in fix for small unrelated bug related to settings translation searches - wait icon was displaying as if search was underway even though search was not performed because it was invalid.


server/src/main/java/org/tbbtalent/server/model/es/CandidateEs.java
```diff
@@ -289,7 +290,15 @@ public void copy(Candidate candidate, TextExtracter textExtracter) {
...
-                final String text = education.getCourseName();
+                String text = null;
+                final EducationMajor educationMajor = education.getEducationMajor();
+                if (educationMajor != null) {
+                    text = educationMajor.getName();
+                }
+                final String courseName = education.getCourseName();
+                if (courseName != null) {
+                    text = text == null ? courseName : text + " " + courseName;
+                }
```
sadatmalik: I see why you're doing this. But it means the elastic search query will still fail in some cases -because the education major elastic search is specified as a term query, which will look for exact keyword matches.

We could change the way the search is setup in SavedSearchServiceImpl/computeElasticQuery(..) or fix the underlying candidate data before re-building the elastic search index. 

What do you think?

camerojo: I tested this and it worked

sadatmalik: Note we'll need to rebuild the elastic search index before testing.

But even so it will only work in specific cases - i.e. where the education major is null and the course name matches one of the education major drop-down values.

camerojo: I'll take another look, but I did rebuild the ES and tested it and convinced myself that it worked. But maybe I missed something.

camerojo: @sadatmalik I have tested and checked in a new version - can you take a look?

sadatmalik (approved):

PR #307, opened 2023-09-14
Moved export pdf function to resusable component, and added to tabs

### LINKED ISSUE #246: exportToPdf does not work for VisaJobCheckComponent
This code was not compiling in GitHub build of staging - so a new PR #245 was created which commented out call to non existent exportToPdf method.

Also VisaCheckComponentBase did not compile - added @Directive to make it compile - but this class is not used anywhere.

cazcam34:


cazcam34:
There will be a clash with my other pull request ( #310 ) - both working on same file CandidateVisaTab, I know how to resolve conflict just not sure when this happens

camerojo (approved):
Approved

PR #316, opened 2023-09-21
83 selected shareable attachments display issues

### LINKED ISSUE #83: Selected Shareable Attachments display issues
From Naji -

- [ ] Selected Shareable Attachments - CVs selection bar is looking empty (with an x next to the arrow) but the CV is actually connected but not showing the name in the bar. [Caroline]
- [ ] Selected Shareable Attachments - When I upload a CV, sometimes I need to refresh the page few times before it appears in the dropdown list. [Caroline]

cazcam34:
Hard to replicate the bugs - I think I have been able to address the second one, and I asked Naji to provide a screenshot/candidate number for the first bug.


ui/admin-portal/src/app/components/candidates/view/shareable-docs/shareable-docs.component.ts
```diff
@@ -76,6 +77,18 @@ export class ShareableDocsComponent extends AutoSaveComponentBase implements OnI
         this.candidate.listShareableDoc = null;
       }
       this.updatedShareableCV.emit(this.candidate.listShareableCv);
+    } else {
```
cazcam34, line 103: I'm sure there's a much more elegant way to refactor this bulky if/else code block in this method... it's just not coming to me atm and I want to get this pull request out. Any suggestion greatly appreciated!!

camerojo (approved):
Approving - but it does look messy

ui/admin-portal/src/app/components/candidates/view/shareable-docs/shareable-docs.component.html
```diff
@@ -8,38 +8,40 @@
 </div>
 <div *ngIf="candidateSource" class="text-muted mb-3">If left blank, the candidate's default selected shareable documents will be used (if present).</div>
 
-<form [formGroup]="form">
-  <div  class="flex-wrap justify-content-around">
-    <div class="d-flex flex-grow-1 min-width-300 align-items-center justify-content-start mb-2">
+<div *ngIf="this.cvs && this.other">
```
camerojo: Do you understand why neither this.cvs and this.other can be true - I thought it always should be one or the other

camerojo (dismissed):
Approved

ui/admin-portal/src/app/components/candidates/view/shareable-docs/shareable-docs.component.html
cazcam34: They are two seperate filtered arrays from all the attachments, filtered by CV true/false. So both arrays should exist, but they could be empty. Perhaps this isn't right, but that was my thinking. Now we are postponing this bug to next sprint I can have another look

cazcam34:
I ended up in a bit of a web of refactoring trying to fix Naji's bug... I found that the code that was there needed some reflection and I have tried to make it more simple and also fixed the bug. A few key changes:
- I found that by adding {emitEvent: false} to the patchValue, it would stop the api call being done on every candidate change in a list. This I found was causing unnecessary api calls, and was causing bugs in the saving of the docs.
- I added a validation method on the back end to also avoid the wrong attachment being saved to the wrong candidate, which was also addressed by the above dot point on the front end.
- Adding some extra protection to avoid the form being shown before the inputs are loaded, which I believe was a cause of Najis bug on the candidate profile view.
- Fetching the updated candidate when the tab is changed on the candidate portal. This avoids emitting the updated candidate object across multiple components, when we can just fetch the new updated object when needed. I felt this would be simpler.

cazcam34:
Also I'm sorry but it seems something is still wrong with my imports as they are appearing in the files changed in github. I tried to manually change them back, but it won't let me commit the changes to just the imports. It just reverts back. Might need a chat to fix this.

My preferences:
Screen Shot 2023-11-11 at 6 28 40 pm

ui/admin-portal/src/app/components/candidates/view/attachment/view-candidate-attachment.component.ts
```diff
@@ -84,37 +73,7 @@ export class ViewCandidateAttachmentComponent implements OnInit, OnChanges {
...
     if (changes && changes.candidate && changes.candidate.previousValue !== changes.candidate.currentValue) {
       this.doPagedSearch();
-      // Need to get Cvs and Other docs separately as we need them in a list for the dropdown (can't be paged)
-      this.getAllCvs();
-      this.getAllOther();
     }
-
-  }
-
-  getAllCvs() {
```
cazcam34, line 94: These methods weren't used in this component. Zombie code

ui/admin-portal/src/app/components/candidates/view/special-links/view-candidate-special-links.component.ts
```diff
@@ -40,21 +40,6 @@ export class ViewCandidateSpecialLinksComponent implements OnInit, OnChanges {
   ngOnInit() {
   }
 
-  ngOnChanges(changes: SimpleChanges): void {
```
cazcam34, line 43: Removed this code, not longer necessary api call. Fetch the candidate once in parent tab component

sadatmalik:
> Also I'm sorry but it seems something is still wrong with my imports as they are appearing in the files changed in github. I tried to manually change them back, but it won't let me commit the changes to just the imports. It just reverts back. Might need a chat to fix this.

Hi @cazcam34 I've updated the project Readme with some instructions on how to import and setup Google Code styles in IntelliJ. If you follow the instructions in [this PR](https://github.com/Talent-Catalog/talentcatalog/pull/442/files) you should be good to go.

camerojo (approved):
Approved

PR #321, opened 2023-09-21
320 remove tabs from candidate preview card when it is a default search

### LINKED ISSUE #320: Remove tabs from candidate preview card when it is a default search
When I added these tabs (context note, shareable docs + preview CV) it has mistakenly been added to default searches also. This shouldnt appear for those, only for saved searches and lists.

cazcam34:
Small bug I noticed with my added tabs (shareable docs/context notes/preview CV) which was appearing on default searches


camerojo (approved):
Approved

PR #323, opened 2023-09-22
Made CSS changes for card alignment and bottom header alignment

### LINKED ISSUE #322: Card sizing issue in left-hand column (Caroline)

cazcam34:
Moved the margin 16px to the mobile only CSS as this was causing the card size to differ. There was a class coming from the theme (i think) which sets the first child element margin to 0, so thats why it the first card looked fine, and then the second card was showing a margin of 16px.

I've only been able to test the classes directly through the console, so perhaps if you can merge into staging we can give it a better test.

### DIFF
--- a/docs/assets/css/style.scss
+++ b/docs/assets/css/style.scss
@@ -44,18 +44,19 @@
   border: 1px solid #ccc;
   border-radius: 8px;
   padding: 16px;
-  margin: 16px 0;
   background-color: #fff;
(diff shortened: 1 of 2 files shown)


camerojo (approved):
Approved. Magic!! Fingers crossed it works in prod.

camerojo:
You will have to test it. @sadatmalik is taking some well deserved tools down time.

sadatmalik:
It's up on git pages now @cazcam34 , @camerojo - works like a charm!

PR #333, opened 2023-09-26
Temporary filter of the destinations which have functional visa checks

### LINKED ISSUE #332: Remove 'test' visa country destinations
Currently the dropdown when adding a visa check country includes UK and NZ which come from a TBBDestinations value that comes from an environment variable ${TBB_DESTINATIONS:Australia,Canada,New Zealand,United Kingdom}. This is appearing in Production as well as development. When I looked at the env variables in AWS I couldn't see the variable TBB_DESTINATIONS listed, perhaps this is why it is defaulting to the four listed same as development? 

It seems we could either add a new env variable which just contains Australia & Canada for now? Or we update the default to remove New Zealand/UK? Orr as a 'quick fix' I could filter the tbb destinations on the front end so they aren't displayed when selecting a visa.

cazcam34:
See issue for more details


camerojo (approved):
Approved

PR #339, opened 2023-09-29
update versions and add a note about google sheet publishing

sadatmalik:
- spring and elasticsearch versions
- publishing google sheets excludes closed cases


camerojo (approved):
Approved

PR #359, opened 2023-10-05
221 in submission lists the context portion of the side panel display can show the cases progress ie stage next step and next step due date

### LINKED ISSUE #221: In submission lists, the context portion of the side panel display can show the case's progress: ie stage, next step and next step due date
As a partner reviewing candidate cases in a submission list
I would like to be able to easily see and modify the progress of the selected case in the right hand panel

cazcam34:


ui/admin-portal/src/app/components/util/candidate-search-card/candidate-search-card.component.ts
```diff
@@ -89,6 +82,10 @@ export class CandidateSearchCardComponent implements OnInit, AfterViewChecked {
     return this.candidate.selected;
   }
 
+  get isSubmissionList(): boolean {
+    return this.isList && this.candidateSource.sfJobOpp != null
```
camerojo, line 86: I don't think that this is a good use of "get". This is method that performs some logic and returns a value. It is not accessing a property.

If you remove the "get" you need to change the method call in the html - see comment above.

ui/admin-portal/src/app/components/util/candidate-search-card/candidate-search-card.component.html
```diff
@@ -57,6 +57,17 @@ <h5  *ngIf="isContextNoteDisplayed() && !isList">Search Specific</h5>
             </app-cv-preview>
           </ng-template>
         </ng-container>
+
+        <!-- PROGRESS -->
+        <ng-container *ngIf="isSubmissionList" ngbNavItem="progress">
```
camerojo: This should really be "isSubmissionList()" - see comment below on inappropriate use of "get"

camerojo (changes requested):

camerojo (approved):
Approved

PR #362, opened 2023-10-09
Hide the delete icons from visa tab if NOT system admin

### LINKED ISSUE #347: Forced logout for sourcepartneradmins when deleting a visa check. Should throw error instead.
Automatic forced logout for SPA's when it should throw error. BUT should SPAs be allowed to delete a country visa check? What about delete job visa check? I'd suggest no, as these records can be overwritten if needed and really shouldn't be deleted.

cazcam34:


camerojo (approved):
Approved

PR #363, opened 2023-10-10
353 talent catalog alert jobs are being created with non standard country name usa

### LINKED ISSUE #353: Talent Catalog Alert: Jobs are being created with non standard country name: USA
Our standard country name is United States not USA

Alert is sent from copyOpportunityToJobOpp method in SalesforceJobOppServiceImpl.java

To find the issue, the alert should be modified to log more data about the Job Opp so that we can identify where the bad  country is coming from. 

Note that the bad country name should not be coming from Salesforce since Sam put in the change where all country names are selected from a drop down 

The bad name is coming from the TC - then up into SF as Candidate Opps are created from Job opps. SalesforceJobOpp has redundant country text field - which is flagged for removal - see todo's in that class. 

- [x] Flyway to synch up country ids in SalesforceJobOpp with country names

`
update salesforce_job_opp j
set country_object_id =
    (select country.id from country where name = j.country)
where country_object_id is null and country is not null;
`
- [x] Quick fix is to modify existing country fields in DB set to USA
- [x] Remove USA from all SF opps (by changing country in underlying account to something other than United States, then back to United States to force the computed fields to update)
- [x] As suggested in todo's convert over to using country object. Note that sfJobCountry in SavedListBuilderSelector - assuming it is not an object.
- [x] Remove all code access to salesforce_job_opp country text field 

Deprecate salesforce_job_opp country text field for eventual removal

camerojo:
Together with some manual changes to Salesforce, handles the USA instead on 'United States' bug.

Also (in separate commit) replaces some visible references to 'TBB' with 'TC' or 'Talent Catalog'

### DIFF
--- a/server/src/main/java/org/tbbtalent/server/model/db/SalesforceJobOpp.java
+++ b/server/src/main/java/org/tbbtalent/server/model/db/SalesforceJobOpp.java
@@ -76,18 +76,11 @@ public class SalesforceJobOpp extends AbstractOpportunity {
     private User contactUser;
 
     /**
-     * todo Remove this field when we moved across to using countryObject everywhere
-     * FROM SALESFORCE: Name of country where job is located
-     */
-    private String country;
-
-    /**
-     * todo: Once above country field removed, rename countryObject to country
-     * References country object on database (set using the country name above that comes from SF)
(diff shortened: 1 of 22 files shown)


sadatmalik (commented):
Reviewed

PR #376, opened 2023-10-18
20 update cors urls in applicationyml

### LINKED ISSUE #20: Update Cors URLs in application.yml
Edit:

```
tbb:
  cors:
    urls: ${TBB_CORS_URLS:http://localhost:4200,http://localhost:4201,http://localhost:4202,https://tbbtalent-staging.herokuapp.com}
```

edit the heroku url for staging - should be: https://tctalent-test.org

cazcam34:
I changed the cors url as the issue indicated. When pushing it came up with an error of a deprecated property and suggested the change which I put in another commit. (see the image below of what appeared)

Screen Shot 2023-10-19 at 10 29 57 am

### DIFF
--- a/server/src/main/resources/application.yml
+++ b/server/src/main/resources/application.yml
@@ -68,10 +68,6 @@ spring:
   mvc:
     throw-exception-if-no-handler-found: true
 
-  resources:
-    add-mappings: false
-
-
   jpa:
     show-sql: false
     database-platform: org.hibernate.dialect.PostgreSQLDialect
@@ -80,6 +76,9 @@ spring:
         order_by.default_null_ordering: last
         temp:
           use_jdbc_metadata_defaults: false
+  web:
+    resources:
+      add-mappings: false
 
 management:
   health:
--- a/server/src/main/resources/application.yml
+++ b/server/src/main/resources/application.yml
@@ -1,6 +1,6 @@
 tbb:
   cors:
-    urls: ${TBB_CORS_URLS:http://localhost:4200,http://localhost:4201,http://localhost:4202,https://tbbtalent-staging.herokuapp.com}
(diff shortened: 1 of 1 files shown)


cazcam34:
There are still three intellij warnings in this application.yml file which I thought was worth flagging (see screenshot). They appear as warnings not errors so I didn't change anything - however it seems these are incorrect config properties and I wonder if it could be causing us issues?

Screen Shot 2023-10-19 at 10 31 05 am

server/src/main/resources/application.yml
```diff
@@ -1,6 +1,6 @@
 tbb:
   cors:
-    urls: ${TBB_CORS_URLS:http://localhost:4200,http://localhost:4201,http://localhost:4202,https://tbbtalent-staging.herokuapp.com}
+    urls: ${TBB_CORS_URLS:http://localhost:4200,http://localhost:4201,http://localhost:4202,https://tctalent-test.org}
```
camerojo: Don't want to add tctalent-test.org - that will be in the staging JSON config.

server/src/main/resources/application.yml
```diff
@@ -80,6 +76,9 @@ spring:
         order_by.default_null_ordering: last
         temp:
           use_jdbc_metadata_defaults: false
+  web:
+    resources:
+      add-mappings: false
```
camerojo: These other changes should be in separate issues. 

Replacing deprecated stuff can be more complicated than just changing this file.

camerojo (changes requested):
See comments.

Split into separate issues.

This issue should be just about the cors update.

server/src/main/resources/application.yml
cazcam34: Reverted back to how it was before - will create new issue

server/src/main/resources/application.yml
cazcam34: Removed the test URL

camerojo (approved):
Approved

PR #381, opened 2023-10-24
313 v210 release notes

sadatmalik:
This PR:

- Moves v2.2 rls notes to a separate file
- Adds a note and image to the search notes highlighting the "intaked" column


camerojo (approved):
Approved

PR #390, opened 2023-10-25
350 performance test harness

### LINKED ISSUE #350: Performance test harness
Plain empty search - 
select c from Candidate where c.status <> 'deleted' etc. . .draft/employed - see TC search screen

- does it need to be paged - yes - replicate this.

Concurrent admin users - 20

sadatmalik:
Gatling test harness with performance tests for the following DB query scenarios:

- New search screen
- "esload" system API

Includes git workflow to run tests on push to staging.


camerojo (approved):
Approved

PR #396, opened 2023-10-26
379 description field required when adding a job experience need better errorrequired indicator

### LINKED ISSUE #379: Description field required when adding a job experience - need better error/required indicator.
Currently it is not clear that description is required in order to save a job experience on the admin portal.

If it is to be required, make clearer and throw a better error/hide the save button until added. If is isn't required, we need to bypass the error and allow to save.

See slack post from Jana here https://refugeejobsmarket.slack.com/archives/C42S4AJ73/p1698094988408729
Screenshot 2023-10-23 at 10 00 00 PM

cazcam34:
Fixed the description bug but two additional things to note:

1. On the front end the validation for creating/updating a job experience is not consistent or present. So I added validation for description field to be required, however I thought it best to create a new issue #397 for fixing up the validation on the other fields. 

2. I found another small but quite ugly bug where if there was a null field for the Contract Type & Employment Type fields, it was defaulting to displaying the equivalent as if the field was false. Thought I'd put in this bug fix branch as it's small and fairly nasty so best to do it in the hot fix.

### DIFF
--- a/server/src/main/java/org/tctalent/server/request/work/experience/CreateJobExperienceRequest.java
+++ b/server/src/main/java/org/tctalent/server/request/work/experience/CreateJobExperienceRequest.java
@@ -16,13 +16,13 @@
 
 package org.tctalent.server.request.work.experience;
 
-import javax.validation.constraints.NotBlank;
-import javax.validation.constraints.NotNull;
-import java.time.LocalDate;
-
 import lombok.Getter;
 import lombok.Setter;
 import org.tctalent.server.util.html.HtmlSanitizer;
+
+import javax.validation.constraints.NotBlank;
+import javax.validation.constraints.NotNull;
+import java.time.LocalDate;
 
 @Getter
 @Setter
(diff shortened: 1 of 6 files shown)


server/src/main/java/org/tctalent/server/request/work/experience/CreateJobExperienceRequest.java
```diff
@@ -16,14 +16,14 @@
...
-import java.time.LocalDate;
-
 import lombok.Getter;
 import lombok.Setter;
 import org.tctalent.server.util.html.HtmlSanitizer;
 
+import javax.validation.constraints.NotBlank;
+import javax.validation.constraints.NotNull;
+import java.time.LocalDate;
+
```
camerojo, line 26: This is just import order. Can you check your Intellij config

camerojo (changes requested):
See comment

server/src/main/java/org/tctalent/server/request/work/experience/UpdateJobExperienceRequest.java
```diff
@@ -53,6 +53,6 @@ public void setCountry(Long countryId) {
     }
 
     public void setDescription(String description) {
-        this.description = HtmlSanitizer.sanitize(description);
+        this.description = description != null ? HtmlSanitizer.sanitize(description) : null;
```
camerojo, line 56: It would be much better to fix HtmlSanitizer.sanitize so that it handles null input.

camerojo (approved):
See comments - better to fix HtmlSanitizer

cazcam34:
Closed with another pull request here #417

PR #400, opened 2023-10-28
299 add consent to be contacted via email checkbox to candidate registration

### LINKED ISSUE #299: Add consent to be contacted via email checkbox to candidate registration
This has come up recently as the comms team would like to be able to contact candidates en masse, but our mailing service Campaign Monitor has refused on the basis that we don't have proof of consent. 

Leaving aside the question of how and why that should be done, we are likely to need/want to do it at some point, in some form, so it would seem prudent to start collecting consent.

The point for further discussion is why the comms team wants to, what form of TC team approval comms of that type should require, and whether it should really come from TBB?

cazcam34:
Have the code in place saving consent to database but I have a few questions before I think it's ready to go:

1. The translations need to be added when in pushed to production. Who do we request these from?

** Review PR and confirm english defaults have been set  (John)

2. I'm unfamiliar with the campaign monitor integration, if there is one. Not sure who best to answer this, but I have two questions:
- How do we update campaign monitor when new candidates are registered?

Sam: "CM uses SF data to make subscriber lists. Once the TC -> SF sync task is complete then we can have the DB field reflected on to SF records and thereby use it to update CM lists."

** Complete the pending issue re: TC -> SF sync (Sam)
** Open a new issue for the additional DB fields (Sam)

- Not sure if we can allow candidates to update their consent because how would we then update campaign monitor?

Sam: "Manual on CM; retrospectively applied as a batch (job) in TC)"

** Open an issue to retrospectively apply any emailed consent removals to the TC DB (Sadat)

3. I have set up the consent fields as default = false. So every existing candidate's consent is false. How do we get our existing candidates to provide consent? A new task?

John: "Leave it as false - the field means explicit consent. Implicit consent is for previous registrations."

** Open an issue to investigate and confirm an alternative mechanism to CM for requesting consents from these "implicitly consenting" candidate registrations (Sadat)


sadatmalik:
From Sam:

Campaign Monitor integrates with SF to create subscriber lists for marketing emails. You can create a SF report and it’ll automatically keep an associated CM subscriber list synced.

Once we’re syncing all TC candidates to SF (a task that narrowly missed the last release), that’ll be one way to keep the CM subscriber list auto-updated with new TC registrants. We’ll need to add a checkbox to SF contacts to indicate consent to be emailed — for now, as long as we have it in the TC DB, that’ s great.

An alternative option would be to connect directly to the CM API from the TC, but the above seems more expedient to me.

In the meantime, I suppose this is a manual task.

server/src/main/java/org/tctalent/server/model/db/Candidate.java
```diff
@@ -705,6 +707,12 @@ public class Candidate extends AbstractAuditableDomainObject<Long> {
     @Nullable
     private String mediaWillingness;
 
+    @NotNull
+    private Boolean emailConsentRegistration;
```
sadatmalik: Just thinking more about the naming again - would it be better to go with the more inclusie "contactConsentRegistration". Given that candidates could be contacted by other means than email.

Also instead of emailConsentPartners - go with contactConsentPartners?

server/src/main/resources/db/migration/V1_278__add_email_consent_fields.sql
```diff
@@ -0,0 +1,18 @@
...
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE. See the GNU Affero General Public License
+ * for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see https://www.gnu.org/licenses/.
+ */
+
+alter table candidate add column email_consent_registration boolean default false not null;
```
sadatmalik: Same comment for the column names - do we just go with contact_consent_registration and contact_consent_partner

sadatmalik (changes requested):
I think it may be better to go with more inclusive field names. See PR comments.

.idea/codeStyles/codeStyleConfig.xml
```diff
@@ -1,5 +1,5 @@
 <component name="ProjectCodeStyleConfiguration">
   <state>
-    <option name="PREFERRED_PROJECT_CODE_STYLE" value="GoogleStyle" />
+    <option name="PREFERRED_PROJECT_CODE_STYLE" value="Default" />
```
sadatmalik: Let's agree to keep this to GoogleStyle

camerojo (approved):
Approved - note that I needed to rename Flyway

server/src/main/java/org/tctalent/server/model/db/Candidate.java
camerojo: Yes let's do that

camerojo (approved):
Approved

sadatmalik (approved):
Approved

PR #414, opened 2023-10-31
Rename gradle build from tbbtalentv2 and other instances of that label

sadatmalik:
This PR:

- updates settings.gradle rootProject.name
- updates project Readme
- updates git action workflow names


camerojo (approved):
Nice - thanks for picking up some package name changes that I missed

PR #421, opened 2023-11-02
Handle null closing job stage logic and log error and send alert

### LINKED ISSUE #393: Bug: null_exception: java.lang.NullPointerException: Cannot invoke "java.util.EnumMap.get(Object)" because "currentToClosingStageMap" is null
Talent Catalog alert

camerojo:


sadatmalik (approved):
Approved

PR #438, opened 2023-11-11
Hotfix of bug where Watched search notifications were not going out

camerojo:
Made the processing Transactional so that there was a PersistenceContext (Hibernate Session)


sadatmalik (commented):
Reviewed.

PR #464, opened 2023-11-24
372 widen colour palette on admin portal tc lacking strong accent colour very bluesgreys

### LINKED ISSUE #372: Widen colour palette on admin portal TC - lacking strong accent colour. Very blues/greys.
As a user of the admin portal,
I would like to see more contrast of colours,
So that it is visually easier to navigate.

Currently use TBB branding colours ([here](https://drive.google.com/file/d/1nMtS50U6mvKJDgcgfCJldtybnr2Wie4I/view?usp=sharing)) which is hard to use as the bright TBB blue & teal colour doesn't contrast and display very well on our white background. Also doesn't have a strong accent colour, so it all appears very grey/blue. So it is difficult when trying to highlight different elements as we are limited in our colours currently.

Also as we want to make the TC less TBB branded, perhaps we should look at using more standardised (eg. Bootstrap standard) colours. Then we could also incorporate the partner brand colour as a variable 'brand' colour.

cazcam34:
Apologies for the large PR! Updated admin portal colour theming and some other redesigning of components. 

The SCSS files are still confusing with some old CSS and will need more cleaning up. I'm trying to use the classes that bootstrap provides (color being primary, secondary, info, success, warning, danger) as much as possible:
- btn-{color}
- link-{color}
- text-{color}

I also use the spacing classes such as margin/padding classes: https://getbootstrap.com/docs/5.0/utilities/spacing/

Note bootstrap doesn't provide classes for the accent colours, so if needed can create those in a global scss file eg:
`.btn-accent-1 { background-color: $accent1, color: $white }`

Another note, I am getting these intellj warnings regarding imports in SCSS files (bootstrap and font-awesome) here's an example below when pushing changes to the common.scss file. The imports haven't changed in this PR, and it still seems to be importing fine. However just wanted to flag.
Screen Shot 2023-11-24 at 11 16 30 am

### DIFF
--- a/ui/admin-portal/src/app/components/infographics/infographic.component.html
+++ b/ui/admin-portal/src/app/components/infographics/infographic.component.html
@@ -90,23 +90,32 @@
   <!--Loop through reports data, displaying each report-->
   <div *ngIf="statReports">
 
-    <nav class="navbar sticky-top navbar-expand-md navbar-light bg-light second-nav">
+    <nav class="navbar sticky-top navbar-expand-md second-nav">
       <span class="navbar-brand">Stats for {{statsName}}</span>
         <ul class="navbar-nav ms-auto">
-          <button class="btn btn-outline-secondary float-end" [disabled]="!dataLoaded" (click)="exportStats()">
-            <i class="fas fa-file-excel"></i> Export
-          </button>
-
-          <li class="nav-item" ngbDropdown display="dynamic" placement="bottom-right">
-            <a class="nav-link" tabindex="0" ngbDropdownToggle id="navbarDropdown3" role="button">
-              Jump to stats report
-            </a>
-            <div ngbDropdownMenu aria-labelledby="navbarDropdown3" class="dropdown-menu scroll-dropdown">
+          <div ngbDropdown display="dynamic" placement="bottom-end">
+            <button type="button" class="btn btn-info float-end" id="jumpToStats" ngbDropdownToggle>Go to stat <i class="fa-solid fa-hand-point-down"></i></button>
+            <div ngbDropdownMenu aria-labelledby="jumpToStats">
               <ng-container *ngFor="let statReport of statReports; let i = index">
                 <a ngbDropdownItem (click)="scroll(i)">{{statReport.name}}</a>
               </ng-container>
             </div>
-          </li>
+          </div>
+
+          <button class="btn btn-secondary float-end  ms-2" [disabled]="!dataLoaded" (click)="exportStats()">
+            Export <i class="fas fa-file-excel"></i>
+          </button>
+
+<!--          <div class="nav-item" ngbDropdown display="dynamic" placement="bottom-right">-->
+<!--            <a class="btn btn-info" tabindex="0" ngbDropdownToggle id="navbarDropdown3" role="button">-->
+<!--              Jump to stats report-->
+<!--            </a>-->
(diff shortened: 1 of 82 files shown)


samschlicht:
Great to read this through for my own understanding and saw no red flags! Appreciate the SCSS coming into a bit more order and it all looked great in the meeting walkthrough. 👍

camerojo:
@cazcam34 - this might help regarding the bootstrap paths - https://www.freecodecamp.org/news/how-to-add-bootstrap-css-framework-to-an-angular-application/

Not sure why it seems to work anyway but why not try and follow the above guidelines and see if that fixes it.

ui/admin-portal/src/app/components/infographics/infographic.component.html
```diff
@@ -90,23 +90,32 @@
...
+<!--          <div class="nav-item" ngbDropdown display="dynamic" placement="bottom-right">-->
+<!--            <a class="btn btn-info" tabindex="0" ngbDropdownToggle id="navbarDropdown3" role="button">-->
+<!--              Jump to stats report-->
+<!--            </a>-->
+<!--            <div ngbDropdownMenu aria-labelledby="navbarDropdown3" class="dropdown-menu scroll-dropdown">-->
+<!--              <ng-container *ngFor="let statReport of statReports; let i = index">-->
+<!--                <a ngbDropdownItem (click)="scroll(i)">{{statReport.name}}</a>-->
+<!--              </ng-container>-->
+<!--            </div>-->
+<!--          </div>-->
```
sadatmalik: Hey Caroline, this may just be an oversight, but unless there's a reason to keep it, is it Ok to remove this commented out code?

cazcam34: Ah yes apologies my mistake! I'll remove that and repush!

cazcam34:
> @cazcam34 - this might help regarding the bootstrap paths - https://www.freecodecamp.org/news/how-to-add-bootstrap-css-framework-to-an-angular-application/
> 
> Not sure why it seems to work anyway but why not try and follow the above guidelines and see if that fixes it.

Thanks @camerojo, I gave that a try and it updated the bootstrap version but other than that didn't change anything. It still is appearing as it can't resolve the directory at `@import '~bootstrap/scss/bootstrap';` it also isn't finding the font awesome directory or ng select. Basically none of the external imports are resolving in intellij however they are all working.

I followed the bootstrap doc here: https://getbootstrap.com/docs/5.0/customize/sass/. changing the import to `@import "../node_modules/bootstrap/scss/bootstrap";` got rid of the intellij error and resolved the path. However we still have the same issues with the font-awesome and ng-select imports. Can't find a solution for those imports, it seems to be pretty standard what we are doing. Need to investigate further... could be a broader issue / an intellij issue? Especially since all the imports are working fine?

**Edit:** Ok I found a solution! It just required me to remove the ~ infront of the import path (e.g. instead of `@import '~bootstrap/scss/bootstrap'` make it `@import 'bootstrap/scss/bootstrap'`). I was looking [here](https://intellij-support.jetbrains.com/hc/en-us/community/posts/360000048404-SCSS-import-paths-cannot-be-resolved) and found that perhaps it was something to do with the resource root. Did we update the resource root when we renamed the project tctalent? Just a thought - not sure why/how it worked but it seemed to!

FYI @sadatmalik @samschlicht

camerojo (approved):
Approved

PR #468, opened 2023-11-27
Added environment variable to application.yml and created additional-spring-configuration-metadata in main directory which helps Intellij work with new runtime properties

### LINKED ISSUE #465: Implement environment service for backend
On the frontend I recently added env.service.ts, which enables environment-contingent methods and variables — we could use something similar on the backend.

samschlicht:
https://www.baeldung.com/spring-boot-properties-env-variables
https://www.baeldung.com/intellij-resolve-spring-boot-configuration-properties#2-metadata-without-a-configurationproperties-bean


camerojo (approved):
Approved

Issue #311, opened 2023-09-18
Remove Recaptcha


samschlicht:
See pic for the little error at the bottom of the screen. The full text reads 'ERROR for site owner: Invalid domain for site key Privacy - Terms'

Not critical as presumably an attack on tc-test is unlikely — but not very elegant, as we start to make more use of tc-test.

A quick reading suggests that we just need to log into Google an add this domain name to the site key. https://help.whmcs.com/m/troubleshooting/l/1189046-troubleshooting-recaptcha-invalid-domain-for-site-key

[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBb1dwIiwiZXhwIjpudWxsLCJwdXIiOiJibG9iX2lkIn19--26ab80971a56f29382999dccaafe33d4d6ef7774/Screenshot%202023-09-18%20at%202.34.37%20pm.png)

camerojo:
We no longer need recaptcha on admin now that we have standardized on MFA - so I will remove it altogether

PR #473, opened 2023-11-28
#453 Candidate portal missing status case

### LINKED ISSUE #453: Candidate portal login of candidates who have status unreachable incorrectly tells them that their registration is incomplete
The issue is in the home.component.html which uses an ngSwitch on candidate status.

There is no case for status=unreachable so the default case is used - which displays an inappropriate message.

camerojo:


ui/admin-portal/src/app/components/login/login.component.ts
```diff
@@ -77,13 +77,14 @@ export class LoginComponent implements OnInit {
...
-        console.log(error);
-      }
-    );
+    // const action = 'login';
+    // this.reCaptchaV3Service.execute(action).subscribe(
+    //   (token) => this.loginWithToken(token),
+    //   (error) => {
+    //     console.log(error);
+    //   }
+    // );
```
sadatmalik, line 86: So this will bypass the recaptcha token - is this what we want to do?

ui/candidate-portal/src/app/components/account/login/login.component.ts
```diff
@@ -71,13 +71,14 @@ export class LoginComponent implements OnInit {
...
-        console.log(error);
-      }
-    );
+    // const action = 'login';
+    // this.reCaptchaV3Service.execute(action).subscribe(
+    //   (token) => this.loginWithToken(token),
+    //   (error) => {
+    //     console.log(error);
+    //   }
+    // );
```
sadatmalik, line 80: Same comment here as with the admin-portal login - bypassing recaptcha - is that intentional?

sadatmalik (commented):
Just a question on the recaptcha token code

ui/admin-portal/src/app/components/login/login.component.ts
camerojo: Yes - recaptcha is for the chop. It is no longer needed since we moved to MFA. Checking is alreday disabled on the server - has been for a long time.
My current "In progress" issue #311 will pull it out altogether.

sadatmalik (approved):

PR #474, opened 2023-11-29
Now removes field by dragging away

### LINKED ISSUE #272: Cannot remove published fields by dragging them away
[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBdUFYIiwiZXhwIjpudWxsLCJwdXIiOiJibG9iX2lkIn19--7c7d5292e87bc17dcba58f1396d49aec6187219d/Screenshot%202023-08-18%20at%206.58.25%20pm.png)

cazcam34:
Fixed without needing to upgrade (just needed removeOnSpill set to true) - followed the doc here: https://valor-software.com/ng2-dragula/index.html

Modified some styling as per new TC style

### DIFF
--- a/ui/admin-portal/package-lock.json
+++ b/ui/admin-portal/package-lock.json
@@ -2440,6 +2440,15 @@
         "node": ">=6.9.0"
       }
     },
+    "node_modules/@colors/colors": {
+      "version": "1.5.0",
+      "resolved": "https://registry.npmjs.org/@colors/colors/-/colors-1.5.0.tgz",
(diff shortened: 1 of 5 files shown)


cazcam34:
Side note... I ran npm install as I was instructed to since updating staging. I ran it while in this branch, so it's pushed up my updated package lock files. What is the best way to do the npm installs, it seems wrong that those package lock files are in my commit?

camerojo:
> Side note... I ran npm install as I was instructed to since updating staging. I ran it while in this branch, so it's pushed up my updated package lock files. What is the best way to do the npm installs, it seems wrong that those package lock files are in my commit?

We have always done that - and I think it was recommended somewhere - but I don't know why. It is never used in builds - the builds only look at the package.json - and autogenerate a new package lock file. So I don't think it matters,

camerojo (approved):
Approved - see comment

PR #475, opened 2023-11-29
306 remove old fields from candidate visa check table

### LINKED ISSUE #306: Remove old fields from Candidate Visa Check table
No longer use the fields 'eligibility' or 'assessmentNotes' these were added at the very start of the visa check development, but aren't used.

cazcam34:
Removed the old fields in the code and in a flyways branch.


camerojo (approved):
Approved

Issue #169, opened 2023-07-13
SF staging sandbox


sadatmalik:

samschlicht:
@camerojo steps so far:
- Logged into SF sandbox (Prod SF > Setup > Sandboxes)
- Setup > Apps > App Manager
- Replicated prod connection settings except for the root URL and digital certificate, which was generated using this command: openssl req -x509 -newkey rsa:1024 -nodes -keyout key.pem -out cert.pem -days 36500 -subj "/O=Talent Catalog/CN=www.tctalent-test.org/"
- Next, we need to point the test app at the SF staging sandbox (application.yaml / terraform.tfvars ?), using the new private key

samschlicht:
Logging a Q/concern - there are files such as...
 
- ui/admin-portal/src/app/model/base.ts /  
- ui/admin-portal/src/app/services/salesforce.service.ts
- server/src/main/java/org/tbbtalent/server/util/SalesforceHelper.java
- server/src/main/java/org/tbbtalent/server/model/sf/SalesforceObjectBase.java

...that use the SF root URL, which is different for the sandbox. This may be a tricky aspect of making this work.

samschlicht:
[ ] update configuration JSON on AWS test

samschlicht:
- [ ] Once happy with all the above steps, the next will be to test that connection to SF prod still works by including the env var in the TEST AWS JSON
- [ ] Assuming that it does, remove it, then add it to a new PROD AWS JSON

We should then be confident that we have a working SF Sandbox connection in TC dev and test, and a working SF Prod connection in TC prod.

samschlicht:
To see SF client id Setup > App Manager > right-hand arrow > 'View' > 'Manage Consumer Details'

samschlicht:
move client_id to tbb_secrets etc instead of in version control

samschlicht:
Deleted my previous comments because I'm using quite a different approach than initially planned.

Rather than create a new SF connected app, I'm able to use the existing 'tbbtalent' SF _connected app_ that TC prod is using for connection to SF prod. The same credentials work for my sandbox user to gain access to the sandbox, using a different set of variables for the JWT bearer token (which obtains an access token from SF) and URLs to point to and edit/create objects in the respective environment.

I've manually edited these variables and everything's working from dev. Next steps:

[...]

PR #495, opened 2023-12-04
all live candidates nightly SF->TC sync

### LINKED ISSUE #402: Sync TC candidates to SF
As a TBB staff member
I would like to be able to access and analyse live-ish candidate data using Salesforce and Sopact reporting tools
To better understand and present information regarding our candidates

This was primarily instigated by the Sopact-TBB team's request for access to this data for the impact dashboard. There have also been other requests, such as from the comms team to populate their candidate mailing list and more efficiently produce the Talent Catalog snapshot.

John met with Sopact and tentatively settled on the most efficient and secure way of achieving this being to extend our SF API connection to include all live candidate records.

The Sopact team and other stakeholders will need to decide what data they need to gather on SF and as part of this issue we'll adapt the contact records accordingly. Then we can set up a nightly sync.

samschlicht:


samschlicht:
@sadatmalik @camerojo 
Just noting that the failed check is the same as last time I had to amend the terraform build — Sadat surveyed it as harmless and we went ahead without issues!

camerojo (dismissed):

sadatmalik:
Here's the analysis from the previous time when we have encountered an issue with the Terraform plan action:

https://refugeejobsmarket.slack.com/archives/C0583HJ9CHM/p1694697103881089

The conclusion at that time was this could be safely ignored. The issue was that the Plan step was failing because it could not yet retrieve the terraform.tfvars file - which is what we would expect.

The issue this time is a different failure cause - it is the Init step that is failing, not the Plan step like before.

Needs further investigation before merging.

sadatmalik (changes requested):
See my PR comment - need to check the failed Terraform action. It is failing on a different step than the previous time.

samschlicht:
Thanks, @sadatmalik! The errors on the checks for my previous SF issue #169 were actually the same, as pictured:
Screenshot 2024-01-10 at 1 01 40 pm

Screenshot 2024-01-10 at 1 02 21 pm

I think the confusion may stem from your Slack post having addressed the errors that were causing the fail (Exit code 1), but not the deprecation warnings. Noting that I decided to fix those warnings anyway, moving forward as #557. 

There's a similar deprecation warning for tc-build so I've logged that as issue #558.

samschlicht:
@sadatmalik I'm actually thinking we should do this TODO to avoid future confusion resulting from Terraform checks, so I've logged it as issue #559.

`      - name: Plan Terraform
        id: plan
        run: |
          terraform plan -no-color -input=false
        # TODO: terraform.tfvars should be copied from github secrets to the env.INFRA_PATH directory
        continue-on-error: true
        working-directory: ${{ env.INFRA_PATH }}`

samschlicht:
@sadatmalik actually just realised I didn't fully address your comment — you may not remember last time we were up against this failed terraform check, we actually took a little run at it with John in the dev meeting. 

We appended the init statement like so: `terraform init -upgrade`, which got us past the init stage and onto plan, where it subsequently failed anyway. It's at that point that you did your investigation resulting in the Slack post. So, unless I'm missing something, we have the same error. I've made a note to discuss in our 1-1.

sadatmalik:
Hey @samschlicht just noting that the error was indeed different. In the action summary you will see it terminate with exit code 1 in each case. This is what you have posted above. But it is a little misleading as the action itself is named Terraform plan, so it appears as if it's failed on the "plan" step in both cases. However you need to click into the action details to see where it actually failed:

Pre-Christmas - init step failing:
Screenshot 2024-01-12 at 09 17 42

Now - plan step failing:
Screenshot 2024-01-12 at 09 18 48

You are correct though that it is the `terraform init -upgrade` that was required at the init step failure - well done for remembering. I do recall now that we did this in a 1-1 when troubleshooting it the first time around. Worth documenting all this.

Now that it is failing at the plan step - it is - as before - safe to ignore. Therefore I'm removing the check PR status. Unfortunately it's picked up some merge conflicts; but it is good to go once they are resolved.

sadatmalik (dismissed):

sadatmalik (approved):
Approved

Issue #161, opened 2023-07-12
Sync TC candidate dependants with SF [Linking family size information from TC intakes to SF candidate opportunities]


sadatmalik:
SF can have dependants (synced from TC) and dependants relocating. If dependants relocating is blank, it can default from dependants.

That way, updating from TC will only overwrite dependents on SF. Dependants relocating can be updated manually (eg removing some dependants who are not actually going to relocate)

See https://refugeejobsmarket.slack.com/archives/C7QK72T4J/p1647914873328889?thread_ts=1647764457.197539&cid=C7QK72T4J

sadatmalik:
Migrated from Trello following discussion with Zeynep.

Duplicate Trello tickets:
- Link dependants gender field to salesforce to calculate relocation information for dependents
- Linking family size from TC to people moves report on SF. Will need: Relocating men, Relocating women, Relocating boys, Relocating girls

tbbadmin:
- [x] #148

samschlicht:
**Draft plan**
Will test this in TC Staging-SF Sandbox before making any changes to SF Prod - noting slight concern re object limit on partial-copy sandboxes, which is 10,000 per object. I'm thinking I'll only sync the first 5,000 from TC Staging, so there'll need to be a bit of if logic to deal with that. Judging by some online comments, hitting that upper limit can make the sandbox do some strange things that I'd rather not have to deal with.

 Add desired to fields to SF SANDBOX contact records, which Sarah, Zeynep and I compiled in this doc (will add link to this as soon as I find it!) - including the new email consent fields that Caroline has added.
-[ ] Amend ContactRequest nested class in SalesforceService to include these fields
-[ ]  Make a @Scheduled method in SalesforceService using SalesforceService.createOrUpdateCandidates, passing an ordered list of all live candidates on DB, incl. also the method from CandidateOpportunityService.createUpdateCandidateOpportunities to update the SF link for all of them (which is why the list needs to be ordered) (Q - is it a problem to do this for all candidates every time? I'm thinking we will need to update existing SF records, so I don't know what the alternative would be)
-[ ] After merge, check that it's working on TC Staging and SF Sandbox and, if so, replicate the object changes on SF prod ready for the next release (or is there a way to check an @Scheduled method from dev?)
[...]

PR #496, opened 2023-12-04
416 visa related candidate dependant information needs to be stored separately on the tc database

### LINKED ISSUE #416: Visa related candidate dependant information needs to be stored separately on the TC database
As a destination partner working on relocation visas
I need to be able to specify which of a candidate's dependants (if any) will be relocating with them
so that the visa application can be submitted.

As a work around - this information about who is coming with the candidate can be recorded in notes or externally.

See discussion at https://refugeejobsmarket.slack.com/archives/C03U6MWRTN0/p1698835871742759?thread_ts=1697750466.992479&cid=C03U6MWRTN0

cazcam34:
Added new field to allow users to select which dependants are to be included on each visa job check. Although a many to many table seemed to make sense (one visa job check can have multiple dependants, but also a dependant can be associated with multiple visa job checks) it felt a little unnecessary, as we dont need the inverse relationship of looking at a dependant and seeing which visa jobs are associated. 

I discussed with @camerojo and it seemed saving a list of ids of the candidate dependants was a simpler and better solution for what we are needing. I added some doc to the code as to why it is saved as a string (and I used a similar example of countryIds string in SavedSearch). This then required some conversion to a list of ids so that we could display these in the ng select.

### DIFF
--- a/server/src/main/java/org/tctalent/server/api/admin/CandidateVisaCheckAdminApi.java
+++ b/server/src/main/java/org/tctalent/server/api/admin/CandidateVisaCheckAdminApi.java
@@ -178,6 +178,7 @@ public class CandidateVisaCheckAdminApi
                 .add("occupationSubCategory")
                 .add("englishThreshold")
                 .add("englishThresholdNotes")
+                .add("relocatingDependantIds")
                 ;
     }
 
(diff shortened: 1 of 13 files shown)


cazcam34:
Just remembered that this influences #161 - is the string of dependant ids going to be useful when synced to SF? @samschlicht this might be worth a look to see if this will work with your plans for #161.

sadatmalik (approved):
Reviewed + approved.

PR #528, opened 2023-12-18
Clarify registration screen

### LINKED ISSUE #131: Clarify registration screen
As a Candidate
I would like it to be clearer what my email address will be used for
So that I know it will be used as my account username / login

Agree a way with Noura/Zeynep etc. how to convey this info to candidates in the appropriate place with the appropriate wording so that

i) candidates are clear on how their email will be used and will remember what they are inputting in this field
ii) candidates consent to their email address being used for us to contact them regarding their application early on in the registration process - i.e. so we can contact them should they fail to fully complete their registration to remind them to do so

Desirable -
It would be good to send an email verification link to candidates - this would in no way restrict them from continuing with their registration - but it is useful info for admins to have so they can see which email accounts were verified during the registration process and which were not

sadatmalik:
This PR:

Introduces a new step in the candidate registration process for the explicit creation of a user account prior to proceeding with further registration steps.

See:
https://www.canva.com/design/DAFy2GEU1hs/DTN3tqBnCQxZpcPoLCFkQw/edit


sadatmalik:
Done:

cazcam34 (approved):

PR #551, opened 2024-01-08
Added new relocated no job offer pathway stage

### LINKED ISSUE #257: New inactive stage for "No Job Offer" candidates - "Relocated through no job offer pathway"
Canada has a pathway which selects refugees based on their skills, without needing any sponsoring job offer.

This is a controversial pathway (refugee selection based on skills rather than vulnerability) and does not match TBB's employer led labour mobility model. However, TBB has chosen to use this pathway in Canada.

We need a way of tagging candidates that relocate through this pathway so that they can be reported on separately.

cazcam34:


server/src/main/java/org/tctalent/server/model/db/CandidateOpportunityStage.java
```diff
@@ -19,9 +19,8 @@
 /**
  * Note that the string values of this enum MUST match the actual stage names for candidate
  * opportunities on Salesforce.
- * <p/>
  * See https://docs.google.com/document/d/1B6DmpYaONV_yNmyAqL76cu0TUQcpNgKtOmKELCkpRoc/edit#heading=h.qx7je1tuwoqv
- * <p/>
```
camerojo, line 24: Why did you remove this spacing? It makes the generated Javadoc look better.

camerojo (approved):
Approved but see comment about the Javadoc
@cazcam34

PR #554, opened 2024-01-08
351 create sf jobs from tc

### LINKED ISSUE #351: Create SF jobs from TC
TC staff create jobs directly on SF at the moment. This doesn't work for recruiter access, who will not have access to SF.

We still need SF jobs to be created but this should all be done from the TC.

camerojo:


server/src/main/java/org/tctalent/server/model/db/AbstractSalesforceObject.java
```diff
@@ -0,0 +1,43 @@
...
+
+/**
+ * Base class of entities with corresponding objects on Salesforce.
+ *
+ * @author John Cameron
+ */
+@Getter
+@Setter
+@MappedSuperclass
+public class AbstractSalesforceObject extends AbstractAuditableDomainObject<Long> {
```
sadatmalik, line 31: `public abstract class AbstractSalesforceObject extends AbstractAuditableDomainObject<Long> {`

sadatmalik, line 31:
```suggestion
public abstract class AbstractSalesforceObject extends AbstractAuditableDomainObject<Long> {
```

server/src/main/java/org/tctalent/server/model/db/AbstractOpportunity.java
```diff
@@ -32,7 +32,7 @@
 @Getter
 @Setter
 @MappedSuperclass
-public class AbstractOpportunity extends AbstractAuditableDomainObject<Long> {
+public class AbstractOpportunity extends AbstractSalesforceObject {
```
sadatmalik, line 35:
```suggestion
public abstract class AbstractOpportunity extends AbstractSalesforceObject {
```

sadatmalik (dismissed):
Approved

server/src/main/java/org/tctalent/server/model/db/AbstractSalesforceObject.java
camerojo: I initially had this as abstract - but I recall it lead to JPA problems

PR #561, opened 2024-01-11
Set Pageable on Stats elasticsearch query to unpaged…

### LINKED ISSUE #388: Bug: Mismatch between search results and stats
From Lea:

https://tctalent.org/admin-portal/infographics/search/2575

As you can see in the first snapshot the total number retrieved from the search is 820, however when I run the stats (screenshot 1, 2, and 3) the total number is way less than 820. It seems like the stats are not running for all candidates.

[Slack Message](https://refugeejobsmarket.slack.com/archives/C42S4AJ73/p1698245892994139?thread_ts=1698245892.994139&cid=C42S4AJ73)

Notes:

Look at the combination of search criteria. This is quite a complex search including elastic search + status + location. Look in this area to reproduce the issue locally.

samschlicht:
…using an instance of Pageable. Overrides a default that was capping results at 10 candidates and causing the bug.


sadatmalik:
Nice one @samschlicht !

sadatmalik (approved):
Approved!

PR #575, opened 2024-01-20
570 readunread jobchat notifications back end

### LINKED ISSUE #570: Read/Unread - JobChat notifications - Back end
Each chat will have a toggling button "Mark read/unread" (or something equivalent)

Clicking that button will send a message to the server, updating the database recording whether or not the chat has been read by the user.

The data can be stored in a table UserReadChat which stores an entry for each chat read by a user

user/chat/timestamp(or last read post id)

That entry records that the user has indicated that they have read the chat.

When a new post is made on a chat, all entries for that chat will be removed from the table - because the chat now contains new information unread by any user.

On the front end, a new component can be created which monitors and displays the read/unread status of a chat. Instances of this component are used to notify the user of new chat posts that they have not read. It may simply display an asterisk (for example) indicating that the chat has unread posts. This component queries the initial "read status" of the chat (from the UserReadChat table on the server), and then subscribes to the stream of post updates received by subscribing to the chat, merged with a stream of "user read chat" events. The latter stream is generated completely locally from the Angular code (rather than events from the server).
The component subscribes to the stream when it is displayed, an unsubscribes when it is not displayed. A new chat post causes the component to display "unread", a "user read chat" event causes the component to display "read".
 @sadatmalik

_Originally posted by @camerojo in https://github.com/Talent-Catalog/talentcatalog/issues/445#issuecomment-1826731942_

camerojo:


ui/admin-portal/src/app/services/chat.service.ts
```diff
@@ -119,46 +130,131 @@ export class ChatService implements OnDestroy {
...
-    //Set an initial value.
-    userMarkedChatAsRead$.next(isRead);
-
-    let chatReadStatus$ = merge(newPosts$, userMarkedChatAsRead$).pipe(
-      takeUntil(this.destroyStompSubscriptions$)
+    //The read status of a chat is driven by the above two streams.
+    //One which is driven by incoming posts to the chat and the other driven by the user clicking
+    //on the chat's Mark as Read button.
+    //Merge these two boolean streams - indicating the current read status of the chat.
+    let chatReadStatus$ = merge(newPosts$, markedChatAsReadSubject$).pipe(
```
sadatmalik, line 238: Nice!

sadatmalik (approved):
Approved

Issue #567, opened 2024-01-16
Add consent to participate in monitoring & evaluation projects to TC intake


samschlicht:

samschlicht:
Check out the epic for context — but briefly: because the World Bank impact evaluation survey is to include unsuccessful candidates, we need to get their proactive consent, and the research team felt this was best obtained during intake. In part because that offers an opportunity briefly to explain the project, to avoid any confusion and better the chances of the person choosing to participate.

Suggest reviewing #299 which did a very similar thing for the candidate portal, and #402, which synced those new fields to SF.

Here are the broad steps:
- [ ] Add new boolean field to candidate entity (something like, 'monitoring and evaluation consent')
- [ ] Add new tab (to my eyes, this is needed — note to discuss w group) and checkbox to full intake form
- [ ] Add field to SF candidate sync code (SalesforceServiceImpl)
- [x] Add checkbox to SF candidate contact records, prod and sandbox

samschlicht:
Noting that I've contacted Sarah to check if they have a view on how the Q should be phrased on the intake form — will add to this issue anything that comes back. I've advised they be fairly generic with the TC Q, since it would ideally continue to apply to M & E projects after the World Bank one expires.

(They plan to have some accompanying materials ready for any candidates who want to know more — an email containing FAQs and a video of a TBB alum explaining the project, and a script for the source team conducting the intake.)

samschlicht:
@cazcam34 just a note — if you do pick this one up — that I've already added the checkbox to SF staging and prod. I needed it to move ahead with my survey automation. Let me know if you want to revise for any reason, it's not set in stone at this point, of course.

Here are the details:
Object Name: Contact
Data Type: Checkbox
Field Label: Monitoring & Evaluation Consent	
Field Name: Monitoring_Evaluation_Consent
API Name: Monitoring_Evaluation_Consent__c	 	 
Help Text: Indicates consent given at TC intake to be contacted for monitoring and evaluation projects

[...]

PR #577, opened 2024-01-22
create index on candidate_id in candidate_dependant table

sadatmalik:
This PR:

- adds an index on candidate_id in the candidate_dependants table
- this gives a performance improvement on blank searches of ~15x


sadatmalik:
Do not merge before #567 - i.e. flyways 291

camerojo (approved):
Wow - brilliant find and quick win!

PR #611, opened 2024-02-01
205 unit tests for other admin apis part 5

### LINKED ISSUE #205: Unit tests for Other Admin Apis part 5
Unit tests for

- [x] SavedSearchAdminApi
- [x] SavedSearchCandidateAdminApi

samschlicht:
All done - just need to run by the team a method in SavedSearchCandidateAdminApi that doesn't have a path - I started writing a test and then realised!


camerojo:
Hey @samschlicht - yes the method does have a path. If you look at the method it overrides the standard definition defined by IManyToManyApi.

I introduced a number of those standard API interfaces in order to standardize common API methods - which were all over the place in the original rewrite.

So, in this case, the method inherits the standard "search-paged" path.

These standard API methods are pretty well documented in org/tctalent/server/api/admin/ITalentCatalogWebApi.java - and the interfaces that extend it.

Not all controllers have been converted over to extend these standard API's - but eventually they should be (more tech debt!)

camerojo (changes requested):
See my comment. The method you mention does have a path.

samschlicht:
Ah okay great, I'll take a look and resubmit the PR with that test
included. Thanks, @John Cameron ***@***.***>!

samschlicht:
@camerojo I'm getting a 404 error, which I recall now was happening before.

I'm trying: /api/admin/saved-search-candidate/search-paged

samschlicht:
Thanks, @camerojo, I've resubmitted with that test included!

samschlicht (commented):
@camerojo I've amended per your advice

camerojo (approved):
Nice!

PR #616, opened 2024-02-01
206 unit tests for other admin apis part 6

camerojo:


camerojo:
Doing the test on SiteRedirectController exposed a couple of minor issues with that controller which I corrected.

The intention of SiteRedirectController was to redirect urls like this:

https://tctalent.org/backend/jobseeker/view-resume/?id=120764

to this

https://tctalent.org/admin-portal/candidate/120764

It does that successfully - but I doubt that it happens any more.

- There was an unused CandidateService wired into the controller - which I removed
- The controller uses a configuration from application.yml to pick up the base url for redirection. That was pointing to an obsolete Heroku staging- which we never really used - at least for test. I removed it and made that config hard coded (ie not using environment variables which are therefore now not needed) to our production url https://tctalent.org/admin-portal. The purpose of this controller was to redirect urls generated from the original Version 1 of the TC to equivalent urls on our current Version 2 TC. If any redirects are happening they should only go to production. Most likely nobody still has references to those old urls.

PR #620, opened 2024-02-06
HotFix 2.2.0 Fix bug resulting in jobs not being editable

camerojo:
loggedInUser was not being populated


sadatmalik (approved):
Approved

PR #626, opened 2024-02-08
615 candidate status changes from active need to reflect on sf

### LINKED ISSUE #615: Candidate status changes from active need to reflect on SF
With this release we've added TC status to SF candidate contact records. The nightly TC-SF sync only includes 'active' candidates (active, pending, incomplete). 

While the rationale for that made sense at the time, the following has occurred to me...

Currently, changing a candidate case on the TC to an employed stage will also change the candidate's status. This won't reflect on SF, since that candidate is no longer active and so won't be included in the sync.

Likewise, if a previously active candidate is marked 'unreachable' in the TC, that won't be picked up in the SF sync.

So, we're piling up candidates who're actually inactive but will appear to be active in SF.

My suggested fix for this is to just remove the status filter and include all candidate contact records. 

SF candidate stats will just have to be filtered the same way we do on the TC. Having the two databases aligned probably just makes more sense.

samschlicht:


camerojo (approved):
Approved

Issue #642, opened 2024-02-09
Closed cases are inadvertently included for publishing when the submission list is first loaded


sadatmalik:
From Jana:

When I publish a list on TC, the closed cases still show up on the published sheet although I have not checked the "show closed cases" box. 

Is there anything I am doing wrong:eyes: For reference, I only press on _Publish List_, choose the _Fields_ that I want and then click _Publish_

For reference, <https://tctalent.org/admin-portal/list/8138|this is the list> I am working on

[Slack Message](https://refugeejobsmarket.slack.com/archives/C42S4AJ73/p1707488790061359)

Note: This is related to issue #222 - resolving #222 will also rectify this bug.

PR #643, opened 2024-02-09
Hotfix-2.2.0-Fix-issue-642-closed-opps-being-published

camerojo:
Fix bug where Closed cases were being published because checkbox was initializing as null instead of false.

Fixes issue #642


sadatmalik (approved):
Thanks for picking this up John. 

Approved 🙌

PR #674, opened 2024-02-20
Edited email validation regex in all 3 usages to be case insensitive.

### LINKED ISSUE #669: Make email validation case-insensitive?
Our email validation currently doesn't allow an email address with any uppercase letters. This was something I didn't give full consideration to when doing #462.

It might be confusing for candidates since email addresses are actually case-insensitive and the error message doesn't give its reason.

Suggest we change the regex pattern on candidate and user registration to: 
`[a-zA-Z0-9._%+-]+@[a-zA-Z0-9.-]+\.[a-zA-Z]{2,4}$`

(Previously it was `[a-z0-9._%+-]+@[a-z0-9.-]+\.[a-z]{2,4}$`.)

Our portal logins are already case-insensitive, so there's no concern about people mixing up case and getting locked out of their accounts, which is one of the red flags when it comes to implementing this.

- [x] registration-create-account.component.html 
- [x] create-update-user.component.html
- [x] edit-candidate-contact.component.html

samschlicht:


camerojo (approved):

camerojo:
Wondering is there was any way not to have to duplicate that same regex everywhere

PR #694, opened 2024-03-01
684 filtered displays of jobs or cases with unread messages

### LINKED ISSUE #684: Filtered displays of jobs/cases/candidates with unread messages
Showing just opps with messages is the missing thing needed to make chats really usable.

- [x] Any job with message
- [x] Any case with message
- [x] #693

camerojo:
This also improves the detail display of selected jobs - by displaying the normal job display.

### DIFF
--- a/server/src/main/java/org/tctalent/server/repository/db/CandidateOpportunitySpecification.java
+++ b/server/src/main/java/org/tctalent/server/repository/db/CandidateOpportunitySpecification.java
@@ -148,7 +148,7 @@ public class CandidateOpportunitySpecification {
 
 
                 final Predicate oppHasUnreadChats = SpecificationHelper
-                    .getOppHasUnreadChats(loggedInUser, query, builder,
+                    .hasUnreadChats(loggedInUser, query, builder,
                         numberOfChatsToRead, numberOfChatsToReadRoot, chatBelongsToOpp);
 
                 conjunction.getExpressions().add(oppHasUnreadChats);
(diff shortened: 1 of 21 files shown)


server/src/main/java/org/tctalent/server/repository/db/CandidateOpportunitySpecification.java
```diff
@@ -124,6 +128,32 @@ public static Specification<CandidateOpportunity> buildSearchQuery(
                 }
             }
 
+            //UNREAD MESSAGES
+            if (request.getWithUnreadMessages() != null && request.getWithUnreadMessages()) {
```
sadatmalik, line 132:
```suggestion
            if (Boolean.TRUE.equals(request.getWithUnreadMessages()) {
```

server/src/main/java/org/tctalent/server/repository/db/JobSpecification.java
```diff
@@ -153,6 +155,33 @@ public static Specification<SalesforceJobOpp> buildSearchQuery(
                 conjunction.getExpressions().add(published);
             }
 
+            //UNREAD MESSAGES
+            if (request.getWithUnreadMessages() != null && request.getWithUnreadMessages()) {
```
sadatmalik, line 159:
```suggestion
            if (Boolean.TRUE.equals(request.getWithUnreadMessages()) {
```

sadatmalik (approved):
Approved.

server/src/main/java/org/tctalent/server/repository/db/JobSpecification.java
camerojo: Thanks for the suggestion. I didn't merge it in because although it is shorter, I don't think it is clearer. The way it is currently written is consistent with the processing of all other request fields: namely, if the request field is present, process it. In this case, if the Boolean field is present, use its value.

PR #717, opened 2024-03-04
463 fix visibility of chats

### LINKED ISSUE #463: Fix visibility of chats
As a user
I should not see chats before I should

## Visibility examples
- Candidate only sees Candidate Recruiting Chat once they have passed the CVReview stage
- For example, a job creator can only chat with candidates once those candidates have been short listed (ie at the CV review stage).

cazcam34:


ui/admin-portal/src/app/model/candidate-opportunity.ts
camerojo, line 1: This is not much of a function - it is a function to check that one number is greater than or equal to another.

Nobody is going to call this when they can just write

if (selectedNumber >= desiredNumber)

camerojo: Given that you have put in in the candidate-opportunity module, it would be more useful to make it something that takes two CandidateOpportunity stages

camerojo (changes requested):

ui/candidate-portal/src/app/components/profile/view/tab/opps/opp/candidate-opp/candidate-opp.component.ts
```diff
@@ -50,24 +55,28 @@ export class CandidateOppComponent implements OnInit, OnChanges {
...
       jobId: this.selectedOpp?.jobOpp?.id
     }
 
     forkJoin( {
       'sourceChat': this.chatService.getOrCreate(sourceChatRequest),
-      'destinationChat': this.chatService.getOrCreate(destinationChatRequest),
-      'allJobCandidatesChat': this.chatService.getOrCreate(allJobCandidatesChatRequest),
+      'destinationChat': this.showDestinationChat ? this.chatService.getOrCreate(destinationChatRequest) : of(null),
+      'allJobCandidatesChat': this.showAllChat? this.chatService.getOrCreate(allJobCandidatesChatRequest) : of(null),
     }).subscribe(
```
camerojo, line 80: Actually I think that this looks OK.

The other way of doing it would be to construct the Object passed to to forkJoin...

`
    let sourcesObject = {}
    sourcesObject['sourceChat'] = this.chatService...;
    if (this.showDestinationChat) {
      sourcesObject['destinationChat'] = this.chatService...;
    }  

    forkJoin(sourcesObject).subscribe(...

`

But this looks good too. Let's go with what you have.

camerojo (changes requested):

camerojo (approved):

PR #721, opened 2024-03-06
696 mini intake confirm general information intake screen changes only

sadatmalik:
This PR updates the mini-intake general information tab:

- Adds candidate’s age next to their DOB
- Adds candidate gender
- Adds UNRWA registration status
- Allows edits to contact information directly from the mini-intake tab without having to jump back to the general tab


camerojo (approved):
Approved

PR #722, opened 2024-03-06
show “factors preventing. . . etc." only for “yes” to international opps

sadatmalik:
Very tiny PR that adjusts the Interest in International Recruitment:

- Flips the "factors preventing immediate availability" questions
- Displays only for "Yes" to international opportunities

PR amended following team refinement: We will not flip the question but clarify its meaning instead. See PR comments.

### DIFF
--- a/ui/admin-portal/src/app/components/candidates/intake/avail-immediate/avail-immediate.component.html
+++ b/ui/admin-portal/src/app/components/candidates/intake/avail-immediate/avail-immediate.component.html
@@ -40,7 +40,7 @@
     </div>
   </div>
 
-  <div *ngIf="availImmediate === 'No'|| availImmediate == 'Unsure'">
+  <div *ngIf="availImmediate === 'Yes'">
     <div class="mb-3">
       <label class="form-label" for="availImmediateReason">Factors preventing immediate availability for international work.</label>
       <ng-select id="availImmediateReason" [formControlName]="'availImmediateReason'" (clear)="setNoResponse('availImmediateReason')"
--- a/ui/admin-portal/src/app/components/candidates/intake/avail-immediate/avail-immediate.component.html
+++ b/ui/admin-portal/src/app/components/candidates/intake/avail-immediate/avail-immediate.component.html
@@ -19,7 +19,7 @@
 </div>
 <form [formGroup]="form">
   <div class="mb-3">
-    <label class="form-label" for="availImmediate">Are you interested in international employment opportunities now?</label>
+    <label class="form-label" for="availImmediate">Are you available for international employment opportunities now?</label>
     <div class="float-end">
       <app-autosave-status
         [saving]="saving"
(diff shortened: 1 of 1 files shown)


camerojo:
This change does not make sense to me. It also changes the sense of the existing data.

The question is "Are you available ... NOW?" - ie immediately. In the database that is stored (and has always been stored) that the candidate is immediately available for international recruitment. (The relevant Candidate field is availImmediate - which has the values, Yes/No/Unsure)

However, now if they answer YES - we immediately ask them questions assuming that they are NOT immediately available - and record answers about why they are not immediately available.

So on the database with the proposed change we can store the following data for a candidate:

- YES - they are immediately available
- Here are the reasons why they are NOT immediately available

camerojo (changes requested):
This change does not make sense to me. See earlier comments.

sadatmalik:
I had the same initial thought as you when I first saw this request. Discussing with Noura and co. I can see why they're asking for this.

The problem arises from the first question which doesn't actually ask if the candidate is "available" for work now. It asks if they are "interested".

Thus when they are doing the intakes, candidates may answer Yes they are interested right now - but not necessarily immediately available for various reasons. I understood this better in our recent conversation with our alumni in Afghanistan - he's interested in work right now but he isn't immediately available to relocate. If we proceed with the change it captures this kind of scenario.

On the flip-side, with the questions as they currently stand; when a candidate is asked if they are interested, if they reply NO they are then asked about factors preventing immediate availability, which doesn't quite make sense either.

I do see your point though, but I also see where source are coming from in making this request. 

To leave it as it currently is, I think we should at least edit the first question, instead of:

"Are you **interested** in international employment opportunities now?"

Make it:

"Are you **available** for international employment opportunities now?"

Alternatively, if we proceed with the request from source, we should change the first question to:

"Are you interested in international employment opportunities?" (lose the 'now')

And then the follow up question would be:

"Any factors preventing immediate availability for international work?"

camerojo:
> "Are you **available** for international employment opportunities now?"

I think that this makes sense - and it is also the only practical option because of the existing data, because that has been the understanding of the data up until this point. That is clearly reflected in the naming of the fields on the database - "availImmediate". This naming came from the briefing of the source team at the time - who were the people actually conducting the interviews with candidates - we didn't make that up from nowhere - and the "Immediate" part clearly reflects the "now" in the existing question.

I think that changing the question to simply "Are you interested in international employment opportunities?" invalidates the sense of the existing data. It is also a bit of a nothing question. Why on earth would anybody go to the trouble of registering on the Talent Catalog if they weren't interested in international employment opportunities?

So my strong preference would be to adopt your suggestion of inserting the word "available" into the question - and leave the "now". That is perhaps clearer, but also consistent with what I am confident is the the sense of the data that we already have gathered.

sadatmalik:
Ok let's do that. As suggested we can simply change the question (leaving everything else as it currently is) to:

"Are you **available** for international employment opportunities now?"

This is in keeping with the sense of the data itself. And clarifies any potential confusion arising from the current naming. 
Once the changes are in staging we can follow up with the users to confirm to them this is what we are doing and the reasons why, reiterating the history and underlying design of the data.

Thanks for the input @camerojo

camerojo (approved):

PR #729, opened 2024-03-07
Change - Accordion panel title as requested

### LINKED ISSUE #709: Full Intake: Dependents - intake screen changes only

Implement intake screen only changes to Dependents from the relevant slides.

cazcam34:
ONLY the change to the panel (dependants to family members/dependants). The other changes to the dependants questions are removals which need more discussion.

### DIFF
--- a/ui/admin-portal/src/app/components/candidates/view/tab/candidate-intake-tab/candidate-intake-tab.component.html
+++ b/ui/admin-portal/src/app/components/candidates/view/tab/candidate-intake-tab/candidate-intake-tab.component.html
@@ -167,7 +167,7 @@
           <div class="container-fluid gx-0">
             <div class="row">
               <div class="col">
-                <span>Dependants</span>
+                <span>Family Members / Dependents</span>
               </div>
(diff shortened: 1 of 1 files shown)


camerojo (changes requested):
Typo - Dependants

See https://www.easybib.com/guides/grammar-guides/vocabulary/confusing-words/dependant-vs-dependent/

cazcam34:
I just followed the spelling from the intake doc provided. Looking into it also it seems dependent is following american english which is probably where they are coming from, what english whould we be going off? https://www.merriam-webster.com/grammar/spelling-variants-dependent-vs-dependant
Screen Shot 2024-03-12 at 8 13 30 pm

camerojo:
I would change it. The article you quoted said you can use either in the US - but outside US there is a difference.

And in the article it says "To be safe, use 'dependant' as a noun ... and in Britain." It is being used as noun here.

cazcam34:
Ready for a review again

camerojo (approved):

PR #730, opened 2024-03-07
711 full intake host country status intake screen changes only

### LINKED ISSUE #711: Full Intake: Host Country Status - intake screen changes only

Implement intake screen only changes to Host Country Status from the relevant slides.

cazcam34:
Just changing switching two components places, and a question text. All other requests in this section are removals.


camerojo (changes requested):
I think it should be "emigration to a 3rd country", not "immigration to a 3rd country" - although I know that is what they asked for.

Can we just change that?

cazcam34:
Ready for a review again

camerojo (approved):

PR #740, opened 2024-03-09
489 job chat centric candidate and admin messaging

camerojo:

PR #742, opened 2024-03-09
741 switch emails from talentbeyondboundaries.org to talentcatalog.net

### LINKED ISSUE #741: Switch emails from talentbeyondboundaries.org to talentcatalog.net
Should just required changes to application.yml configuration and environment variables in AWS task definitions for Staging and Prod

- [x] Set up new emails alert@talentcatalog.net and tc@talentcatalog.net 

Currently messages to the above talentcatalog.net emails autoforward to john@cameronfoundation.org - but that can be changed.

camerojo:
Note that this PR is based on #740 

Merge that into staging first - and most of the changes will disappear. There are only a couple of very minor file changes here in fact.


sadatmalik (approved):
Approved

PR #766, opened 2024-03-18
Update job after updating information

### LINKED ISSUE #762: Adding a link to Job information - eg a link to a JD - works but does not display
For adding job information you can either upload a file or supply a link. 

If you add a link, it works, but the added link does not display, so the user will think that it hasn't worked.

To make it happen:

- Go to any job display
- Click on the Job Information tab
- Click on Add/Edit link
- Fill in the the name and url of the link and click OK

Nothing displays. But if you refresh the browser, the link appears. So the add was successful but just didn't display

cazcam34:


ui/admin-portal/src/app/components/job/jobs-with-detail/jobs-with-detail.component.ts
```diff
@@ -37,6 +38,9 @@ export class JobsWithDetailComponent extends MainSidePanelBase implements OnInit
    */
   @Input() chatsRead$: BehaviorSubject<boolean>;
 
+  //Pick up reference to child CandidateOppsComponent - so we can call methods on it - see below
```
camerojo: Should this comment refer to JobsComponent

ui/admin-portal/src/app/components/job/view/uploads/view-job-uploads/view-job-uploads.component.ts
```diff
@@ -74,6 +74,7 @@ export class ViewJobUploadsComponent implements OnInit {
     this.saving = true;
     this.jobService.updateJobLink(this.job.id, docType, updateLinkRequest).subscribe(
       job => {
+        this.job = job;
```
camerojo: This shouldn't be necessary, should it? The job will get updated by the emit, which will updated the parent job, which then gets input into this component.

camerojo (changes requested):
Looks good - but let's chat about minor comments

camerojo (approved):
Approved

PR #773, opened 2024-03-20
71 Recruiter only sees candidate names after they have reached CV review stage

camerojo:


server/src/main/java/org/tctalent/server/api/admin/CandidateBuilderSelector.java
```diff
@@ -92,15 +101,37 @@ public DtoBuilder selectBuilder() {
         User user = userService.getLoggedInUser();
         Partner partner = user == null ? null : user.getPartner();
 
+
+        //If partner is a job creator, look up the candidates for whom they can see all properties
+        //(ie candidates who have got past the CV Review stage in any job opps managed by the partner)
+
+        //Start by fetching all the opps associated with the partner
+        List<CandidateOpportunity> partnerOpps = candidateOpportunityService.findJobCreatorPartnerOpps(partner);
+
```
sadatmalik, line 110: Could this potentially become a very large list?

sadatmalik (approved):
Approved.

server/src/main/java/org/tctalent/server/api/admin/CandidateBuilderSelector.java
camerojo: Yes - it could get quite big - but never the whole database or anything like that. 

I could work with pages.

PR #777, opened 2024-03-21
Base searches now added to query as a MUST clause

### LINKED ISSUE #732: Elasticsearch with base search should use AND joins
This was my mistake and if no one objects I'd like to have a go at fixing it! 

The branch with the bug in is not in prod yet — but it is in staging, so obviously this should be done before 2.2.1

samschlicht:


server/src/main/java/org/tctalent/server/service/db/impl/SavedSearchServiceImpl.java
```diff
@@ -889,7 +889,7 @@ private Specification<Candidate> computeQuery(
         return query;
     }
 
-    private BoolQueryBuilder computeElasticQuery(BoolQueryBuilder boolQueryBuilder,
+    private BoolQueryBuilder computeElasticQuery(
```
camerojo, line 892: Why did we remove the builder from the call?

camerojo (commented):
See comment. Recursion can get complicated - so need to understand why the change was needed.,

server/src/main/java/org/tctalent/server/service/db/impl/SavedSearchServiceImpl.java
samschlicht: Because in my previous go I was adding the keyword and filters to my original query one-by-one. That's what caused the issue — the keyword clause was being added as a MUST (which is what we want) but the filters were being combined into a single query with the original (not what we want at all and I can't believe I missed it). Now I'm building a separate query and adding it altogether to the original query as an additional MUST clause.

So, the bit you've commented on, I build the first query. Then for any joins, I compute them as a MUST addition to the first. I'll tag you on that bit.

server/src/main/java/org/tctalent/server/service/db/impl/SavedSearchServiceImpl.java
```diff
@@ -1179,11 +1181,12 @@ private BoolQueryBuilder addElasticQuery(BoolQueryBuilder boolQueryBuilder,
         Set<Candidate> excludeCandidates =
             computeCandidatesExcludedFromSearchCandidateRequest(request);
 
-        // Each recursion, if any, builds on the query
-        boolQueryBuilder = computeElasticQuery(boolQueryBuilder, request,
-            simpleStringQuery, excludeCandidates);
+        // Each recursion, if any, is added to the query as an additional must clause
+        boolQueryBuilder.must(
```
samschlicht, line 1185: @camerojo no need to pass the builder because I'm adding to it like this.

server/src/main/java/org/tctalent/server/service/db/impl/SavedSearchServiceImpl.java
camerojo: OK - sounds good. Thanks for the explanation.

camerojo (approved):

PR #786, opened 2024-03-23
Add help to the referer search field

### LINKED ISSUE #785: Add help to Referer field in search
Referer field is case insensitive and can take wild cards - we need to let user know that.

camerojo:
Added help to referer field - and also reformatted help of Exclusion list - without changing content - just breaking it up into multiple lines to make it more readable.


samschlicht (approved):
Approved without comment!

sadatmalik (approved):
Approved.

PR #787, opened 2024-03-23
784 displayed columns configuration of saved search gets lost when update search is clicked

### LINKED ISSUE #784: Displayed columns configuration of saved search gets lost when Update Search is clicked
Modify the displayed columns of a saved search
Click on Update Search
Refresh the browser
You will see that the displayed columns change is lost.

The problem is on the server in the way that updates are done by creating a new saved search. See SavedSearchServiceImpl.updateSavedSearch TODO comments.

camerojo:
Also retains search description which was being lost on update.


sadatmalik (approved):
Approved.

PR #813, opened 2024-03-29
Update to use ServiceException.

ajt001:
Use ServiceException.

### DIFF
--- a/server/src/main/java/org/tctalent/server/exception/InvalidCredentialsException.java
+++ b/server/src/main/java/org/tctalent/server/exception/InvalidCredentialsException.java
@@ -16,10 +16,15 @@
 
 package org.tctalent.server.exception;
 
-public class InvalidCredentialsException extends RuntimeException {
+public class InvalidCredentialsException extends ServiceException {
 
-    public InvalidCredentialsException(String message) {
-        super(message);
-    }
+  private static final String DEFAULT_MSG = "Invalid credentials provided.";
 
+  public InvalidCredentialsException() {
+    this(DEFAULT_MSG);
+  }
+
+  public InvalidCredentialsException(String message) {
+    super("invalid_credentials", message);
+  }
 }


camerojo:
Hi Andrew! What a pleasant surprise! @ajt001 

Are you interested in contributing to the TC? You would be very welcome!

We could have a chat some time - together with TBB's CTO Sadat Malik.

ajt001:
> Hi Andrew! What a pleasant surprise! @ajt001
> 
> Are you interested in contributing to the TC? You would be very welcome!
> 
> We could have a chat some time - together with TBB's CTO Sadat Malik.

Hi John,
I've had a look a couple of times at the code, but never went as far as doing everything to get it running. 
I still need a database and the AWS keys so maybe someone could share those :) 

If I could contribute then I'd love to. I might need a lot of hand-holding 🔥 . As you might expect I've never had enough time, but I have a little more now being unemployed!
It would be worth a discussion, as I think knowing the flows, architecture etc needs a little insight from the experts!

camerojo:
Let's chat then after Easter. I am in East Melbourne. Could we catch up for a coffee some time?

camerojo:
I can't find your personal email address. Let's line up a catch up on email. Can you drop me an email at john@cameronfoundation.org @ajt001

sadatmalik:
Hi @ajt001 

Great seeing you here! Would love to have a catch up with you and John 😄 

We've a couple of other open issues around exception handling noted in our backlog. . . but I don't mean to jump the gun!

ajt001:
> Hi @ajt001
> 
> Great seeing you here! Would love to have a catch up with you and John 😄
> 
> We've a couple of other open issues around exception handling noted in our backlog. . . but I don't mean to jump the gun!

Hi Sadat, 
I've been doing a small bit of looking through the code around how exceptions are done - I think there is some opportunities. Agree not to jump the gun :) 

Look forward to a catch up with yourself and John. I'm still planning out my schedule, I'll keep in touch.

PR #814, opened 2024-04-01
348 highlight the es search term in the candidate profile

### LINKED ISSUE #348: Highlight the ES search term in the Candidate profile
From Jana Al Mawed:

I would like to suggest the following to enhance our Candidate Search experience:

When using the Elastic Search and looking for an exact term using the double quotes, can the TC highlight where in the candidate's profile this exact term appears? This will save us so much time trying to look for it - I hope my suggestion makes sense, happy to explain further if not.

[Slack Message](https://refugeejobsmarket.slack.com/archives/C42S4AJ73/p1695917919509629?thread_ts=1695917919.509629&cid=C42S4AJ73)

sadatmalik:
This PR:

- adds a new service that parses and propagates ES search terms
- adds a new directive that applies ES search term highlighting to a given element
- applies the directive to the General, Experience, Education, and Additional Info candidate sidebar tabs
- does not apply the directive to Tasks, Notes, and Jobs sidebar tabs as these are not data provided by the candidate

### DIFF
--- a/ui/admin-portal/src/app/components/candidates/view/additional-info/view-candidate-additional-info.component.html
+++ b/ui/admin-portal/src/app/components/candidates/view/additional-info/view-candidate-additional-info.component.html
@@ -25,7 +25,9 @@
   </div>
 
   <div class="card-body">
-    <p>{{candidate?.additionalInfo}}</p>
+    <p [innerHtml]="candidate?.additionalInfo"
+       appHighlightSearch>
+    </p>
   </div>
 
 </div>
--- a/ui/admin-portal/src/app/directives/highlight-search.directive.ts
+++ b/ui/admin-portal/src/app/directives/highlight-search.directive.ts
@@ -1,20 +1,20 @@
-import {Directive, ElementRef, OnDestroy, OnInit} from '@angular/core';
+import {AfterContentChecked, Directive, ElementRef, OnDestroy, OnInit} from '@angular/core';
 import {SearchQueryService} from "../services/search-query.service";
 import {Subject} from "rxjs";
 import {takeUntil} from "rxjs/operators";
 
 /**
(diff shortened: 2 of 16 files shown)


ui/admin-portal/src/app/directives/highlight-search.directive.ts
camerojo, line 1: Nice! I never considered do it this way. I was only thinking about uploading the Elasticsearch results which can give information about where it found hits. But this "front end only" makes sense - parsing the query and then doing an equivalent regex search in the candidate data.  

Although I would have been scared away by the parsing and conversion to regex - well done getting that all sorted out.

Also walking through the text nodes!! A lot of great stuff in here!

Also this is a great example of using @Directive which I now understand a bit better.

camerojo (approved):
Wow!

ui/admin-portal/src/app/components/candidates/view/additional-info/view-candidate-additional-info.component.html
samschlicht, line 1: Some great stuff in here — a really cool feature, superbly executed. Will make some time to review this again tomorrow!

samschlicht, line 1: I'd add that one of the complaints I've heard a couple times about ES is that you can't filter by where the search terms appears — e.g., you specifically need someone with a nursing _degree_ (noting this example because it's not an option in the 'Education major' dropdown but is often searched for) and get results for anyone who has nursing pretty much anywhere in their profile. Might be scope to extend this also to filter or sort results by where the keyword appears?

E.g. in the Keyword Search section you could have a multi-select dropdown to choose the documents to search.

Just a thought, this is extremely useful in itself.

PR #839, opened 2024-04-04
833 add caching and index for users

### LINKED ISSUE #833: Add caching for user lookups
With the index on lowercase username performance is better. But still seeing a lot of hits on the same query:

Going to consider caching for this query to prevent too many database queries. Return from the cache the user info when searching for the same lowercase username.

This is to confirm performance any gains, will subsequently need to consider cache eviction where needed for user updates.

sadatmalik:
This PR:

- adds caching support for frequent database calls
- applies caching to user retrieval by lowercase username


sadatmalik:
Brings a lot of improvement to find user by lowercase username calls:



Next avenues to explore - CandidateOpportuntyRepository opp/check-unread-chats:

camerojo (approved):

PR #855, opened 2024-04-07
Disable scheduled nightly DB copy

### LINKED ISSUE #854: No longer need old scheduled DB Copy
We should disable this scheduled copy.

The ultimate solution is to provide a live anonymised drop copy of the TC database. See #233

camerojo:


sadatmalik (approved):
Approved.

PR #862, opened 2024-04-09
856 process canada help exported from tettra as website to generate associated helplinks

### LINKED ISSUE #856: Process Canada help exported from Tettra as website to generate associated HelpLinks
Instead of manually adding help links, this can be automated by using JSoup to process the HTML.

Can be used as example for processing other help docs that can be exported as HTML

camerojo:
Bit of an odd way of executing these stand alone generators - I just run them as a unit test. Is there a better way of doing this?

Includes Flyway for adding in Canada and Australia help links


sadatmalik (approved):
Approved.

sadatmalik:
> Bit of an odd way of executing these stand alone generators - I just run them as a unit test. Is there a better way of doing this?
> 
> Includes Flyway for adding in Canada and Australia help links

Another way to handle operations intended to be executed on demand is to implement a Spring command-line runner. This can be run manually as required without needing to comment or uncomment annotations.

Here's an example of how you could set this up:

```
import lombok.RequiredArgsConstructor;
import org.springframework.boot.CommandLineRunner;
import org.springframework.stereotype.Component;

@Component
@RequiredArgsConstructor
public class HelpLinkGeneratorRunner implements CommandLineRunner {

    private final HelpLinkGeneratorFromHtml generator;

    @Override
    public void run(String... args) throws Exception {
        if (args.length > 0 && "generate-help-links".equals(args[0])) {
            String url = "https://tchelp.tettra.site/canada-1/operations-manual-1"; // Could also be passed as an argument
            generator.generateCanadaHelpFromHtml(url);
        }
    }
}
```

Then start the service passing `generate-help-links` as a command line argument.

camerojo:
> Another way to handle operations intended to be executed on demand is to implement a Spring command-line runner. This can be run manually as required without needing to comment or uncomment annotations.

Love Spring!!!! Thanks @sadatmalik

PR #920, opened 2024-04-26
919 simplify jobchat code now that both admin and candidate portals have support for translation keys

### LINKED ISSUE #919: Simplify JobChat code now that both admin and candidate portals have support for translation keys
As a developer
I want to be able to write code which displays text using translation keys
because that allows me to have common code which works both on the admin portal and candidate portal

camerojo:


cazcam34 (dismissed):
Looks good to me!

cazcam34:
Ok I've pushed two commits to your branch, one making the html same across portals and the other making the ts same across portals.
The html was just making some extra margin the same, extra margin was added to admin portal to help fix the problem of the emoji picker sitting on top of the header which was fixed by just adding some margin to push the picker element slightly further down. Added to candidate portal is irrelevant as the candidate portal shows the emoji picker as a modal pop up. But the added margin doesn't negatively impact anything on the cand portal, so it's easy to make the same.
The ts was made the same but adding some extra code that was needed for the admin component (again related to the fact that the emoji picker is a dropdown on admin portal, but it is a modal pop up on candidate portal) this code is redundant on the candidate portal, but at least it makes the components the same.
Just so you know the emoji picker appears as a modal pop up on candidate portal as this is best for mobile, but is a dropdown on admin portal as it is best for desktop. This difference occurs in the View Post components code (html & ts) of the portals - so I don't know if these need to be made the same again?

sadatmalik (approved):
Approved.

PR #958, opened 2024-05-06
948 upgrade java version v2

### LINKED ISSUE #948: Upgrade Java version
As a developer
I want to work with the latest Java version
because the latest Spring libraries demand

ajt001:
Minimum changes for JDK upgrade. 
Required Gradle upgrade and Scala upgrade.


camerojo (changes requested):
We also need to the modify the GitHub workflows so that GitHub builds using Java 21 - see .github/workflows

Can you also add that to this PR?

camerojo (approved):
Approved. Thanks Andrew.

PR #961, opened 2024-05-07
protractor-http-mock

### LINKED ISSUE #922: Setup Environment and Dependencies for Functional Testing
- [x] Identify dependencies required for functional testing
- [x] Resolve any configuration issues that may hinder the setup process.
- [x] Configure Protractor to interact with the admin portal's UI elements, such as forms, buttons, and menus.
- [x] Write E2E simple test script using Protractor's syntax to simulate user interactions and verify expected behavior in the admin portal.

EhsanEhrari:
protractor-http-mock,
jasmine-spec-reporter,
types/jasmine
These added to our project dependencies for enhancing our testing capabilities.


ui/admin-portal/e2e/protractor.conf.js
```diff
@@ -16,19 +16,19 @@
...
+const httpMock = require('protractor-http-mock/lib/httpMock');
 
 exports.config = {
   allScriptsTimeout: 11000,
   specs: [
     './src/**/*.e2e-spec.ts'
   ],
   capabilities: {
-    'browserName': 'chrome'
+    'browserName': 'firefox'
```
camerojo, line 28: Why not Chrome? I think that most of us use Chrome. Can we support both?
@EhsanEhrari

camerojo (approved):
Approved - but see my comment about supporting Chrome

ui/admin-portal/e2e/protractor.conf.js
EhsanEhrari: @camerojo 
Chrome's frequent updates sometimes lead to compatibility issues with the latest version of ChromeDriver, making it necessary to downgrade Chrome to run tests smoothly. By including Firefox, we ensure flexibility and reliability in our testing setup. While Firefox serves as the default browser due to its stability and consistent compatibility, we can certainly support both browsers to accommodate various testing scenarios effectively.

So, in summary:
- If Firefox is available, Protractor will use it as the default browser.
- If Firefox is not available, Protractor will attempt to use another browser specified in the configuration, such as Chrome, as a fallback option.

What do you think?

sadatmalik: Hi @EhsanEhrari, and if both browsers are available, can we support both?

EhsanEhrari: Hi @sadatmalik ,

Protractor does support running tests in multiple browsers simultaneously. However, based on my testing experience, I've found that Chrome tends to execute the tests much faster compared to Firefox. Therefore, I will go with the Chrome to ensure tests run quicker and faster. Additionally, if there are specific scenarios where testing in Firefox is necessary or desired, we can certainly configure Protractor to support both browsers as needed.

PR #972, opened 2024-05-10
PR for upgrading email library.

### LINKED ISSUE #951: Upgrade Mail Libraries Version
As a developer
I want to update to the latest mail libraries
because the existing libraries are becoming outdated and unsupported, and the new versions of Spring will require change.

ajt001:
Build file changes that includes the new packages. Upgrade classes that refer to these. 
Tests pass. 
Not certain if/how more testing of emails would work - is there a need to write test cases to validate?


camerojo (approved):
Approved

PR #975, opened 2024-05-10
942 named look ups breaking on first and last name

sadatmalik:
This PR:

- adds elasticsearch (es)  infra deployment scripts + instructions
- adds an es service and implementation for es queries
- uses the es service to find candidates by name (this will be extended in future PRs)
- fixes Java 21 gradle build issues


camerojo (approved):
Approved

PR #997, opened 2024-05-17
Add python venv to gitignore

### LINKED ISSUE #996: Add new python virtual environments to gitignore.
The introduction of new functionality to manage elastic search indices requires python installed and the creation of a virtual environment (within ./infra/elasticsearch)

As a developer
I want to make sure that files from pythons venv are put into gitignore
So the python venv files are not accidentally committed to the repository.

ajt001:
Line to avoid python venv ending up in git.
Also avoids having > 700 files showing as uncommitted and accidently missing adding/committing files that should be in the repo.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,6 +27,8 @@
 
 ### END John Camerons extra changes
 
+# Python virtual environment
+infra/elasticsearch/tc_env*
 
 # User-specific stuff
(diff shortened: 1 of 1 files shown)


sadatmalik (approved):
Approved - thanks!

PR #1004, opened 2024-05-21
Explicitly specify the version of the library as it is no longer pack…

### LINKED ISSUE #1003: Specify thymeleaf version
As a developer
I want the version of thymeleaf-extras-java8time to be specified in the build file
As it is no longer bundled as a spring dependency with springboot 3.

ajt001:
…aged with spring.
The version in 2.7 is the same as this specified version as the library hasn't been updated since 2019.


sadatmalik (approved):
Approved.

PR #1006, opened 2024-05-21
Let spring manage the version of jdom so upgrades are ok.

### LINKED ISSUE #1005: Spring managed jdom version
As a dev
I want the version of jdom to be managed by spring dependencies
So it's always matching the version of spring.

ajt001:
Let spring manage the version and be included in the upgrade process.


sadatmalik (approved):
Approved.

PR #1023, opened 2024-05-23
Fix User Creation Error When Partner Not Selected

### LINKED ISSUE #966: Error accessing unknown fields on JPA entity during active session
When attempting to create a new user without selecting a partner from the dropdown menu, the system throws an error message:

`**org.tctalent.server.util.dto.DtoBuilderException: Error while accessing property 'firstName' on User$HibernateProxy$9fd3g8HE**`
Additionally, even though the error occurs, the user is still registered in the database. Upon further investigation, I found that after selecting a partner, the system does not allow the creation of the user as it already exists in the database due to the incomplete creation process.

**Steps to Reproduce:**

1. Log in as a System Admin.
2. Navigate to the User Creation page.
3. Attempt to create a new user without selecting a partner from the dropdown menu.
4. Observe the error message mentioned above.
5. Note that despite the error, the user is still registered in the database.
6. Try to complete the user creation process by selecting a partner, and observe that the system does not allow the creation due to existing user data in the database.

EhsanEhrari:
**Fixed Both Front-End and Server-Side Code for User Creation When no Partner is specified**
- Ensure the Save button remains inactive if a partner is not specified during user creation.
- Implement server-side validation to throw an exception when no partner is specified, instead of attempting to apply a default partner.

### DIFF
--- a/server/src/main/java/org/tctalent/server/service/db/impl/UserServiceImpl.java
+++ b/server/src/main/java/org/tctalent/server/service/db/impl/UserServiceImpl.java
@@ -236,15 +236,8 @@ public class UserServiceImpl implements UserService {
                 }
             }
         } else {
-            //If user does not already have a source partner, assign one
-            if (currentPartner == null) {
-                if (creatingUser == null) {
-                    //If we do not know who created this user, set up a default partner
-                    newSourcePartner = partnerService.getDefaultSourcePartner();
-                } else {
-                    newSourcePartner = creatingUser.getPartner();
-                }
-            }
+            //Throw an exception if no partner is specified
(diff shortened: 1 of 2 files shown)


samschlicht (commented):
Nice work, @EhsanEhrari! Candidates are also created as users and that's where the PartnerService.getDefaultSourcePartner() method really is useful. 

This change doesn't affect the candidate code and I agree that our admin users just shouldn't be able to create another admin user without a partner specified (it's usually me doing the creation and I always know who the partner is, which is probably why we never discovered this bug 😎). Nice catch. 👍

sadatmalik (approved):
Approved.

PR #1030, opened 2024-05-24
Ensure the comparisons are against like types.

### LINKED ISSUE #1028: JPA specification critiera change

ajt001:
Changes to ensure comparisons use ID rather than the object. This has an impact when the new version of JPA/Hibernate/Spring arrives, so doing this as a small check-in to separate from the rest.


sadatmalik (approved):
Approved. In anticipation of upgrade to Hibernate 6.x which introduces stricter type-checking on nested properties.

PR #1033, opened 2024-05-27
Stop keydown.enter default behaviour for current and future candidate-search inputs

### LINKED ISSUE #1026: Bug: prevent default behaviour opening Keyword Search tooltip when enter pressed in inputs
If you press enter in any candidate-search text or number input field - e.g. 'Min age' - the 'Keyword Search' tooltip opens.

samschlicht:
Also added a small amount of styling to align 'Referrer' tooltip with 'Keyword search': using inverted commas to examplify search queries is confusing, since they themselves can be used to enhance search queries — which is why we utilise <code> formatting to distinguish examples.

Screenshot 2024-05-27 at 4 31 26 PM

👇

Screenshot 2024-05-27 at 4 31 05 PM


camerojo (approved):
Looks like a good solution

PR #1035, opened 2024-05-28
1034 Remove Deprecated Abstract Class

### LINKED ISSUE #1034: RemoveDeprecatedWebsocketConfigurer

ajt001:
https://docs.spring.io/spring-security/site/docs/current/api/org/springframework/security/config/annotation/web/socket/EnableWebSocketSecurity.html

### DIFF
--- a/server/src/main/java/org/tctalent/server/configuration/WebSocketSecurityConfig.java
+++ b/server/src/main/java/org/tctalent/server/configuration/WebSocketSecurityConfig.java
@@ -16,13 +16,15 @@
 
 package org.tctalent.server.configuration;
 
-import static org.springframework.messaging.simp.SimpMessageType.CONNECT;
-import static org.springframework.messaging.simp.SimpMessageType.DISCONNECT;
-import static org.springframework.messaging.simp.SimpMessageType.UNSUBSCRIBE;
-
+import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
-import org.springframework.security.config.annotation.web.messaging.MessageSecurityMetadataSourceRegistry;
-import org.springframework.security.config.annotation.web.socket.AbstractSecurityWebSocketMessageBrokerConfigurer;
+import org.springframework.messaging.Message;
+import org.springframework.security.authorization.AuthorizationManager;
+import org.springframework.security.config.annotation.web.socket.EnableWebSocketSecurity;
+import org.springframework.security.messaging.access.intercept.MessageMatcherDelegatingAuthorizationManager;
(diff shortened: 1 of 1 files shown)


server/src/main/java/org/tctalent/server/configuration/WebSocketSecurityConfig.java
```diff
@@ -32,20 +36,16 @@
...
             .simpSubscribeDestMatchers("/topic/**").authenticated()
             .anyMessage().authenticated();
+        return messages.build();
     }
-
-    @Override
-    protected boolean sameOriginDisabled() {
-        return true;
-    }
-
```
camerojo, line 50: How have you tested that there are no unexpected side effects introduced by dispensing with the deprecated class.

For example, is SameOrigin still disabled?

camerojo (commented):
See comments.

Is there Spring standard migration doc which states that simply removing the deprected class in this way will have no side effects?

server/src/main/java/org/tctalent/server/configuration/WebSocketSecurityConfig.java
ajt001: @camerojo The reference page is this:
https://docs.spring.io/spring-security/reference/servlet/integrations/websocket.html#websocket-sameorigin

My reading of it is that Spring adds extra security to protect cross domain access. 

> By default, Spring Security requires the [CSRF token](https://docs.spring.io/spring-security/reference/features/exploits/csrf.html#csrf) in any CONNECT message type. This ensures that only a site that has access to the CSRF token can connect. Since only the same origin can access the CSRF token, external domains are not allowed to make a connection.

So if you want to allow other domains to access the websocket, we can disable the protection and revert to the legacy Abstract class. 

I wasn't sure why you would want to have external domains connect this way - but given the code exists already to disable the security, I guess there is a reason?

camerojo: Yes - there is a reason. We are specific about the sites that can access. 

It is not clear how to replace the necessary configuration that we currently have (which took a while to get working) using the ill defined alternatives to the deprecated class.

So to be safe for now we will keep the deprecated approach - rather than trying to include it in our Spring Boot 3 upgrade. We can tackle the deprecation later - after we have successfully migrated to Spring Boot 3.

camerojo (changes requested):
See comments - we will not do this now. Let's get the Spring upgrade completed first./

camerojo:
That has been addressed in #1614

PR #1048, opened 2024-05-30
934 automated tc message when a candidate is removed from the submission list

### LINKED ISSUE #934: Automated TC message when a candidate is removed from the submission list

cazcam34:
Posting to:
- Prospect chat (candidate and source partner)
- Recruiting chat (candidate, source partner and destination partner)
- Job Creator Source Partner chat (job creator and source partner)
- 

Is notifying all these chats for a removal the right thing to do? Doubling up on audiences for some (e.g. a candidate gets sent this notification twice in two chats) and source partner gets this notification three times!


server/src/main/java/org/tctalent/server/service/db/impl/CandidateOpportunityServiceImpl.java
```diff
@@ -610,32 +590,37 @@ private CandidateOpportunity updateCandidateOpportunity(
...
-                        autoPostCandidateOppStageChange, jcspChat, userService.getSystemAdminUser());
-
-                    // Publish the chat post
-                    chatPostService.publishChatPost(candidateOppStageChangeChatPost);
+                    // If stage is changing to CLOSED (e.g. removed from submission list) automate removal post.
+                    if (stage.isClosed()) {
+                        publishRemovedFromSubmissionListPosts(opp, stage);
+                    } else {
+                        // If non closing stage change, automate post to JobCreatorSourcePartner chat
+                        // Find the relevant job chat
```
camerojo: Given that you have other nice publish methods,  maybe it would be good to put this non closing stage change  post into its own method. It might make the code look simpler.

camerojo (changes requested):
See comment

server/src/main/java/org/tctalent/server/service/db/impl/CandidateOpportunityServiceImpl.java
cazcam34: I held back as this was Sam's code, but now I see you like the private publish methods I feel ok to put this into it's own private method.

cazcam34: @camerojo made the changes as requested

camerojo (approved):

PR #1049, opened 2024-05-30
Added EntityManager and fixed potential null pointer exception for TC-SF candidate sync

samschlicht:
Very exciting development here that seems to fix my candidate sync issue entirely, and may also work for other looped batch processing headaches.

It's working fantastically in testing and reading the doc and some Stack Overflow chatter this morning, I'm quite satisfied that this is a reasonable solution.

There's a general prohibition against using `EntityManager.clear()` to paper over cracks in poor design. However, as noted [here](https://www.developerfusion.com/article/84945/flush-and-clear-or-mapping-antipatterns/#:~:text=A%20typical%20usage%20scenario%20for%20clear()%20is%20in%20batch%20processing.), for example, batch processing is a common and sensible use-case.

Screenshot 2024-05-30 at 4 54 05 PM

Screenshot 2024-05-30 at 4 53 54 PM


camerojo (approved):
Great work Sam!

PR #1073, opened 2024-06-06
1067 implement job deletion in systemadminapi

### LINKED ISSUE #1067: Implement Job Deletion in SystemAdminApi
As part of our ongoing database management efforts, we need to implement functionality in the SystemAdminApi to handle the deletion of records from the salesforce_job_opp table. 

- Ensure cascading effects on related records are handled appropriately.
- Maintain referential integrity across all linked tables.

EhsanEhrari:
Although everything is being deleted from the database, there remains one row in the SavedList table without a job ID. This row does not get removed, which might indicate another bug that should be addressed in a separate issue. The exact cause of this anomaly is unclear at this point and requires further investigation to understand why it lacks a job ID.

### DIFF
--- a/server/src/main/java/org/tctalent/server/api/admin/SystemAdminApi.java
+++ b/server/src/main/java/org/tctalent/server/api/admin/SystemAdminApi.java
@@ -42,14 +42,20 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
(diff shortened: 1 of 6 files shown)


camerojo (approved):
Nice work Ehsan. 

The other savedLists when a job is created deliberately do not reference the job. Only the special submission list needs to reference the job.

It does not matter if those lists are not deleted when you deleted a job.

PR #1088, opened 2024-06-10
Deleted 3k spreadsheet export limit

### LINKED ISSUE #1040: Large candidate-search spreadsheet exports causing CPU spikes
Turns out one cause of our 100% CPU spikes are spreadsheet exports from candidate search. 

Below is prod monitoring after I requested a 25k export — it pretty much ground to a halt and until I cancelled the export:

We discussed this in team meeting and these are the first steps to tackle this:
- [x] Test in staging to find a reasonable amount to set as a candidate-limit for exports
- [x] Submit a hot fix PR to implement that limit as a temporary measure
- [x] Review the code to understand better what's causing the performance issue
- [x] Implement a more permanent solution

samschlicht:
Big exports are still rather slow and unreliable, but not unsafe.

### DIFF
--- a/ui/admin-portal/src/app/components/candidates/show/show-candidates.component.ts
+++ b/ui/admin-portal/src/app/components/candidates/show/show-candidates.component.ts
@@ -624,12 +624,6 @@ export class ShowCandidatesComponent implements OnInit, OnChanges, OnDestroy {
   }
 
   exportCandidates() {
-    // if (this.results.totalElements > 3000) {
-    //   const csvError: string = "Spreadsheet exports are currently capped at 3,000 candidates — please " +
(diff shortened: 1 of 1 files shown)


camerojo (commented):
@samschlicht Let's chat on our next engineering call - but I am thinking that we should leave the limit in there for now.

samschlicht:
No probs, @camerojo — I've made spike issue #1099 for settling some the questions on this function.

And also noting that currently the limit code is commented out in staging! So we'll need a new PR if we're keeping it.

samschlicht:
Closing this PR to avoid any confusion - will put issue back in product backlog

PR #1103, opened 2024-06-13
1054 stats response times high cpu + memory consumptions

### LINKED ISSUE #1054: Stats response times - high cpu + memory consumptions
**_Note that this went into production as a Hot Fix._**

Note also that there is still a bug to be fixed. That bug has changed the priority of issue #55 which I have moved out of the IceBox into ProductBacklog. That bug is raised in issue #1160

Running stats for this search results in out-of-memory errors:

https://tctalent.org/admin-portal/search/2976

camerojo:
This is already in prod as a hot fix


sadatmalik (approved):
Approved.

PR #1105, opened 2024-06-14
1104 upgrade to java 17

### LINKED ISSUE #1104: Upgrade to Java 17
As a developer 
I want to upgrade to Java 17
So that I can keep the software current.

ajt001:
Upgrade to Java 17.

### DIFF
--- a/server/build.gradle
+++ b/server/build.gradle
@@ -156,9 +156,6 @@ dependencies {
 }
 
 jib {
-    from {
-        image = "eclipse-temurin:21-jre"
-    }
     to {
         if (project.hasProperty("test-tc-system")) {
             image = "231168606641.dkr.ecr.us-east-1.amazonaws.com/test-ecs"
--- a/server/build.gradle
+++ b/server/build.gradle
@@ -39,10 +39,10 @@ plugins {
 
 java {
     toolchain {
-        languageVersion.set(JavaLanguageVersion.of(javaTargetVersion.toString()))
+        languageVersion = JavaLanguageVersion.of(17)
     }
-    sourceCompatibility = javaTargetVersion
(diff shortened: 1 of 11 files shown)


server/build.gradle
```diff
@@ -39,10 +39,10 @@ plugins {
 
 java {
     toolchain {
-        languageVersion.set(JavaLanguageVersion.of('21'))
+        languageVersion.set(JavaLanguageVersion.of(javaTargetVersion.toString()))
```
sadatmalik: @ajt001 Can I check that you have tested that the Gradle build picks up these variables? 

When you switched to using the variable previously, the build failed for John and me, which is why we explicitly set it to 21. If you've tested the build and it works for you, then leave it as it is; otherwise, please use 17 explicitly.

server/build.gradle
```diff
@@ -39,10 +39,10 @@ plugins {
...
     }
-    sourceCompatibility = 21
-    targetCompatibility = 21
+    sourceCompatibility = javaTargetVersion
+    targetCompatibility = javaTargetVersion
```
sadatmalik: As above.

sadatmalik (changes requested):
Left some comments in build.gradle.

server/build.gradle
ajt001: @sadatmalik I always run the tests and builds from the command line (I prefer that than IDE), and had no issues. 
In any case, I've committed in changes to specify directly with JavaVersion.VERSION_17.

I have also added the 4 other build files so they're all consistent.

camerojo (approved):

sadatmalik (approved):
Approved.

PR #1123, opened 2024-06-28
Reaction methods -> non-standard names and API  implements IJoinedTableApi

### LINKED ISSUE #776: Modify reactions admin api and associated service methods per JC advice
From Slack:
Probably still best to implement ITableApi. You are not obliged to override the standard methods.
That way those other standard methods are always there if they need to be used.
I also think that it is confusing to have the methods you added using standard naming but departing from the standard semantics.
createReaction is not really a standard create, as it returns a list. I think that "addReaction" would be a better name.
Similarly updateReaction is confusing. Maybe modifyReactions?

samschlicht:
per JC advice - no other changes, all tested and still working


sadatmalik (approved):
Approved.

PR #1124, opened 2024-06-28
**REVIEW #1123 FIRST** CP emoji and reaction pickers refactoring

samschlicht:
Just some tidying and simplification. Admin Portal and Candidate Portal are implemented slightly differently, and I was able to get rid from the CP of some of the code I had duplicated between them.

Chiefly there's a bit of complex logic on the AP to avoid two pickers being open at once, but that's not an issue on the CP because for mobile view I preferred a centred picker over an overlay. Whenever the overlay is clicked (i.e. outside the picker), it closes. So I don't need any logic to handle more than one picker being open!


sadatmalik (dismissed):
Approved.

sadatmalik (dismissed):
Approved.

sadatmalik (dismissed):
Approved.

samschlicht:
@sadatmalik sorry not sure what happened here - retested and resubmitted!

sadatmalik (approved):
Approved.

PR #1126, opened 2024-06-29
646 create uk visa check

### LINKED ISSUE #646: create UK visa check
Need to continue to roll out the visa checks on the TC. Canada is the latest with the updated 'standardised' style, so use that as the structure.

cazcam34:
Simple UK visa check that just displays the dependants sync field & button (and a read only display of the candidate dependants)


sadatmalik (approved):
Approved.

PR #1127, opened 2024-06-29
Add filter for jobs to only show for appropriate visa check

### LINKED ISSUE #1122: Fix visa job checks filter to display only jobs for the visa destination
Currently all candidate opps appear in the visa job dropdown when selecting a job to associate a visa job check for. 
This means a candidate opp can be signed to multiple countries visas, this is incorrect and also causes issues with the dependants sync component which is being used across visa job checks.

cazcam34:


sadatmalik (approved):
Approved.

PR #1135, opened 2024-07-02
1107 build model jpa test suites

### LINKED ISSUE #1107: Build model JPA test suites
As a developer
I want to create a set of JPA test suites
So I can be more confident in making changes while ensuring code works as expected.

ajt001:
All the work required to run integration tests on the repositories and specifications. Test suites/cases.

Validation of the SQL script to produce a 'clean' database dump file is the most important aspect of ensuring there are no privacy leaks.

Testcontainers is introduced as a repeatable, isolated way to run the tests against the actual schema.

There are no changes to production source files. 

Gradle build script has been modified to enable the integration testing.


camerojo (changes requested):
Wow - a lot of great work here - but, although some of us, including Sadat and myself, are fans of Kotlin and know it a bit - other members of the team aren't.

It is likely that at some time we will start making use of Kotlin, but only after discussion with the whole team and at a time when we feel that those who need to get up to speed on it have the bandwidth. When we do decide to introduce Kotlin, I imagine that we will do so in a fairly small way initially - maybe implementing a small utility or service in Kotlin. Mandating it as the standard way we do all JPA related testing is too big a jump.

Having said that, your approach looks really neat. It would be great if you could do a show and tell for the team on the approach. The singleton instance of the database created in BaseDBIntegrationTest looks really cool - and worth talking through with those of us - like me - who are not familiar with that approach.

I hate to waste all that Kotlin work you have done (perhaps we can keep the branch), but we will need to stick with Java for these tests for now.

ajt001:
@camerojo @sadatmalik 
I'll create a new branch and re-create the basics in Java. 
Having the test cases done is at least a starting point, like documentation of what needs to be done. It should be relatively trivial to convert over to Java. Perhaps it's not bad to see what the two look like, doing the same job. A way to start to see like for like in the two.

PR #1147, opened 2024-07-04
FAO JC: 1095 create candidateprospect chats functionality for source partners

### LINKED ISSUE #1095: Create CandidateProspect chats functionality for source partners and bug fixes
The evolution of this issue can be understood from the comments below.

John and I identified a bug whereby source partners create a CandidateProspect chat automatically by viewing _any_ full-screen candidate profile (including anonymised profiles of candidates they don't manage), but the tab is not visible to _any_ source partner except TBB (irrespective of whether they manage the candidate) — until the candidate has an opportunity at CV Review stage or later, when it is visible to TBB and the managing source partner.

UPDATE: I've also found that the candidate viewing their own profile also auto-creates a CandidateProspect chat, and that they will see the chat from the point when they are added to a submission list (due to the automated message telling them they're a prospect).

The solution to this bug is also an enhancement:

As a source partner
I want to be able to create and view CandidateProspect chats
So that I can correspond with promising candidates _before_ they're associated with an opportunity at CV Review stage or later

As John suggests, I've created a separate issue for blocking these CandidateProspect chats.

samschlicht:


camerojo (approved):
This looks good! Nice one.

PR #1150, opened 2024-07-05
Add fix to stop outdated candidate object replacing updated candidate

### LINKED ISSUE #1137: Bug: intake data disappearing, profile reverting to draft
HOTFIX went in to prod on 15th July at about 3.30pm melbourne time

Hello <@U02NYRN57V5> <@U055YR1QT0A>, hope all is well!

There's a candidate (Behrad (Hassan) Amjadi 244772) that I have completed both the mini and full intake last week (Wed 26th) but all info from both intakes was deleted (except the "Destinations" and "Citizenships" in the mini intake). Also the sentence near the title (intake conducted by _person_ on _date_) is removed but it's still present in the "Notes" (attached is a screenshot for your reference). I noticed that he changed his status from "active" to "draft" (I changed it back to "active"), could that be the reason? And if so, is there a way to recover the data or should I reschedule with him another call? Thank you in advance!

cc: <@U04Q0BPA68H>

[Slack Message](https://refugeejobsmarket.slack.com/archives/C42S4AJ73/p1719942698985459?thread_ts=1719942698.985459&cid=C42S4AJ73)

cazcam34:
As mentioned in the issue, this is a fix in one spot that addresses the particular scenario that was flagged (candidate added an occupation after the intake in an already logged in account). However this problem will occur in other actions on the candidate portal, so I think we can make this PR larger. I only had time to fix this one occurrence.

Note: Includes some import changes but this will just be reverting back to our standardised settings


cazcam34:
I've updated the PR to fix the problem in more locations I think this is occurring in. Where the outdated logged in candidate is being updated and saved, instead of fetching the updated candidate object.

server/src/main/java/org/tctalent/server/service/db/impl/CandidateAttachmentsServiceImpl.java
```diff
@@ -143,19 +143,10 @@ public CandidateAttachment createCandidateAttachment(CreateCandidateAttachmentRe
         User user = authService.getLoggedInUser()
                 .orElseThrow(() -> new InvalidSessionException("Not logged in"));
 
-        Candidate candidate;
         String textExtract;
 
-        // Handle requests coming from the admin portal
-        if (request.getCandidateId() != null) {
-            candidate = candidateRepository.findById(request.getCandidateId())
+        Candidate candidate = candidateRepository.findById(request.getCandidateId())
```
cazcam34: The request always contains the candidateId now, so we can just fetch the candidate from the request and don't need the two seperate methods.

server/src/main/java/org/tctalent/server/service/db/impl/CandidateServiceImpl.java
```diff
@@ -2624,10 +2624,7 @@ public Candidate getCandidateFromRequest(@Nullable Long requestCandidateId) {
                     .orElseThrow(() -> new NoSuchObjectException(Candidate.class, requestCandidateId));
         } else {
             // Coming from Candidate Portal
-            candidate = authService.getLoggedInCandidate();
-            if (candidate == null) {
-                throw new InvalidSessionException("Not logged in");
-            }
+           candidate = getLoggedInCandidate().orElseThrow(() -> new InvalidSessionException("Not logged in"));
```
cazcam34, line 2712: This getLoggedInCanddiate() method was already in the code and basically does what we want, in that it uses the loggedIn object to get an ID which we can then fetch the updated object with. So this is used to replace where we used to have authService.getLoggedInCandidate() which just used the stored candidate object that belonged to the loggedInUser object.

server/src/main/java/org/tctalent/server/service/db/impl/CandidateOccupationServiceImpl.java
```diff
@@ -171,10 +170,11 @@ public List<CandidateOccupation> updateCandidateOccupations(UpdateCandidateOccup
...
                 .orElseThrow(() -> new InvalidSessionException("Not logged in"));
-        Candidate candidate = authService.getLoggedInCandidate();
-        if (candidate == null) {
-            throw new InvalidSessionException("Not logged in");
-        }
+
+        // Fetch updated candidate object from the DB to collect all data updates that may have been made since logging in.
+        Candidate candidate = candidateRepository.findById(user.getCandidate().getId())
+                .orElseThrow(() -> new InvalidSessionException("Not logged in"));
+
```
sadatmalik, line 172: For consistency, can you route this call through the candidate service instead of directly to the candidate repository?

There's a getCandidate(long id) method on CandidateService for this. The implementation sets a few transient fields that will otherwise be missed. Those fields are not especially relevant to what you're doing here, but going through the service is a good pattern for consistency and avoiding future oversights.

sadatmalik (changes requested):
Just leaving one comment - otherwise looks great!

server/src/main/java/org/tctalent/server/service/db/impl/CandidateAttachmentsServiceImpl.java
```diff
@@ -143,19 +143,10 @@ public CandidateAttachment createCandidateAttachment(CreateCandidateAttachmentRe
         User user = authService.getLoggedInUser()
                 .orElseThrow(() -> new InvalidSessionException("Not logged in"));
 
-        Candidate candidate;
         String textExtract;
 
-        // Handle requests coming from the admin portal
-        if (request.getCandidateId() != null) {
```
camerojo, line 149: @cazcam34 
Why remove the null check? Just because you know that everyone populates that attribute now, it doesn't mean that someone might add code in future and forget to populate the field.
Protecting against a null value is safer if it is at all possible that it could be null - because throwing a null pointer exception on the server is not desirable.

Also with hotfixes you try to keep the number of code changes down to a minimum - and this change is not necessary to the hotfix.

camerojo (changes requested):
See comment about removing null check

server/src/main/java/org/tctalent/server/service/db/impl/CandidateAttachmentsServiceImpl.java
cazcam34: Added null check back in and if it fails throw a invalid request exception.

server/src/main/java/org/tctalent/server/service/db/impl/CandidateOccupationServiceImpl.java
cazcam34: No problem I replaced with that candidateService method

server/src/main/java/org/tctalent/server/service/db/impl/CandidateAttachmentsServiceImpl.java
```diff
@@ -145,15 +145,11 @@ public CandidateAttachment createCandidateAttachment(CreateCandidateAttachmentRe
...
-        // Handle requests coming from the admin portal
         if (request.getCandidateId() != null) {
             candidate = candidateRepository.findById(request.getCandidateId())
                     .orElseThrow(() -> new NoSuchObjectException(Candidate.class, request.getCandidateId()));
         } else {
-            candidate = authService.getLoggedInCandidate();
-            if (candidate == null) {
-                throw new InvalidSessionException("Not logged in");
-            }
+            throw new InvalidRequestException("Missing candidate ID");
```
sadatmalik, line 152: I'm not sure about this. The original code has 2 branches. The first, where candidate ID is null, is for the admin portal; the other, I presume, must be for the candidate portal.

If so, I think we need to leave the second branch intact but go to the candidate service instead of the auth service.

So, following the pattern in all your other changes, the code would become:

```
        // Handle requests coming from the admin portal
        if (request.getCandidateId() != null) {
            candidate = candidateRepository.findById(request.getCandidateId())
                    .orElseThrow(() -> new NoSuchObjectException(Candidate.class, request.getCandidateId()));
        } else {
            candidate = candidateService.getLoggedInCandidate()
                    .orElseThrow(() -> new InvalidSessionException("Not logged in"));
        }

```

I hope I'm understanding that right. Can you take one more look before we hotfix this?

sadatmalik (changes requested):
Left another comment @cazcam34 - could you take a look?

server/src/main/java/org/tctalent/server/service/db/impl/CandidateAttachmentsServiceImpl.java
cazcam34: Hi @sadatmalik - in this particular example, a candidateID is always present whether it comes from the candidate portal or the admin portal. However it did used to come seperately hence the two if/else statements for getting a candidate. We used to call the createCandidateAttachment straight from the portal api create() method (and this request didn't contain the candidateID), however that is no longer used and zombie. Instead we call the upload method from the portal api, and then in that upload method we create the request, pass in the candidate Id and then call the createCandidateAttachment method. So that is why this method always has a candidateId.

Should I delete any zombie code I find in this hotfix PR? Or safest to create a new issue?

sadatmalik: Thanks @cazcam34 , best to leave any further cleanup for a separate PR pls. With your above comments, I'm happy with this to go as a Hotfix.

sadatmalik (approved):

sadatmalik (approved):
Approved.

camerojo (approved):
Approved

PR #1154, opened 2024-07-05
1138 implement a mechanism for redis cache flushing

sadatmalik:
This PR:

- Introduces a cache management service in the new services/db/cache package
- Implements a users cache flush mechanism
- Exposes an endpoint on the system admin api to trigger the flush


camerojo (approved):
Nice!

PR #1159, opened 2024-07-05
Script to delete data from the db.

### LINKED ISSUE #1157: Create sanitised database dump file
As a developer
I want to have a sanitised dump file available
So I can work with and test against the database without concern for privacy leaks

Related to #1107

ajt001:
Use this script against existing dump file. It will delete most data other than static data.

If OK with the results, then create a 'new' dump file from these results and commit that file named '`integration-test-dump.sql`' into `./server/src/intTest/db`

### DIFF
--- a/server/src/intTest/db/docker-compose.yml
+++ b/server/src/intTest/db/docker-compose.yml
@@ -1,0 +1,12 @@
+services:
+  local-database:
+    container_name: postgres-common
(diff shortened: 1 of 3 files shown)


EhsanEhrari:
Closing this PR in favor of #2117 . All related issues will be tracked there.

PR #1170, opened 2024-07-10
1162 utils files implement unit tests part 4 final part

### LINKED ISSUE #1162: [Utils Files] Implement Unit tests (Part 4) Final Part
- [x] Year Helper.ts
  - [x] should generate an array of years from 1950 to the current year by default
  - [x] should generate an array of years from the specified start year to the current year
  - [x] should generate an array of years from the specified start year to the specified end year
  - [x] should generate an array of years from the specified start year to the end year with an offset
  - [x] should generate an array of years in reverse order when reverse flag is true
  - [x] should generate an array of years from default start year to the current year with offset
- [x] URL.ts
  - [x] should generate the correct external URL
- [x] String.ts
  - [x] should truncate the string to the specified length and add "..." at the end
  - [x] should return the original string if its length is less than or equal to the specified length
  - [x] should return the original string if the string is empty
  - [x] should handle undefined or null strings
  - [x] should return true for a string containing HTML tags
  - [x] should return false for a string without HTML tags
  - [x] should return false for an empty string
  - [x] should return false for null or undefined inputs
  - [x] should return true for a string containing self-closing HTML tags
- [x] QR.ts
  - [x] should create an instance of QrImage with the given base64 encoding
  - [x] should validate a correct base64 encoding
  - [x] should invalidate an incorrect base64 encoding
  - [x] should invalidate an undefined base64 encoding
- [x] File.ts
  - [x] should call msSaveBlob if navigator.msSaveBlob is available
  - [x] should create a link element and set attributes correctly if msSaveBlob is not available
- [x] Enum.ts
  - [x] should convert an enum to an array of EnumOptions
  - [x] should convert an array of keys to an array of EnumOptions
  - [x] should handle an empty array
  - [x] should return all keys in the enum
  - [x] should return all string values in the enum
  - [x] should return true for valid EnumOption
  - [x] should return false for invalid EnumOption
  - [x] should return true for valid EnumOption array
  - [x] should return false for invalid EnumOption array
  - [x] should return false for non-array input
  - [x] should return false for an array of numbers
  - [x] should return the index of a key in the enum
  - [x] should return -1 for a non-existent key
- [x] Clipboard.ts
  - [x] should create a textarea element and append it to the document body
  - [x] should set the value of the textarea element to the provided string
  - [x] should set the correct styles for the textarea element
  - [x] should focus and select the textarea element
  - [x] should execute the copy command
  - [x] should remove the textarea element from the document body
- [x] DateAdapter.ts
  - [x] should return the oldest possible date for null input
  - [x] should convert a date string to a Date object with only year, month, and day
  - [x] should convert a Date object to a new Date object with only year, month, and day
  - [x] should return a correct date for an input date in different time zones
  - [x] should format the date as "dd MMM yy"
- [x] ExtendDatePipe.ts
  - [x] should transform a Date object using customDefault format
  - [x] should transform a Date object using customMonthYear format
  - [x] should transform a string date using customDefault format
  - [x] should transform a Date object with a different timezone
  - [x] should return null for null input
  - [x] should return null for undefined input
- [x] CustomDateAdapter Class
  - [x] should convert string to NgbDateStruct
  - [x] should create an instance
  - [x] should return null when converting empty string
  - [x] should convert NgbDateStruct to string
  - [x] should return null when converting null NgbDateStruct
- [x] CustomDateAdapter Class
  - [x] should create an instance
  - [x] should parse string to NgbDateStruct
  - [x] should return null when parsing empty string
  - [x] should format NgbDateStruct to string
  - [x] should return null when formatting null NgbDateStruct

EhsanEhrari:


sadatmalik (approved):
Approved.

PR #1179, opened 2024-07-11
1165 pipes implement unit tests

### LINKED ISSUE #1165: [Pipes] Implement Unit tests
- [x] TruncatePipe
  - [x] should create an instance
  - [x] should truncate text to specified length with default suffix
  - [x] should not truncate text shorter than specified length
  - [x] should handle empty input
  - [x] should handle null input
  - [x] should handle undefined input
  - [x] should handle custom suffix
- [x] SafePipe
  - [x] should create an instance
  - [x] should call DomSanitizer.bypassSecurityTrustResourceUrl
  - [x] should return null for empty url
  - [x] should return null for null url
  - [x] should return null for undefined url
- [x] FilterPipe
  - [x] create an instance
  - [x] should return original value when searchText is falsy
  - [x] should filter by full name
  - [x] should filter by candidate number
  - [x] should handle undefined searchText
- [x] CandidatePipe
  - [x] create an instance
  - [x] should return full name when args is "name"
  - [x] should return default candidate name when user is undefined and args is "name"

EhsanEhrari:


sadatmalik (approved):
Approved.

PR #1180, opened 2024-07-11
comment out code for posting on submission list addition

sadatmalik:
This PR:

- applies a hotfix already pushed to master
- comments out automated notification to candidates on addition to submission list
- this will be reviewed with source for inclusion in the next planned release 2.2.2.

PR #1182, opened 2024-07-12
Hotfix increase export limit to 5000

sadatmalik:
This PR:

- Sets an export limit of 5,000
- Has already been applied to master as a hotfix
- Need to merge #1180 first


camerojo (dismissed):
Approved

PR #1188, opened 2024-07-16
1186 models implement unit tests part 2

### LINKED ISSUE #1186: [Models] Implement Unit tests (Part 2)
- [x] Chat
  - [x] should create a Post instance
  - [x] should create a JobChat instance
  - [x] should create a JobChatUserInfo instance
  - [x] should create a ChatPost instance
  - [x] should create a CreateChatRequest instance
- [x] Country
  - [x] should create a Country instance
- [x] Data-Row
  - [x] should create a DataRow instance
- [x] Education Level
  - [x] should create an EducationLevel instance
- [x] Education Major
  - [x] should create an EducationMajor instance
- [x] Help Link
  - [x] should create a HelpLink instance
  - [x] should create a SearchHelpLinkRequest instance
  - [x] should create an UpdateHelpLinkRequest instance
- [x] Industry
  - [x] should create an Industry instance
- [x] Job
  - [x] should create a Job instance
  - [x] should correctly determine if an opportunity is a Job
  - [x] should create a SearchJobRequest instance
  - [x] should create an UpdateJobRequest instance
- [x] Job-OPP-Intake
  - [x] should create a JobOppIntake instance
  - [x] should allow undefined properties
- [x] Job-Prep-Item
  - [x] should return true if submission due date is set
  - [x] should return true if job description (JD) is provided
  - [x] should return false if job description (JD) link is not provided
  - [x] should return true if job summary is provided
  - [x] should return false if job summary is not provided
  - [x] should return true if job opportunity intake (JOI) is provided
  - [x] should return false if employer cost commitment is not provided
  - [x] should return false by default (empty is true)
  - [x] should return false if suggested searches are not provided
- [x] JWT Response
  - [x] should create a valid JwtResponse object
- [x] Language
  - [x] should create a valid Language object
  - [x] should create a valid SystemLanguage object
- [x] Language Level
  - [x] should create a valid LanguageLevel object
- [x] Occupation
  - [x] should create a valid Occupation object
- [x] Opportunity
  - [x] should create a valid Opportunity object
- [x] Partner
  - [x] should return a comma-separated string of country names
  - [x] should return an empty string if sourceCountries is null
  - [x] should return an empty string if sourceCountries is empty array

EhsanEhrari:


ui/admin-portal/src/app/model/chat.spec.ts
```diff
@@ -0,0 +1,94 @@
...
+  it('should create a Post instance', () => {
+    const post: Post = {
+      id: 1,
+      content: 'Sample content',
+    };
+
+    expect(post).toBeTruthy();
+    expect(post.id).toBe(1);
+    expect(post.content).toBe('Sample content');
+  });
```
camerojo, line 30: I don't understand what logic that this is testing.

const post: Post = {
      id: 1,
      content: 'Sample content',
};

Doesn't run any of our logic, does it. It just populates an object which could be any object - and then checks that the fields that you just populated are in fact populated. There is no TC logic being tested here, is there?

EhsanEhrari: @camerojo 

I understand your concern regarding the current test. Models, by their nature, are simple data structures and don't contain much logic. Therefore, testing them typically involves verifying their properties.

The main logic involving models is usually tested in the components or services that use them. The current test ensures that the model is correctly instantiated with the expected properties, which is a necessary step before these models are used in more complex logic within components.

EhsanEhrari: Model tests checks that for example the JobChat object is correctly instantiated with the expected properties. If we change the JobChat model, for example, by renaming a property, changing its type, adding a new property, or removing an existing one, this test will fail. This ensures that any changes to the model structure are caught and reviewed.

EhsanEhrari: However, I agree that we should focus on testing the actual logic in the components (done) or services where these models are utilized. The current test is a foundational check, and the more complex behaviors should be tested in the context where they are applied.

EhsanEhrari: If you have specific scenarios or functionalities in mind that need thorough testing, please let me know, and I will ensure those are covered in our test suite.

camerojo: I am still not sure that we need such simple tests. For example, if someone renames a property or removes an existing one, that will cause a Typescript syntax error and the code won't compile.

If a new property is added, your test won't fail.

camerojo: Happy to approve it though - just a thought.

camerojo (approved):
Approved.

ui/admin-portal/src/app/model/chat.spec.ts
EhsanEhrari: I completely agree with you, Since these models are simple models with no methods or logic, unit tests focusing on its properties won't add much value. However, I can test how this model is used within the larger application context, such as within services or components that manipulate or validate Model instances.

Currently, about 10 models remained, which I plan to implement in a separate ticket. Before proceeding, I would like to get your input: Should I go ahead with writing these tests, or would it be better to stop?

EhsanEhrari: However, I want to create .spec files to maintain consistency in testing structure, I can create basic tests to ensure the model's properties are correctly instantiated.
What do you think?

PR #1192, opened 2024-07-17
1183 change text of registration email so that candidates do not respond

### LINKED ISSUE #1183: Change text of registration email so that candidates do not respond
Candidates are responding to the email they receive when they register. We should tell them that this is a No Reply email.

EhsanEhrari:


sadatmalik (approved):
Approved.

PR #1211, opened 2024-07-25
Change user notification email to noreply@talentcatalog.net

### LINKED ISSUE #1198: Registration email: send from no-reply@ and set up OOO in case of reply
Complementing Ehsan's work on #1183

- [x] Create new NoReply email on Google
- [x] Set up out of office reply
- [x] Update code to use new email for notifications instead of tc@...
- [x] Modify registration notification text to emphasize that there is no point replying.

camerojo:


sadatmalik (approved):
Approved.

PR #1212, opened 2024-07-25
706 mini intake english exams registration process changes

### LINKED ISSUE #706: Mini Intake: English Exams - Registration process changes

Review and implement registration screen changes for English Exams from the relevant slides. 

To clarify, it will only make sense to do this for all languages, not just english exams.

EhsanEhrari:


server/src/main/java/org/tctalent/server/api/portal/CandidateExamPortalApi.java
```diff
@@ -0,0 +1,69 @@
...
+import org.tctalent.server.exception.NoSuchObjectException;
+import org.tctalent.server.model.db.CandidateExam;
+import org.tctalent.server.request.candidate.exam.CreateCandidateExamRequest;
+import org.tctalent.server.request.candidate.exam.UpdateCandidateExamsRequest;
+import org.tctalent.server.service.db.CandidateExamService;
+import org.tctalent.server.util.dto.DtoBuilder;
+
+@RestController
+@RequestMapping("/api/portal/candidate-exam")
+public class CandidateExamPortalApi implements ITableApi<CreateCandidateExamRequest,UpdateCandidateExamsRequest,UpdateCandidateExamsRequest> {
```
camerojo: You have not implemented the correct interface - because CandidateExam is joined to Candidate, the correct interface to implement is IJoinedTableApi - which supports the concept of "parent ids". The parent id is the id of the parent Candidate record.

server/src/main/java/org/tctalent/server/api/portal/CandidateExamPortalApi.java
```diff
@@ -0,0 +1,69 @@
...
+  private final CandidateExamService candidateExamService;
+
+  @Autowired
+  public CandidateExamPortalApi(CandidateExamService candidateExamService) {
+    this.candidateExamService = candidateExamService;
+  }
+
+  @PostMapping("/{candidateId}")
+  public @NotNull Map<String, Object> createCandidateExam(@Valid @RequestBody CreateCandidateExamRequest request, @PathVariable Long candidateId)
+      throws InvalidRequestException, NoSuchObjectException {
```
camerojo: You need to override the relevant interface methods - that is what standardizes the API.

For example, this method should implement the IJoinedTableApi.create method. That defines the standard method name "create" (not "createCandidateExam"), the type of mapping (Post) and defines any path mappings - parentId in this case .

camerojo (changes requested):
See my comments. Let me know if you would like to chat about how the standard ITalentCatalogWebApi interfaces are supposed to be used.

server/src/main/java/org/tctalent/server/api/portal/CandidateExamPortalApi.java
EhsanEhrari: Thanks @camerojo,
I completely understand what you meant, thank you. I correctly used the overridden create method of IJoinedTableApi. Regarding the update method, because I implemented the method in this approach, if the request contains an exam ID, it will update; otherwise, it will create a new record.

If you look at updateCandidateExam in CandidateExamServiceImpl, you will see the implementation.

If I use the overridden update method, it requires an {id} as a path variable, and when a candidate wants to add exams for the first time and the exams are not available, I can't have the exam ID to pass.

I think I have three solutions here:

- Use the overridden update method of IJoinedTableApi but pass an ID of 0 from the front end because it will get the exam ID from the request itself.
- Create a method in IJoinedTableApi to do the same but without requiring an ID as a path variable.
- Use a method with a different name for updating instead of using the overridden method.

What do you think?
If you have time for 5 minutes, we can talk about it.

camerojo: Hi Ehsan

You need to change the declaration of CandidateExamPortalApi to...

public class CandidateExamPortalApi implements **_IJoinedTableApi_**<...

Note also that the first parameter of ITableApi and IJoinedTableApi is the SEARCH request, not the CreateRequest

Then you need to implement the methods defined in the interface.

Here is a little video showing what I mean. Let's chat after you have looked at the video.

https://app.screencastify.com/v3/watch/tXYiafv7aapnqD9HrBHZ

EhsanEhrari: Thanks @camerojo ,
Pushed new changes.

server/src/main/java/org/tctalent/server/api/portal/CandidateExamPortalApi.java
```diff
@@ -0,0 +1,112 @@
...
+      CandidateService candidateService
+      ) {
+    this.candidateExamService = candidateExamService;
+    this.candidateService = candidateService;
+  }
+
+  /**
+   * Creates a new record from the data in the given request.
+   * @param request Request containing details from which the record is created.
+   * @return Created record
```
camerojo, line 42: Missing @param doc for parentId

ui/candidate-portal/src/app/components/common/candidate-exam-form/candidate-exam-form.component.html
```diff
@@ -0,0 +1,66 @@
...
+    <label class="form-label" for="notes">{{ 'REGISTRATION.EXAM.LABEL.NOTES' | translate }}</label>
+    <textarea class="form-control" id="notes" [formControlName]="'notes'"></textarea>
+    <app-form-control-error [control]="form.controls.notes"></app-form-control-error>
+  </div>
+  <div class="text-center mt-4 submit">
+    <button class="btn btn-primary"
+            [disabled]="form.invalid"
+            (click)="save()">
+      {{ 'REGISTRATION.EXAM.BUTTON.ADD' | translate }}
+    </button>
```
camerojo, line 63: I think that this button should be Save - that is how it is on certifications for example.
They are not "adding" a new exam - they are saving (ie updating) an existing exam.

camerojo (changes requested):
Looking very close now @EhsanEhrari . See just one comment about the user interface where a button is saying "Add" when you really want to tell the user to "Save".

server/src/main/java/org/tctalent/server/api/portal/CandidateExamPortalApi.java
EhsanEhrari: This one is available in the doc.
   * @param parentId ID of parent record

ui/candidate-portal/src/app/components/common/candidate-exam-form/candidate-exam-form.component.html
EhsanEhrari: Done

camerojo (approved):
Approved

PR #1218, opened 2024-07-29
989 Employer Access data issues

### LINKED ISSUE #989: Employer Access data issues
Problems when adding new job - nothing loads for employer partners/job creators (non TBB). 

Update: Issues due to data, employer ids not present causing errors. Also issue around employer access job creation, partners need to have a SF account link that is Partner (recruitment). Currently we have most accounts as 'Employer' (all minus two).

camerojo:
Added validations to Angular partner configuration to solve issue.

Also fixed validation on Salesforce prod and test (sandbox)


sadatmalik (approved):
Approved.

PR #1225, opened 2024-07-30
1210 issue with cv download for candidate

### LINKED ISSUE #1210: Issue with CV download for candidate
hi <@U055YR1QT0A> <@U02NYRN57V5> - I’m trying to download this candidate’s CV but I keep getting this error message. Other CVs have worked for me - any idea what’s happening?

[Slack Message](https://refugeejobsmarket.slack.com/archives/C42S4AJ73/p1721422696292509?thread_ts=1721422696.292509&cid=C42S4AJ73)

sadatmalik:
This PR:

- Resolves an issue with PDF generation not working for candidate profiles that have LSEP characters in their job experience descriptions

Also:

- Tidies up a few IntelliJ warnings in related classes
- Improves general logging for CV download requests
- Uses try-with-resources for byte array i/o streams

### DIFF
--- a/server/src/main/java/org/tctalent/server/logging/LogField.java
+++ b/server/src/main/java/org/tctalent/server/logging/LogField.java
@@ -51,7 +51,7 @@ import lombok.RequiredArgsConstructor;
 public enum LogField {
 
   USER_ID("uid", 100),
-  CANDIDATE_NUMBER("cn", 101),
+  CANDIDATE_ID("cid", 101),
   JOB_ID("jid", 102),
   LIST_ID("lid", 103),
   SEARCH_ID("sid", 104),
--- a/server/src/main/java/org/tctalent/server/service/db/util/PdfHelper.java
+++ b/server/src/main/java/org/tctalent/server/service/db/util/PdfHelper.java
@@ -20,6 +20,7 @@ import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
 import java.io.UnsupportedEncodingException;
 import java.util.regex.Pattern;
+import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Value;
@@ -47,6 +48,7 @@ import org.xhtmlrenderer.pdf.ITextRenderer;
  */
 @Service
 @Slf4j
+@RequiredArgsConstructor
 public class PdfHelper {
 
     @Value("${server.url}")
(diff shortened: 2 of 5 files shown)


server/src/main/java/org/tctalent/server/service/db/util/PdfHelper.java
```diff
@@ -71,34 +77,69 @@ public Resource generatePdf(Candidate candidate, Boolean showName, Boolean showC
...
+            xHtml = NULL_BYTE_PATTERN.matcher(xHtml).replaceAll("");
 
             // And finally, we create the PDF:
-            ByteArrayOutputStream outputStream = new ByteArrayOutputStream();
-            renderer.createPDF(outputStream);
-            outputStream.close();
-            return new ByteArrayResource(outputStream.toByteArray());
+            return createPdf(xHtml);
+
+        } catch (Exception e) {
```
sadatmalik, line 85: Recommendation generated by Amazon CodeGuru Reviewer. Leave feedback on this recommendation by replying to the comment or by reacting to the comment using emoji.

**Problem**: While re-throwing the caught exception with modifications , information about the caught exception is being lost, including information about the stack trace of the exception.

**Fix**: If the caught exception object does not contain sensitive information, consider passing it as the &#34;rootCause&#34; or inner exception parameter to the constructor of the new exception before throwing the new exception. (Note that not all exception constructors support inner exceptions. Use a wrapper exception that supports inner exceptions.) 
[Learn more](https://www.ibm.com/support/pages/best-practice-catching-and-re-throwing-java-exceptions)

Suggested remediation:
Consider passing the exception object either as  &#34;rootCause&#34; or as an inner exception parameter, to the constructor of the new exception before throwing new exception. This will help keep stack trace.
```
@@ -90,3 +90,3 @@
 
-           throw new PdfGenerationException(e.getMessage());
+           throw new PdfGenerationException(e.getMessage(), e);
         }

```

server/src/main/java/org/tctalent/server/logging/LogField.java
```diff
@@ -51,7 +51,7 @@
 public enum LogField {
 
   USER_ID("uid", 100),
-  CANDIDATE_NUMBER("cn", 101),
+  CANDIDATE_ID("cid", 101),
```
camerojo, line 54: I note that the JavaDoc is not updated.

However, the long term intention is to eventually convert candidate numbers to numbers and make them the primary key - dispensing with candidate ids altogether. Candidate numbers are all users ever see - candidate ids are for internal use only at the moment. It is often convenient to see candidate numbers in logs - given that is what typically displays. 

Perhaps we allow either to be logged depending on what is most relevant or accessible at the time of logging.

camerojo (approved):
Approved but see comments about JavaDoc and logging ids instead of candidate numbers

server/src/main/java/org/tctalent/server/logging/LogField.java
sadatmalik: Discussed in team meeting. Agreed to use candidate ID for logging in the service code but can easily extend to support candidate number as an additional log field if/when needed. Currently, candidate number is not logged anywhere by the service code.

Separately, a ticket has been opened to follow up on the future intention for using candidate number, which is currently a String.

Lastly, the JavaDoc is tidied up in my next PR connected to issue 1205.

PR #1227, opened 2024-08-01
651 set candidate opportunity relocating dependants from candidate opportunity

### LINKED ISSUE #651: Set candidate opportunity relocating dependants from candidate opportunity
- We're adding the ability to update SF candidate opportunity relocation information from the CA visa assessment (#161) and also expanding and refining the visa tool for other countries (#647, #646) to offer this functionality.
- But Zeynep warns that some countries may never reliably use the visa assessment tool, and in any case, family changes can happen at any time, past the point where the visa tab seems the obvious place to go to update.

Discussing this w John raised the notion that this info could be viewable and modifiable from a tab on the TC candidate case too.

cazcam34:


server/src/main/java/org/tctalent/server/model/db/CandidateOpportunity.java
```diff
@@ -93,6 +97,35 @@ public class CandidateOpportunity extends AbstractOpportunity {
...
+     * It is a string as opposed to a List of ids due to the error: ''Basic' attribute type should not be a container'
+     */
+    private String relocatingDependantIds;
+
+    /**
+     * Get the string of relocating dependant ids and convert to a comma separated list of ids(long).
+     * @return List of candidate dependant ids
+     */
+    public List<Long> getRelocatingDependantIds() {
+        return relocatingDependantIds != null ?
```
camerojo, line 113: This probably should be a Set rather than a List because the same dependant should not ever appear twice.

For example, see AbstractCandidateSource.get and set WatcherUserIds

server/src/main/java/org/tctalent/server/request/candidate/dependant/UpdateRelocatingDependantIds.java
```diff
@@ -0,0 +1,35 @@
...
+ * candidate opportunity.
+ *
+ * @author John Cameron
+ */
+@Getter
+@Setter
+public class UpdateRelocatingDependantIds {
+    @Nullable
+    private List<Long> relocatingDependantIds;
+}
```
camerojo, line 35: Probably best as a Set - so that same dependant can't appear twice

camerojo (approved):
Wow!! Great work - this was very fiddly and had to be done carefully. No wonder it took a while. I am not surprised.

I made a couple of comments about the fact that dependants should really be a Set rather than a List - seeing that the same person can't appear twice as a dependant. Having a Set just would automatically do that check. But I know that it was a list in the existing code - and this job was complicated enough!

Maybe one day we will make it a Set - but not really that important.

PR #1233, opened 2024-08-02
693 do display of all candidates that source has chat with

### LINKED ISSUE #693: Do display of all candidates that source has chat with
As a source partner
I want a display where I can see all candidates that I have opened chats with
so that I can see a notification when there a new posts from any candidate

samschlicht:
Added 'Show unread only' which works but isn't auto-updating (YET!) - other than that, looking pretty good and performing well in local tests.

Screenshot 2024-08-02 at 5 24 12 PM

Screenshot 2024-08-02 at 5 39 26 PM


server/src/main/java/org/tctalent/server/api/admin/CandidateAdminApi.java
```diff
@@ -342,4 +345,35 @@ public String generateToken(@PathVariable("cn") String candidateNumber,
...
+     */
+    @PostMapping("fetch-candidates-with-chat")
+    public Map<String, Object> fetchCandidatesWithChat(
+        @Valid @RequestBody FetchCandidatesWithChatRequest request
+    ) {
+        Page<Candidate> candidates = candidateService.fetchCandidatesWithChat(request);
+        DtoBuilder builder = builderSelector.selectBuilder();
+        return builder.buildPage(candidates);
+    }
+
```
camerojo, line 398: Were these methods just to help you with testing? Or are they intended to be used going forward by System Admins?

server/src/main/java/org/tctalent/server/util/StringHelper.java
```diff
@@ -0,0 +1,42 @@
...
+
+  /**
+   * Takes a string and returns true if it is all digits.
+   * @param str the string to be evaluated
+   * @param strLength the length of that string
+   * @return true if all digits, false if not
+   */
+  public static boolean
+  onlyDigits(String str, int strLength)
+  {
```
camerojo: Why is length passed in as a parameter? When would it be different from str.length()?

Also, I would use a standard method rather than writing our own. Elsewhere in the code we use ApacheCommons StringUtils.isNumeric - see  SavedListServiceImpl

And just one last comment - when we do write our own little utils, they are perfect candidates for a unit test - and perfect candidates for a TDD (Test Driven Development) approach. ie write the test first, then implement something that passes the test. As an example, I recently added new functionality to LocaleHelper.getOffsetDateTime - and started by adding a new method to LocaleHelperTest.getOffsetDateTimeTestFallbackToUTCOnBadTimezone

TDD feels odd - and you worry that it is slowing you down - but in fact it doesn't really. We (including me) should do more of it. I also used a TDD approach recently to getting the new stats going - see, for example, SearchCandidateRequestTest.

ui/admin-portal/src/app/components/chat/view-source-candidates-with-chats/view-candidates-with-chat.component.ts
```diff
@@ -0,0 +1,317 @@
...
+import {BehaviorSubject, forkJoin, Observable, Subscription} from "rxjs";
+import {JobChat, JobChatUserInfo} from "../../../model/chat";
+
+@Component({
+  selector: 'app-view-source-candidates-with-chats',
+  templateUrl: './view-candidates-with-chat.component.html',
+  styleUrls: ['./view-candidates-with-chat.component.scss']
+})
+export class ViewCandidatesWithChatComponent implements OnInit {
+
```
camerojo: This is clearly based on showCandidatesComponent. As you mentioned in our engineering meeting you could not see how you could use showCandidatesComponent - which is fair enough. Maybe there is some refactoring needed to make that component more reusable. So you copied and modified it for now - but it would be helpful to note that in the comments - ie "This is a copied and modified version of showCandidatesComponent. That will be helpful as a heads up to anyone using it (or reviewing it!).

ui/admin-portal/src/app/components/chat/view-source-candidates-with-chats/view-candidates-with-chat.component.ts
```diff
@@ -0,0 +1,317 @@
...
+  @ViewChild("searchFilter") searchFilter: ElementRef;
+
+  error: any;
+  loading: boolean;
+  currentCandidate: Candidate;
+  pageNumber: number = 1;
+  pageSize: number = 25;
+  monitoredTask: Task;
+  sortField = 'id';
```
camerojo: I don't think that this Task related stuff is fully supported, is it? I suspect that it is just code copied from ShowCandidatesComponent but only partially supported in this component. For example monitoredTask is never given a value because assignTask has been removed. 

I think you can have the choice of fully supporting tasks in this component, or not supporting tasks at all (making the component a lot simpler). 

Not surprisingly, there is a LOT of duplicated code in here from ShowCandidatesComponent, which is undesirable.

I think that what you really want is a simplified view candidates - without a lot of the functionality that ShowCandidates has. Perhaps the way to achieve that is to factor out the simpler component, and then have ShowCandidatesComponent subclass that simpler component, adding in all its extra functionality (notably its support for all the icons - eg show tasks, download CVs).

In the meantime, let's have a chat about the best way forward to get an MVP

camerojo (changes requested):
I see the problem you ran into. Let's chat about the best way of getting to an MVP

server/src/main/java/org/tctalent/server/api/admin/CandidateAdminApi.java
samschlicht: @camerojo this is the method that fetches the candidates with active chats who populate the component - I'm not sure I follow the question, they're definitely intended to be used.

server/src/main/java/org/tctalent/server/util/StringHelper.java
samschlicht: The length parameter was just to add some low-hanging extra functionality that I didn't need in this instance but might be useful in others (i.e. you could just test part of a string, instead of the whole thing).

I wasn't aware of the standard method, I'll just use that instead!

100% agreed on TDD - it's how we were taught on our course and also seems to be the bedrock of much of the Agile material I've read. It's top of my agenda for the next release cycle, when hopefully we have a bit more time to get used to a new workflow.

ui/admin-portal/src/app/components/chat/view-source-candidates-with-chats/view-candidates-with-chat.component.ts
samschlicht: I'll definitely add some doc to that effect, that was an oversight 👍

ui/admin-portal/src/app/components/chat/view-source-candidates-with-chats/view-candidates-with-chat.component.ts
samschlicht: Discussed via Slack: I'll get rid of the dropdown menu, the tasks, CV and SF icons to make this a MVP with less replicated code.

Subsequently I'll make a ticket for refactoring of show-candidates and it's counterparts (the list view is another component that replicates lots of this code). This may include adding a simple-show-candidates base class that the others can extend, and making some of these icons and the dropdown menu their own reusable components.

samschlicht (commented):
Implemented JC suggested changes

samschlicht:
@camerojo suggested changes have been implemented - ready for another review 👍

server/src/main/java/org/tctalent/server/repository/db/CandidateRepository.java
```diff
@@ -688,4 +689,175 @@ List<Object[]> countByReferrerOrderByCount(@Param("gender") String gender,
...
+              AND users.partner_id = :partnerId
+              AND (
+                LOWER(users.first_name) LIKE :keyword
+                OR LOWER(users.last_name) LIKE :keyword
+                OR candidate.candidate_number LIKE :keyword
+                )
+            """, nativeQuery = true
+    )
+    List<Long> findCandidatesWithActiveAndUnreadChat(
+        @Param("partnerId") long partnerId,
```
camerojo, line 902: Would be better to name this findCandidate_**Ids**_WithActive...

It is a bit confusing to have two very similarly named "findCandidates..." methods - one which returns candidates and the other returns candidate ids.

camerojo (dismissed):
Approved - but minor renaming of methods suggested - see comment

server/src/main/java/org/tctalent/server/repository/db/CandidateRepository.java
samschlicht: Done 👍

camerojo (dismissed):

camerojo (approved):
Approved

PR #1254, opened 2024-08-12
1252 search by candidate number has become unworkably slow

### LINKED ISSUE #1252: Search by candidate number has become unworkably slow
Naji has stopped using it - and found other ways to locate candidate by candidate numbers.
It used to be quick.

Actually Postgres is pretty quick doing wild card searches on an indexed field - see below. It is possible that doing candidate number search by Elastic search will not improve performance.

For example doing the following Postgres query on the production database from Intellij completes in about 0.25 of a second.

<pre>
select candidate_number from candidate where candidate_number like '123%' limit 10;
</pre>

Searching for 123 on the TC takes just under 6 seconds. Of course there will be extra for the HTTP requests to complete - but it seems unlikely that would add that much delay.

camerojo:
This change does two things:
1. Hopefully speeds up search for candidate by partial matches of things like name, number etc by fetching and returning minimal DTO data
2. Reinstates Postgres implementation of partial candidate number search (using text wildcard % in search). The expectation is that the move to Elastic search was unnecessary because the reason for slowness had nothing to do with the wild card search.

### DIFF
--- a/server/src/main/java/org/tctalent/server/api/admin/CandidateAdminApi.java
+++ b/server/src/main/java/org/tctalent/server/api/admin/CandidateAdminApi.java
@@ -16,6 +16,12 @@
 
 package org.tctalent.server.api.admin;
 
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.List;
+import java.util.Map;
+import javax.servlet.http.HttpServletResponse;
+import javax.validation.Valid;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.tomcat.util.http.fileupload.IOUtils;
 import org.springframework.beans.factory.annotation.Autowired;
@@ -67,13 +73,6 @@ import org.tctalent.server.service.db.SavedSearchService;
 import org.tctalent.server.service.db.UserService;
 import org.tctalent.server.util.dto.DtoBuilder;
 
-import javax.servlet.http.HttpServletResponse;
-import javax.validation.Valid;
-import java.io.IOException;
-import java.io.InputStream;
-import java.util.List;
-import java.util.Map;
-
 @RestController()
 @RequestMapping("/api/admin/candidate")
 @Slf4j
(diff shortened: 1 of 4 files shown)


sadatmalik:
The use of minimal DTO data is a good strategy for reducing the payload size and, hopefully, improving the speed of data retrieval.

However, I have concerns about reverting to the wildcard search with PostgreSQL. The root cause of the original performance issue was related to the inefficiency of wildcard searches as the number of candidates grew. Specifically, these wildcard searches weren't taking advantage of the column indexing, leading to slower queries over time.

Since we switched to Elasticsearch, we've had a significant improvement in search-by-number performance in the staging environment. The expectation is that we will observe similar gains in production once the necessary Elasticsearch index rebuild is completed with the next release, which is why this change hasn't been hotfixed yet.

Reintroducing the wildcard search with PostgreSQL could reintroduce the same performance bottlenecks we've seen before, especially as the candidate dataset grows. I would recommend proceeding with the Elasticsearch approach for now, but with the addition of the minimal DTO, which should hopefully further improve performance gains.

camerojo:
Looking into it - and testing on the production database (by executing read-only queries) the wild card PostgresSQL queries seemed to be making use of the index on candidateNumber - and were very fast - see the figures in the issue. With an in memory sorted index of candidate numbers, you can imagine how it could be every efficient to use the index to compute matches of the form xxx%

I think that the performance issues were database related - but not related to the search, but rather to the fetching of the data to be returned.

Can we push into staging just to see? On the larger production database - the actual wild card Postgres search takes less than 0.25 seconds.

sadatmalik:
Thanks for the thorough analysis, based on which I agree with your conclusions.

Initially, I believed that the candidate name search, which uses a double wildcard (%xxx%), avoids the index, leading to slower performance. Elasticsearch, which we already use, seemed like a good alternative. I then opted to reuse the same Elasticsearch-based solution for the candidate_number queries.

However, I see how the trailing wildcard in the candidate_number search (xxx%) could still be able to take advantage of indexing. Interestingly, like yourself, after re-testing the double wildcard scenario for candidate names, I'm also seeing good performance there, which suggests the index might be utilised even in that scenario. This is quite interesting, and I'm not sure how that's working, but it's promising.

```
select u.first_name, u.last_name from candidate
inner join users u on candidate.user_id = u.id
where u.first_name || u.last_name like '%malik%'
limit 10;
```

Given these observations, I agree with your suggestion. Let's go ahead and push this to staging. It offers a simpler solution that can potentially be promoted as a hotfix without needing an ES rebuild. Nice one.

sadatmalik (approved):
Approved.

Issue #1245, opened 2024-08-07
Java ChatPost entity setLinkPreviews throws null pointer exception


camerojo:
setLinkPreviews method needs to handle being passed a null value.

It also doesn't clear existing linkPreviews - so they will accumulate.

samschlicht:
@camerojo I found this one earlier this week and addressed it [here](https://github.com/Talent-Catalog/talentcatalog/pull/1236/files#diff-65f26c75b24531f5477e197c9779d60718f99b614d615062de16fefa6e1bef1d) - but that one hasn't been merged yet. Would you mind taking a look there. I opted to initiate an empty list but perhaps you'd advise a null check on the setter instead (or as well as)?

What do you mean by clearing link previews? I do save the scraped version in the DB so that if someone were to load a Chat thread with a lot of links in, then they're not scraping a dozen or more links at once, which would be messy UX - that seems to be the way the big services do it.

camerojo:
This is a different bug @samschlicht .

The bug is in the ChatPost entity, not Post.

<pre>
  public void setLinkPreviews(List<LinkPreview> linkPreviews) {
         //TODO JC This still has problems because it doesn't clear out existing linkPreviews
         if (linkPreviews != null) {
             linkPreviews.forEach(linkPreview -> {
                 linkPreview.setChatPost(this);
                 this.linkPreviews.add(linkPreview);
             });
         }
     } 
</pre>

I put the null check in just so that I could get my test working.

Note that there are two objects called "linkPreviews" - one is the property of ChatPost and the other is the parameter to the setLinkPreviews method. The parameter can be null (in fact was in the test I was trying to run) - without the added check, it crashes.

Also, in the set method, nothing removes the existing content of the linkPreviews property. So if there 2 link previews and the set method is called 10 times, you will end up with 20 link previews in the linkPreviews collection - 10 duplicates of the 2 actual link previews.

samschlicht:
@camerojo by initialising an empty list in Post, which as I understand it is always the precursor for a ChatPost, the setLinkPreviews method in ChatPost doesn't throw a null pointer exception. But I suppose the null check is still worth having in case we change the way that Chat Posts are created. Sounds like you've included that in your PR, so that's great.

On the second point, why would the set method be called 10 times? It's called once per post. I'm testing it now and 3 link previews in a Post = three saved link previews in the link_preview DB table and on the corresponding Chat Post.

Am I missing something about the way the Chat Posts are created?

[...]

PR #1257, opened 2024-08-12
Reaction + LinkPreview entity doc, small template change on sent posts, proper JPA implementation for Chat Post Link Previews

### LINKED ISSUE #1193: Doc for Reaction and LinkPreview entities and small logging change
John rightly pointed out that these entities could do with a bit of explanatory doc.

Also an opportunity to tidy up the overview doc for LinkPreviewService, which contains a typo. It also may be useful to add some doc to the implementation, since that's where the private scraping methods are.

UPDATE: While I'm at it, will just stop the service methods from logging the whole exception stack traces — it's unnecessary and makes the overall logs harder to read. Use e.getMessage() instead, which provides more than enough in this case.

samschlicht:
NB: combines fixes for #1193 and #1245 - they're both pretty small and mostly concern the same files. I think this is still manageable to review and avoids unnecessary conflict resolution. Let me know if not and I'll split them!

- Adds entity doc for Reaction and Link Preview
- Properly implements bidirectional one-to-many relationship for Chat Post > Link Preview, including null check which is the bug that @camerojo pointed out
- Small template change on sent posts to place reactions below link previews:

Screenshot 2024-08-12 at 3 12 16 PM


server/src/main/java/org/tctalent/server/model/db/LinkPreview.java
```diff
@@ -17,44 +17,133 @@
...
 @Setter
 @Entity
 @Table(name = "link_preview")
 @SequenceGenerator(name = "seq_gen", sequenceName = "link_preview_id_seq", allocationSize = 1)
 public class LinkPreview extends AbstractDomainObject<Long> {
 
+    @Id
+    @GeneratedValue
+    private Long id;
+
```
camerojo: No need to add this - it is inherited from AbstractDomainObject<Long>.

Why did you feel you needed to add it?

server/src/main/java/org/tctalent/server/model/db/LinkPreview.java
```diff
@@ -17,44 +17,133 @@
...
+     */
     @Nullable
     private String faviconUrl;
-}
+
+    /**
+     * We override hashCode and equals here to make sure that two entities with the same important
+     * data (like the same ID) are treated as the same, even if they are different objects in memory.
+     * This is particularly important with one-to-many relationships because the child entities
+     * might be used in a set or map, which check for equality; or removal from the parent entity's
```
camerojo: No need to add all this - it is inherited from AbstractDomainObject

camerojo (changes requested):
See comments - inheriting from AbstractDomainObject gives you a lot of what you need - so you don't have to do it.

camerojo (approved):
Approved!

PR #1258, opened 2024-08-12
1205 evict cached user on save candidate

sadatmalik:
This PR:

- Creates a cache eviction repository interface to manage user cache eviction
- Extends that interface in JPA repositories that should evict the user cache when their entity types are changed
- Includes 58 passing unit tests for all possible scenarios that require user cache eviction


camerojo (approved):
Clever!

PR #1260, opened 2024-08-12
add job opp ID and case ID fields to log builder

sadatmalik:
This PR:

- adds 2 additional fields to LogBuilder: jobOppId and caseId


camerojo (approved):
Love this new logger!

PR #1262, opened 2024-08-13
Add optional minimalData option to search requests.

### LINKED ISSUE #1261: Loading of Candidate screen is becoming very slow
The problem may be the number of database accesses required to retrieve the candidate data. This is sometimes referred to as the N+1 problem.

See for example https://medium.com/javarevisited/spring-jpa-when-to-use-join-fetch-a6cec898c4c6

camerojo:
Implemented Saved List fetches from the full candidate display where we only need saved list names and ids. This not only reduces the data sent in the response request but also avoids fetching a lot of unnecessary data and additional  database queries.


server/src/main/java/org/tctalent/server/api/admin/SavedListAdminApi.java
```diff
@@ -58,17 +57,14 @@ public class SavedListAdminApi implements
     private final CandidateService candidateService;
     private final SavedListService savedListService;
     private final CandidateSavedListService candidateSavedListService;
-    private final SalesforceService salesforceService;
```
camerojo, line 61: This was an unused service (highlighted by Intellij) - pulled it out.

sadatmalik (approved):
Brilliant!

PR #1266, opened 2024-08-14
SF Candidate Opp relocating dependants auto-updated when stage == 'Relocated'

### LINKED ISSUE #666: Auto-update SF case relocation info based on visa assessment

samschlicht:


server/src/main/java/org/tctalent/server/service/db/impl/SalesforceServiceImpl.java
```diff
@@ -411,6 +411,25 @@ public void createOrUpdateCandidateOpportunities(
...
+                if (relocationInfo == null && stage == CandidateOpportunityStage.relocated) {
+                    Optional<CandidateOpportunity> candidateOpp =
+                        jobOpportunity.getCandidateOpportunities()
+                            .stream()
+                            .filter(opp -> opp.getCandidate().getId().equals(candidate.getId()))
+                            .findFirst();
+                    if (candidateOpp.isPresent()) {
+                        relocationInfo = processSfCaseRelocationInfo(candidateOpp.get(), candidate);
+                    }
+                }
```
camerojo, line 432: It seems like there should be an easier way to do this. Wouldn't it be easier to retrieve the required TC candidate from the TC JobOpp - matching candidate.sfId with the opp.getCandidate().getId().

Then just fetch the SF candidateOpp using findCandidateOpportunity.

camerojo (changes requested):
See comment.

server/src/main/java/org/tctalent/server/service/db/impl/SalesforceServiceImpl.java
samschlicht: @camerojo the problem is that processSfCaseRelocationInfo requires a Candidate Opportunity (that's where it gets the relocatingDependantIds), whereas findCandidateOpportunity returns an Opportunity.

I already have the candidate object, the tricky part is getting the candidate opp. Usually I would use CandidateOpportunityService but, as I mention in the doc, I can't inject it here because it creates a circular dependency between services.

camerojo: Ahhh - gotcha

camerojo (approved):
Approved

PR #1274, opened 2024-08-15
SF opp title bug

### LINKED ISSUE #1271: Bug: Candidate Opp title string too long, preventing SF record creation
An interesting little bug here: I can instantly see the cause but will have to think about how we fix it.

Yousef is adding candidates to the submission list for [this Job](https://tctalent.org/admin-portal/job/1032) but the corresponding SF Candidate Opps aren't being created. Looking at our logs, it's clear the Job has a very long title to begin with — subsequently, when the candidate name and number are prepended per our naming convention, the resulting string is too long for SF to accept it as a Candidate Opp title:

Reported by Yousef on the #tool-salesforce channel:
[Slack Message](https://refugeejobsmarket.slack.com/archives/C7QK72T4J/p1723712431425249?thread_ts=1723712431.425249&cid=C7QK72T4J)

samschlicht:
Anything over 120 characters, I truncate to 117 and append '...' to indicate missing text.

### DIFF
--- a/server/src/main/java/org/tctalent/server/service/db/impl/SalesforceServiceImpl.java
+++ b/server/src/main/java/org/tctalent/server/service/db/impl/SalesforceServiceImpl.java
@@ -599,8 +599,11 @@ public class SalesforceServiceImpl implements SalesforceService, InitializingBea
 
     public String generateCandidateOppName(
         @NonNull Candidate candidate, @NonNull SalesforceJobOpp jobOpp) {
-        return candidate.getUser().getFirstName()  +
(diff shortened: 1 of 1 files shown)


camerojo (changes requested):
I suggest we change this issue to do extra validation on the max lengths on different name parts.

samschlicht:
@camerojo the other part of the name is currently mostly set in SF where you can't limit characters. In #1756 I've suggested putting a character limit on employer access job naming, which presumably may become the norm - but I still think we need this failsafe. Let me know if I'm missing something here!

camerojo (approved):

PR #1285, opened 2024-08-18
1284 tune the summary candidate results lines used to display search results and list contents

### LINKED ISSUE #1284: Tune the summary candidate results lines used to display search results and list contents
Currently full candidate info is fetched from the database for each returned result - whereas only summary is used to display the result line.

Note also that the detail of the summary info is not fixed but configurable based on the columns selected for display.

See #1286 for a potential next stage of tuning

camerojo:


camerojo:
This was merged into #1055

Issue #1286, opened 2024-08-18
Tuning, stage 2, of candidate results - projecting to summary DTO's instead of entities  


camerojo:
Performance of loading candidate information in response an API request for a set of candidates is dependant on the number of database accesses required to fetch the required data.

This issue is based on issue #1284 (ie it contains all the changes there). Issue #1284 reduced the number of accesses mainly by not fetching associated candidate related data if it was not required in the front end display of the summary candidate display (ie displaying the candidate in a single line in the list of results). This is because each type of candidate related table data (ie data store in a table other than the candidate table) typically requires one or more additional database accesses to fetch it.

This issue seeks performance improvements by not loading candidate entities from the database, but instead loading a special CandidateSummary object (DTO - Data Transfer Object).

This has the potential for further performance improvement by:
- gathering all summary data in a single database access - by computing (in the database query) the potentially displayed data that is dependent on associated tables without actually needing to load those tables (Instead of joins fetch calculated fields from other tables in subselect - eg select id, (select x from other-table o where o.candidate_id=id) as xfield from candidate.
- loading into non entity DTOs - eg CandidateSummary. Loading data into entity objects when you only need the data for readonly purposes is unnecessary and inefficient. See https://thorben-janssen.com/entities-dtos-use-projection/ 

This will make use of the the new SearchCandidateRequest.extractSQL method which identifies the candidates to be returned.

Note that this DTO approach allows us to directly return fully documented Java classes from our API, rather than Maps and Lists of Maps. For example we can return List\<CandidateSummary\>. See #97

Note also that the filtering of data, based on what the logged in user can see can now be applied BEFORE the database fetching instead of after in the DtoBuilder. This is more logical and efficient since we do not fetch data we are not going to use.

camerojo:
### Trial the summary approach
Trial using the example of tasks. Candidate summary display can display following summary task information:
- total number of tasks
- number of completed tasks
- number of overdue tasks
- number of abandoned tasks

- [ ] Compute all above values in database query
- [ ] Build CandidateSummary DTO which contains candidate summary data including the above computed data
- [ ] Fetch that CandidateSummary DTO in a single database access, constructing the query in SQL based on the fetch SQL returned by the new SearchCandidateRequest.extractSQL method.

cazcam34:
[...]

Issue #97, opened 2023-06-15
Refactor DtoBuilder


sadatmalik:
As a developer
I would like to use Jackson serializers
So that I can factor out the DtoBuilder

sadatmalik:
```
@RestController()
@RequestMapping("/api/admin/branding")
@Slf4j
@RequiredArgsConstructor
public class BrandingAdminApi {

    private final BrandingService brandingService;

    // skeleton code - final cut would not publish entity
    // will implement and use explicit DTO classes with mappers
    @GetMapping()
    public ResponseEntity<BrandingInfo> getBrandingInfo() {
        BrandingInfo info = brandingService.getBrandingInfo(null);
        return ResponseEntity.ok(info);
    }

}

```



```
@Getter
@Setter
@ToString
@Builder
public class BrandingInfo {

    /**
     * Optional landing page associated with branding - eg TBB redirects to
     * https://www.talentbeyondboundaries.org/talentcatalog/.
     * <p/>
     * If no landing page is configured, the default is to automatically go to candidate-portal
     */
    @Nullable
    private String landingPage;

    /**
     * Optional path to logo to display with branding.
     */
    @Nullable
    private String logo;

    /**
     * Partner name associated with branding, if known. Null if unknown.
     */
    @Nullable
    private String partnerName;

    /**
     * Website associated with logo
     */
    @Nullable
    private String websiteUrl;

}
```


```
@Service
@RequiredArgsConstructor
public class BrandingServiceImpl implements BrandingService {
    private final PartnerService partnerService;
    private final UserService userService;

    /**
     * Returns the branding information for a partner.
     * </p>
     * If a user is logged-in then get the partner associated with that user otherwise get the partner associated with
     * the specified partner abbreviation.
     * </p>
     * If no partner is found gets the default source partner.
     *
     * @param partnerAbbreviation Optional partner abbreviation
     * @return branding info
     */
    @Override
    @NonNull
    public BrandingInfo getBrandingInfo(@Nullable String partnerAbbreviation) {

        User user = userService.getLoggedInUser();

        Partner partner = user != null
                ? user.getPartner()
                : partnerService.getPartnerFromAbbreviation(partnerAbbreviation);

        if (partner == null) {
            partner = partnerService.getDefaultSourcePartner();
        }

        return extractBrandingInfoFromPartner(partner);
    }

    private @NotNull BrandingInfo extractBrandingInfoFromPartner(@NonNull Partner partner) {
[...]

PR #1287, opened 2024-08-19
1055 adds a preview dto type

sadatmalik:
This PR:

- Adds a "preview" DTO type alongside the "full" and "minimal" DTO types
- Uses the preview DTO type for browsing candidate "previews" in searches and lists to reduce JSON payloads


server/src/main/java/org/tctalent/server/api/admin/SavedSearchAdminApi.java
```diff
@@ -178,7 +178,7 @@ public void clearSelection(@PathVariable("id") long id,
      * @param request Request containing details of the list to be saved to,
      *                the associated user and whether or not the save should
      *                add to or replace existing contents.
-     * @return List which selection was saved to
+     * @return Map which selection was saved to
      * @throws EntityExistsException If a new list needs to be created but the
```
camerojo: List is correct - it returns the details of the target list that the selection was made to (encoded as a Map like most of our API called - but it is and encoding of the Target List)

server/src/main/java/org/tctalent/server/api/admin/CandidateBuilderSelector.java
```diff
@@ -190,29 +187,37 @@ private DtoBuilder candidateDto(
...
+                    .add("fullIntakeCompletedBy", userDto(userPropertyFilter))
+                    .add("contextNote")
+                    .add("shareableNotes")
+                    .add("additionalInfo")
+                    .add("candidateMessage")
+                ;
+            }
+
+            return builder;
     }
```
camerojo, line 221: We both ended up tackling this one!

In fact the Preview of a candidate is more complex than this. There are some other of those related entities that have to be sent up with the preview because the data is used to display something in the summary line.
eg taskAssignments, candidateExams

See my issue #1284 , also in as a PR.

camerojo (changes requested):
I like the addition of the dto type, but we both tackled the Candidate preview DTO in separate issues.

I think we can fairly easily merge the best of both #1055 and #1284.

But also see a new issue #1286 (for next release not this one) that I raised and started a little over the weekend just to test its feasibility. I think that is the next step in this performance tuning journey - and which takes us in a desired direction which you flagged very soon after joining - ie getting away from the whole Map<String, Object> things in our APIs - you raise it in issue #97

server/src/main/java/org/tctalent/server/api/admin/SavedSearchAdminApi.java
sadatmalik: Ah - right! I've changed that back to List.

server/src/main/java/org/tctalent/server/api/admin/CandidateBuilderSelector.java
sadatmalik: Here I've merged changes from #1284 into this PR #1055. taskAssignment and candidateExams will be sent in previews; but also I think we'll need candidateReviewStatusItems for reviewable list previews - so leaving that in the preview DTO's as well.

camerojo: CandidateReviewStatus items are only related to searches, not lists.
And even for searches they are optional. 

Adding them in has a bit of an impact:
~20 more dB accesses and 10K payload.

In production only 13% of user generated searches are reviewable - the default is not reviewable.

However, I think it is not worth the extra complexity now - so agree with you that it should go into the preview.

We will get our big performance boost in this area when we move to stage 2 tuning.

server/src/main/java/org/tctalent/server/api/admin/CandidateBuilderSelector.java
```diff
@@ -354,19 +359,26 @@ private DtoBuilder examsDto() {
...
             .add("fileType")
             .add("migrated")
             .add("cv")
-            .add("createdBy", userDto(userPropertyFilter))
             .add("createdDate")
             .add("uploadType")
             .add("url")
             ;
+
+        if (DtoType.PREVIEW.equals(type)) {
```
camerojo: This should be:

```suggestion
        if (!DtoType.PREVIEW.equals(type)) {
```

Found this by comparing performance stats between my version and merged version - was a biggish discrepancy so I knew something was different. Found this!

camerojo (changes requested):
One very small typo/change that was spoiling performance

camerojo (approved):

PR #1289, opened 2024-08-20
1232 display candidate exam results in admin portal candidate profile

### LINKED ISSUE #1232: Display candidate Exam results in admin portal candidate profile
This is related to #706 which added Candidate exam results to the candidate portal registration process.

On the admin portal, candidate exam results only appear on the Mini Intake - they should also appear on the Candidate's profile page - probably under the Education tab.

EhsanEhrari:


server/src/main/java/org/tctalent/server/api/admin/CandidateExamAdminApi.java
```diff
@@ -57,6 +62,38 @@ public class CandidateExamAdminApi implements IJoinedTableApi<CreateCandidateExa
         return candidateExamDto().build(candidateExam);
     }
 
+    /**
+     * Get all records joined to the given parent
+     * @param parentId ID of parent record
+     * @return All records associated with the given parent
+     */
```
camerojo, line 69: This is duplicate documentation from the interface. If the interface is documented you shouldn't just duplicate it.
Normally when documenting these overriden interface methods, we change the generic documentation to match the particular class. So parentId becomes candidateId and text referring to "records" becomes CandidateExams, parent becomes Candidate.

That will be much more useful to other programmers.

server/src/main/java/org/tctalent/server/service/db/CandidateExamService.java
```diff
@@ -46,4 +46,7 @@ CandidateExam createExam(
      * @return List of updated candidate exam records
      */
     CandidateExam updateCandidateExam(UpdateCandidateExamRequest request);
+
+    List<CandidateExam> list(long id);
+
```
camerojo, line 61: I don't think that these method can return nulls. It is a good idea to annotate them as @NotNull. Similarly if things can be null as @Nullable. That allows Intellij to do some good null checking - highlighting where null pointer exceptions could occur.

ui/admin-portal/src/app/components/candidates/view/exam/create/create-candidate-exam.component.ts
```diff
@@ -0,0 +1,112 @@
...
+    this.loading = true;
+
+    this.years = generateYearArray(1950, true);
+
+    this.candidateForm = this.fb.group({
+      exam: ['', [Validators.required]],
+      otherExam: ['', null],
+      score: ['', [Validators.required]],
+      year: ['', [Validators.required]],
+      notes: ['', [Validators.required]]
```
camerojo: Should notes be required? I don't think that they were on the candidate portal

camerojo (changes requested):

server/src/main/java/org/tctalent/server/api/admin/CandidateExamAdminApi.java
EhsanEhrari: Done

server/src/main/java/org/tctalent/server/service/db/CandidateExamService.java
EhsanEhrari: Done

ui/admin-portal/src/app/components/candidates/view/exam/create/create-candidate-exam.component.ts
EhsanEhrari: Done

EhsanEhrari:
@camerojo Could you review the changes on this PR please?

server/src/main/java/org/tctalent/server/service/db/CandidateExamService.java
```diff
@@ -46,4 +46,7 @@ CandidateExam createExam(
      * @return List of updated candidate exam records
      */
     CandidateExam updateCandidateExam(UpdateCandidateExamRequest request);
+
+    List<CandidateExam> list(long id);
```
camerojo, line 60: I meant something like this. It should also appear on the implementation.

```suggestion
    @NotNull
    List<CandidateExam> list(long id);
```

Also it would be good to have some JavaDoc for this method. The JavaDoc could let the user know what the id is (a candidate id I assume) and what happens when the id is not associated with a known candidate.

server/src/main/java/org/tctalent/server/service/db/CandidateExamService.java
camerojo: I don't see that it has been done - see comment above.

camerojo (changes requested):
The @NotNull requested change is not done. If it is not clear what I meant, lets have a chat on Slack.
Just about done!

server/src/main/java/org/tctalent/server/service/db/CandidateExamService.java
```diff
@@ -45,8 +47,15 @@ CandidateExam createExam(
      * @param request Request containing exam updates
      * @return List of updated candidate exam records
      */
+    @NotNull
     CandidateExam updateCandidateExam(UpdateCandidateExamRequest request);
 
+    /**
+     * List of CandidateExam associated to candidate with given candidate id.
+     * @param id ID of candidate whose CandidateExam we want
+     * @return list of candidate CandidateExam
```
camerojo: Good to document the behaviour if the candidate id is not valid. Some methods will throw an Exception but I note that this method will just not return any exams. That is OK - but it should be documented

```suggestion
     * @return list of candidate CandidateExam - empty if the candidate id is unknown
```

server/src/main/java/org/tctalent/server/service/db/CandidateExamService.java
camerojo: The @NotNull also should appear on the implementation of the list method - unfortunately implementations of interfaces do not automatically copy annotations.

camerojo (changes requested):
Just a couple of very minor things that you did not pick up: see recent comments

camerojo (approved):
Approved

PR #1298, opened 2024-08-21
Implemented System Admin Api > reassignCandidates()

samschlicht:


server/src/main/java/org/tctalent/server/service/db/impl/CandidateServiceImpl.java
```diff
@@ -3035,4 +3035,41 @@ public Page<Candidate> fetchCandidatesWithChat(
...
+     * updates the corresponding elasticsearch index entry.
+     * @param candidateList list of candidates
+     * @param newPartner the new partner to which they will be assigned
+     */
+    private void processCandidateReassignment(
+        List<Candidate> candidateList, Partner newPartner
+    ) {
+        for (Candidate candidate : candidateList) {
+            User candidateUser = candidate.getUser();
+            candidateUser.setPartner((PartnerImpl) newPartner);
```
sadatmalik: It may be worth wrapping this in an 'if (newPartner instance of PartnerImpl) {' check just in case. With some logging in the unexpected event that it isn't.

server/src/main/java/org/tctalent/server/service/db/impl/CandidateServiceImpl.java
```diff
@@ -3035,4 +3035,41 @@ public Page<Candidate> fetchCandidatesWithChat(
         }
     }
 
+    @Override
+    public void reassignSavedListCandidates(SavedList savedList, int partnerId) {
```
sadatmalik, line 3062: I might annotate this method as @Transactional in case of runtime exceptions; it will roll back.

sadatmalik (commented):
Just a couple of comments

camerojo (dismissed):

server/src/main/java/org/tctalent/server/service/db/impl/CandidateServiceImpl.java
samschlicht: 👍 absolutely, will do.

server/src/main/java/org/tctalent/server/service/db/impl/CandidateServiceImpl.java
samschlicht: 👍 Ah great, wasn't familiar with that dimension of the @Transactional annotation. Have read up and makes sense, will add.

camerojo (dismissed):

sadatmalik (approved):
Approved.

PR #1299, opened 2024-08-21
Removed erroneous help text from date picker component

### LINKED ISSUE #1273: Bug: erroneous help text on dependant DOB input
This date picker is in lots of places, and in each that I've tried, this help text is erroneous — 01 just fails the validation and shows an error. Suggest we delete but will run by @cazcam34 in particular in the team meet.

samschlicht:


sadatmalik (approved):
Approved.

PR #1300, opened 2024-08-21
WIKI DOC ONLY - Flyway process

### LINKED ISSUE #750: Update Git Wiki with Flyways process
Something like this, as discussed via Slack w JC:

1. Check out the Flyways branch
2. Add your new file next in the numeric sequence
3. Post on #tool-tcsoftware-tech-int letting team know the number you've taken
2. Merge Staging into Flyways
3. Add your database updates
4. Check out your own branch and merge Flyways into it
5. Rebuild the DB from your own branch

NB also, from John:
Where that process breaks is when DB fields are deleted or renamed. That will break code that is still using the previously unchanged fields.
In that case we need to make sure that the DB changes/Flyway is merged into staging - together with the changed code that uses the fields ASAP - and certainly before someone else needs to add a new Flyway.

samschlicht:


camerojo (approved):
Many thanks Sam!

PR #1303, opened 2024-08-22
New partner assignment logic for candidates

### LINKED ISSUE #1301: Support for Balochistan
We need two partners in one country - Pakistan - with one being the default.

Changed the partner assignment logic which supports the requested "default local partner" idea.

This solution is also a better solution for FCA and IRC in Uganda. We can make FCA the default by reinstating them as country assignment for Uganda.

Change went in as hotfix.

The urgency came from an urgent request through Rachel Beck early this morning, coming from UNHCR Pakistan who wanted to start training with a second partner in Pakistan (Balochistan region) tonight. They are called SEHER https://seher.org.pk/. I have configured them on prod and tested it - so should be good to go tonight.

- [x] Update Tetra doc with new logic. Done - https://app.tettra.co/teams/talentbeyondboundaries/pages/talent-catalog-software-v2#header-bmnrq-partner-assignment
- [x] Notify Zeynep and Rachel. Also changed FCA to be assigned partner for Uganda on prod. That will make them the default - but IRC can still register their own candidates in Uganda using p=irc. Tested this on prod as well.

camerojo:


sadatmalik (dismissed):
Approved.

PR #1304, opened 2024-08-22
1291 salesforce not returning errors to portal when createupdate

### LINKED ISSUE #1291: Salesforce not returning errors to portal when create/update
As an admin user,
I want to know if something has gone wrong when creating or updating job opps/candidate opps,
so that I don't miss out on bugs.

Currently SF only logs errors (so only viewable to our team by viewing logs) and on the front end it appears to have processed without problems. However these are created/updated in batches so we don't want to throw errors on failures as this will stop the whole batch. Process batch and then pull out any errors and return them.

cazcam34:


camerojo (approved):
Nice!

PR #1308, opened 2024-08-22
1288 missing starredstatus of lists or searches selected in side bar other missing data not displaying eg shared by

sadatmalik:
This PR fixes the reported bug:

- gets the full candidate source for the currently selected source in the candidate source browser
- gets the extended candidate source info if the user requests "see more."
- implements a candidate source cache
- caches the full or extended candidate source results as required
- clears cache entries when the browsing tab changes
- fixes breaking unit tests related to changes made in this PR
- fixes a few other broken unit tests not related to changes made in this PR


ui/admin-portal/src/app/components/util/candidate-source/candidate-source.component.ts
```diff
@@ -242,14 +241,28 @@ export class CandidateSourceComponent implements OnInit, OnChanges, OnDestroy {
     });
   }
 
-  private isAlreadyLoaded(dtoType: DtoType): boolean {
```
camerojo: Shame that we have to expose these methods for the unit testing. Bit of a shortcoming in Angular - missing the package private that Java has.
There are some suggestions around alternatives to exposing methods that should be private for unit testing purposes. See for example https://stackoverflow.com/questions/35987055/how-to-write-unit-testing-for-angular-typescript-for-private-methods-with-jasm

Maybe the @ts_expect_error notation might be worth looking at - see https://stackoverflow.com/a/56199483/929968

ui/admin-portal/src/app/components/candidates/show/show-candidates.component.ts
```diff
@@ -506,7 +506,6 @@ export class ShowCandidatesComponent implements OnInit, OnChanges, OnDestroy {
         request.pageSize = this.pageSize;
         request.sortFields = [this.sortField];
         request.sortDirection = this.sortDirection;
-        request.dtoType = DtoType.PREVIEW;
```
camerojo, line 510: This means that we will always do full fetches of candidate data when displaying sets of candidates (lists or search results) - removing all performance optimizations. Was that intentional?
Clearly it does fix the bug where candidate-occupations were not displayed - but you could also do that by moving the candidate_opportunities into the basic candidateDto in CandidateBuilderSelector (out of the !PREVIEW section) - which slows down performance but still leaves some performance optimization in place.

camerojo (changes requested):

ui/admin-portal/src/app/components/util/candidate-source/candidate-source.component.ts
sadatmalik: The `@ts-expect-error` looks like a great alternative - thanks for that. I'll give it a try, and if it works as expected, I'll revert those methods to private.

ui/admin-portal/src/app/components/candidates/show/show-candidates.component.ts
sadatmalik: Yes, this was intentionally reverted to a full fetch for this PR, but it’s not my intention to leave it that way for the release. I should have been clear about that in the PR comments. 

The thought was to fix the 2 bugs as quickly as possible in staging; and take a more careful review of what we need to and don't need to send for show-candidote previews, including selectively sending only the minimum required opportunity fields, in a separate issue/PR. 

However, given this PR is not yet merged, at this stage it may be better to do it all in this PR after all. Will take stock in the morning.

ui/admin-portal/src/app/components/util/candidate-source/candidate-source.component.ts
EhsanEhrari: @sadatmalik Testing private methods in Angular applications using Jasmine is generally discouraged, as private methods are meant to be internal to a class and should not be directly tested. 
However, if you still need to test a private method, you can do so using a few different approaches:

1. Testing via Public Methods
2. Using TypeScript Casting
3. Refactoring for Testability
4. Using Bracket Notation

sadatmalik: Thanks @EhsanEhrari . I agree that the general consensus is to avoid directly testing private methods, though I believe there can be exceptions depending on the context. It's definitely helpful to know there are better ways to test private methods without making them non-private.

I followed your suggestion to use bracket notation, which was also mentioned in the links that @camerojo shared. Noting that using this approach means there was no need to annotate with @ts-expect-error.

ui/admin-portal/src/app/components/candidates/show/show-candidates.component.ts
sadatmalik: I've pushed changes to revert those methods under test mentioned above to private. Additionally, I've opened a new ticket to address the show-candidates preview DTO fields (#1321). I think we can now merge this PR's changes to staging and pick up show-candidates previews as #1321.

camerojo (approved):
Approved

PR #1310, opened 2024-08-23
Candidate search card restores previous scroll y position

### LINKED ISSUE #1305: Remember scroll location on the candidate search card scroll
As an admin user,
When I am scrolling down the search card to view something lower down (e.g. languages section under General tab) I want to keep my scroll position when switching between candidates in a list,
So that I am not having to continue to rescroll down when looking at the same section.

I saw there's something called scroll position restoration in angular - worth looking into, I'm not sure if this is going to be possible for our side profile card even but it would be useful.

See how it jumps back up when changing candidates

samschlicht:


ui/admin-portal/src/app/components/candidates/show/show-candidates.component.ts
```diff
@@ -314,6 +325,19 @@ export class ShowCandidatesComponent implements OnInit, OnChanges, OnDestroy {
     this.localStorageService.set(this.savedListStateKey() + this.showClosedOppsSuffix, showClosedOppsValue.toString());
   }
 
+  /**
+   * Restores candidate profile search card to previous px distance from top if > 0.
+   */
+  public setSearchCardScrollTop() {
+    // Starting at 0 is default behaviour so no action needed in that case.
+    if (this.searchCardScrollTop > 0) {
+      // The card has to fully render before scrolling, so a short delay is necessary.
```
sadatmalik, line 334: This is a reasonable workaround. I wonder if using one of the Angular lifecycle hooks e.g. ngAfterViewInit or ngOncChanges would avoid the need for a timeout? It might be worth a look, but it is only worth spending too much time on if it doesn't give a quick result.

sadatmalik (commented):
Left a PR comment - might be worth a quick look at one of Angular's component lifecycle hooks - e.g. AfterViewInit which may provide an alternative to using a timeout

camerojo (approved):

ui/admin-portal/src/app/components/candidates/show/show-candidates.component.ts
samschlicht: @sadatmalik 
The reason this one took a bit longer than expected was that I exhausted all possibilities! It looks like [afterRender](https://angular.dev/guide/components/lifecycle#afterrender-phases) might address this scenario but isn't included in our Angular version. AfterViewChecked and AfterViewContentChecked are the latest hooks available to us.

AfterViewInit runs _once_ after initialisation of the component — but we need these methods to run every time a new candidate is selected (which doesn't reinitialise the component, just changes its content). 

OnChanges runs even sooner (as soon as the candidate input changes), so didn't even work with a timeout.

The tab selection code was running 6-7 times per candidate change, so my change at least takes care of that. I agree that timeouts aren't the best, but in this case there weren't any other options.

PR #1314, opened 2024-08-26
1156 mini intake revisions to destination intake and registration process

### LINKED ISSUE #1156: Mini Intake: Revisions to Destination Intake and Registration Process
**Context:**
In a meeting with Noura, Mariam, Zeynep, John, and Sadat on 5 July, the following revisions were agreed upon:

**Revisions:**

1. 

- [x] **Destination Preferences Question Section:**

   - Add a **"Destination preferences"** question section to the candidate registration process.
   - Move the questions on **destination interest** and associated **notes** from the mini intake to the registration process.
---
2. 

- [x] **Capturing Notes:**

   - Ensure that notes are captured for all responses (Yes, No, and Unsure) in all relevant sections.
   - Currently, notes are not captured for **"No"** responses. This should be rectified.
---
3. 

- [x] **Relatives Questions:**

   - Move the questions regarding relatives, such as **"Do you have family in Australia and if so what is the closest relative?"**, to the relevant country's Visa eligibility checks section.
   - **Note** - a question arose from Zeynep about the feasibility of the Visa eligibility check admin person communicating with the candidate directly to confirm this information. Discussed with John - there are a number of reasons during the Visa eligibility check where the admin will need to contact the candidate, so this will not be an issue. This will be potentially easier to do in the future with JobChats. Remember this is an advanced stage, presumably they are being considered for the job, so we will be in closer communication with the candidate. A process question: Should someone at destination be doing this for scalability reasons?
---
4. 

- [x] **Removal of Destination Panel from Mini Intake:**

   - Remove the **"Destination"** panel from the mini intake. Because the data capture has moved to the registration process + visa eligibility checks.
   - Incorporate the destination preferences information provided during the registration process into the **General** tab of the candidate profile.

cazcam34:


server/src/main/java/org/tctalent/server/api/portal/CountryPortalApi.java
```diff
@@ -44,6 +44,12 @@ public List<Map<String, Object>> listAllCountries() {
         return countryDto().buildList(countries);
     }
 
+    @GetMapping("destinations")
+    public @NotNull List<Map<String, Object>> listTBBDestinations() {
+        List<Country> countries = countryService.getTBBDestinations();
```
camerojo: Can we change to get rid of TBB - eg listTCDestinations - and elsewhere

camerojo (approved):
Approved

PR #1318, opened 2024-08-27
913 Support for read only users: support employer access hiring manager

### LINKED ISSUE #913: Get support for ReadOnly users working - New job contact type for Employer Access: "Hiring manager"
As a hiring manager at an employer
I would like to be able to monitor the progress of a job that we are advertising
so that I know what kind of candidates are available and so that I can give feedback and prepare for my interview with promising candidates.

Suggestion from @ajt001 

Hiring manager should be able to view jobs relevant to them under myJobs

Note that this is specific to direct Employer access. With indirect access such as through TBB or Recruiter portal, we normally will not have direct contact with hiring managers.

camerojo:
The main focus of this PR is getting support for "readOnly" users working.
This is how we will support of "hiring manager" kind of user in an Employer Access context. They will have a "partner admin" role - but just read only access. 

This involved:
- Fixing a bug where partnerAdmins (eg an Employer doing employer access) could not add other partnerAdmins (eg a read only "hiring manager")
- A review of SecurityConfiguration to make sure that READONLY users were properly authorized for different url patterns
- In the angular code, disabling functionality that wasn't compatible with read only access 
- A tidy up and extension of authorization methods to support readonly access
- A refactoring moving some authorization methods from base.ts to authorization.service.ts
- Renaming of ambiguous Angular authService variables to authorizationService


ui/admin-portal/src/app/components/candidates/show/show-candidates.component.html
```diff
@@ -151,7 +151,7 @@
           Remove <i class="fas fa-user-check"></i>
         </button>
 
-        <button class="btn btn-sm btn-secondary" *ngIf="isJobList()" [disabled]="!isSelection()"
+        <button class="btn btn-sm btn-secondary" *ngIf="isJobList() && isContentModifiable()" [disabled]="!isSelection()"
```
sadatmalik, line 154: Should the iContentModifiable check also apply to "Import employer feedback from published list" below?

sadatmalik (dismissed):
Left one comment but approved.

ui/admin-portal/src/app/components/candidates/show/show-candidates.component.html
camerojo: Yes you're right! Missed that - thanks @sadatmalik

PR #1330, opened 2024-08-29
1102 automated post content needs review and potentially updatable in one place

### LINKED ISSUE #1102: Automated post content - Needs review and potentially updatable in one place
Currently in staging we have automated posts that haven't been reviewed, these need to be looked over as many also will be viewed by candidates. 

Get copy reviewed by @camerojo and Zeynep - put all post copy below in issue for review.

cazcam34:
Fix some things when reviewing posts and post content
1. One post wasn't sending as it wasn't passing the ngIf
2. Date format for next step was confusing and updated to display better see screenshots beow

Before (see next step text)
Screen Shot 2024-08-29 at 1 39 31 pm

After
Screen Shot 2024-08-29 at 4 33 04 pm

Also fixed up the text in next step on the front end - adding a subtext note to explain the automation:
Before

After
(see note after next step field)
Screen Shot 2024-08-29 at 4 34 35 pm

### DIFF
--- a/server/src/main/java/org/tctalent/server/service/db/impl/JobServiceImpl.java
+++ b/server/src/main/java/org/tctalent/server/service/db/impl/JobServiceImpl.java
@@ -899,10 +899,7 @@ public class JobServiceImpl implements JobService {
      * @param stage New stage
      */
     private void changeJobStage(SalesforceJobOpp job, JobOpportunityStage stage) {
-        job.setStage(stage);
-
         //Do automation logic
-
         // If stage changing, send automate post to JobCreatorAllSourcePartners chat
         if (!stage.equals(job.getStage())) {
 
--- a/server/src/main/java/org/tctalent/server/util/NextStepHelper.java
+++ b/server/src/main/java/org/tctalent/server/util/NextStepHelper.java
@@ -37,14 +37,18 @@ public class NextStepHelper {
      */
     public static String nextStepAuditNameDelimiter = " --";
     /**
-     * The audit date ends with this string
+     * The audit date stamp ends with this string
      */
     public static String nextStepAuditDateDelimiter = "| ";
 
     /**
      * If the requested next step is different from the current next step, the next step
-     * will be updated. In that case we add special text (the audit timestamp) to the
-     * end of the nextStep - indicating who has made this change to the next step, wand when.
+     * will be updated. In that case we add special text:
+     * <ul>
+     *     <li> the audit name to the end of the stripped nextStep.</li>
+     *     <li> the audit date to the start of the stripped nextStep. </li>
+     * </ul>
+     * This indicates who has made this change to the next step, and when. 
      * This is useful for auditing purposes.
      * <p/>
      * This method performs this logic, returning the processed next step which is what should be
@@ -52,7 +56,7 @@ public class NextStepHelper {
      * @param name Name of person initiating the next step update
      * @param date Date of the update
      * @param currentNextStep The current next step
-     * @param requestedNextStep Thw requested new next step
+     * @param requestedNextStep The requested new next step
      * @return The processed text which should be used for the next step update
      */
     public static String auditStampNextStep(String name, LocalDate date,
@@ -80,9 +84,9 @@ public class NextStepHelper {
     }
 
     /**
-     * Strips of any existing audit time stamp
+     * Strips of any existing audit stamp
      * @param requestedNextStep Requested new next step
-     * @return Same text just with any existing next step stripped off.
+     * @return Same text just with any existing audit stamp stripped off.
      */
     private static String removeExistingStamp(String requestedNextStep) {
         String stripped = requestedNextStep;
@@ -98,10 +102,11 @@ public class NextStepHelper {
     }
 
     /**
-     * Constructs the audit stamp from the given name and date
+     * Constructs the audited next step from the given name, date and stripped next step
      * @param name Name of user initiating update
      * @param date Date to be used on timestamp
-     * @return The audit stamp text
+     * @param strippedNextStep Next step text to have audit added to
+     * @return The next step text with the audit stamp text added
      */
     public static String constructNextStepAuditStamp(String name, LocalDate date, String strippedNextStep) {
         String dateStamp = nextStepDateFormatter.format(date);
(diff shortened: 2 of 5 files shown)


server/src/main/java/org/tctalent/server/util/NextStepHelper.java
```diff
@@ -30,12 +30,16 @@ public class NextStepHelper {
     /**
      * Format used for date timestamp in next step audit stamp
      */
-    public static DateTimeFormatter nextStepDateFormatter = DateTimeFormatter.ofPattern("yyMMdd");
+    public static DateTimeFormatter nextStepDateFormatter = DateTimeFormatter.ofPattern("ddMMMuu");
```
sadatmalik, line 33: Recommendation generated by Amazon CodeGuru Reviewer. Leave feedback on this recommendation by replying to the comment or by reacting to the comment using emoji.

### DateTimeFormatter.ofPattern [](|67999369-6dcc-4893-b659-89de93884ecc|)
#### Note: “I18N doesn’t apply because this use case is internal” is a misconception. For formatting that isn’t user-facing: Set Locale to Locale.ROOT, set TimeZone to UTC, and check if your format is [predefined in DateTimeFormatter](https://docs.oracle.com/javase/8/docs/api/java/time/format/DateTimeFormatter.html).

Picture strings (for example, `MM-dd-yyyy` and `HH:mm:ss`) that you specify to represent the date won’t be appropriate for all locales, and should not be hardcoded. Different locales use different date and time formats, positioning, punctuation, and even numbers themselves. So, the developer or a translator has to adapt the string for each locale&#39;s specific needs. Otherwise, the results could look strange or wrong to users in another language or locale.

Also, DateTimeFormatter.ofPattern shouldn’t be used to fix issues with SimpleDateFormat, since the two produce similar results. The [Internationalization channel on Broadcast](https://broadcast.amazon.com/channels/28439) has two videos on the challenges of using SimpleDateFormat for [User-displayed customer experience (CX)](https://broadcast.amazon.com/videos/310354) and [Application-internal timestamps and application-specific/application-defined values](https://broadcast.amazon.com/videos/307513).

### Suggested solutions:
#### For user-facing date and time:

1. **[PREFERRED]** [MessageFormat](https://w.amazon.com/bin/view/I18N/I18N_Best_Practices/I18N_Imports/) supports &#34;skeletons&#34; that can automatically select the right pattern for the customer&#39;s locale. To learn more, see [How to format dates](https://w.amazon.com/bin/view/INTech/docs/formatting-values/how-to-format-dates/) and the [MessageFormat Cheat Sheet](https://w.amazon.com/bin/view/INTech/docs/formatting-values/messageformat-tldr/).
    ```
    MessageFormat fmt = new MessageFormat(pattern, locale); 
    Object[] args = { new Date() }; 
    return fmt.format(args);
    ```
If you want to get the string translated, create a resource file for the date and time pattern. Use Panther to get the date and time pattern string translated. MessageFormat is built into [Panther](https://w.amazon.com/bin/view/INTech/Panther). See the [DateTimeFormatter.ofPattern]({0}) wiki.

2. Consider using the ICU4J library `DateFormat.getPatternInstance` method (which uses a date skeleton string) instead of the stock Java `DateFormat` class. Skeleton-based date and time formats adapt to the locale automatically.
    ```
    import com.ibm.icu.text.DateFormat; // note different import!
    DateFormat df = DateFormat.getPatternInstance(&#34;jmsz&#34;, pageLocale);
    ```

#### For application-internal timestamps:
1. Use a predefined ISO 8601 format like `ISO_INSTANT` (which looks like `2011-12-03T10:15:30Z`) from [DateTimeFormatter](https://docs.oracle.com/javase/8/docs/api/java/time/format/DateTimeFormatter.html).
    ```
    String now = DateTimeFormatter.ISO_INSTANT.format(new Date().toInstant());
    ```

2. If your format is not *[predefined in DateTimeFormatter](https://docs.oracle.com/javase/8/docs/api/java/time/format/DateTimeFormatter.html)* and you *must* use a custom format:
    - Make sure you pass in the Locale (for non-user-facing, internal-processing usage, use `Locale.ROOT` )
    - Set the timezone (for example, `UTC`), if you want to ensure it is a certain timezone.
    
    Example:
    ```
    // your custom pattern goes here
    final String somePattern = &#34;yyyy MM dd HH mm ss z&#34;; // $NON-NLS-L$ Custom hard-coded pattern
    
    DateTimeFormatter myFormatter = 
    DateTimeFormatter.ofPattern(somePattern, Locale.ROOT)  // $NON-NLS-L$ custom pattern, invariant locale
            // only add this line if you want to ensure UTC (or some other time zone)
            .withZone(ZoneId.of(&#34;UTC&#34;)); // $NON-NLS-L$ forcing UTC
    
    myFormatter.format(/* some value */);
    ```

server/src/main/java/org/tctalent/server/service/db/impl/JobServiceImpl.java
```diff
@@ -899,10 +899,7 @@ private void updateJobFromRequest(SalesforceJobOpp job, UpdateJobRequest request
      * @param stage New stage
      */
     private void changeJobStage(SalesforceJobOpp job, JobOpportunityStage stage) {
-        job.setStage(stage);
-
```
cazcam34, line 903: this setStage appears again further down - it was a duplicate

cazcam34: Just explaining to make the PR review make better sense

server/src/main/java/org/tctalent/server/util/NextStepHelper.java
```diff
@@ -30,12 +30,16 @@ public class NextStepHelper {
...
      */
-    public static String nextStepAuditStampDelimiter = " --";
+    public static String nextStepAuditNameDelimiter = " --";
+    /**
+     * The audit date ends with this string
+     */
+    public static String nextStepAuditDateDelimiter = "| ";
 
     /**
      * If the requested next step is different from the current next step, the next step
```
camerojo, line 45: This JavaDoc needs to be updated. It talks about the audit stamp being added to the end - update to clarify what the new audit stamp is - ie date at front and user and end.

camerojo (changes requested):

server/src/main/java/org/tctalent/server/util/NextStepHelper.java
cazcam34: Updated

camerojo (approved):

PR #1336, opened 2024-08-30
1281 pressing escape to close column selector modal in search and list view displays an error

### LINKED ISSUE #1281: Pressing escape to close column selector modal in search and list view displays an error
From here:

Pressing escape to close the modal, gives:

sadatmalik:
This PR:

- Allows escape key pressing to close column select modal
- Factors out common methods in show-candidates and candidate-source-results to candidate-source-base


camerojo (approved):
Nice! Good to factor out all that common code.

PR #1337, opened 2024-09-02
1327 update default columns to show intake latest replace updated date col

### LINKED ISSUE #1327: Update default columns to show intake latest (replace updated date col)
- [x] Add IELTS score to default columns, remove updated date from default 
- [x] Add 'Additional Information' section to general tab on candidate search card

Remainder of issue #1253

cazcam34:
- They can only see the Additional Information section (that includes updated date) under the General tab IF they are an admin user. So showing that section doesn't neccessarily replace the Updated column on default.

- Also in order to show the Additional Information section in search card I had to add a missing 'adminUser' input which that section requires in order to show that component. In fact there was a few missing inputs on tab sections. I fixed this up and matched the tab permission inputs (adminUser) on all tabs and now the permissions are the same for the search card and the view.

However there are two auth methods used for permissions:

- isAdminUser which just checks for admin role

- canViewPrivateInformation this checks for admin role but also that the partner matches the candidate partner (or if default partner can view all as long as admin). **Is this the better auth method to use instead of isAdminUser now we are adding partners?**

### DIFF
--- a/ui/admin-portal/src/app/components/util/candidate-search-card/candidate-search-card.component.html
+++ b/ui/admin-portal/src/app/components/util/candidate-search-card/candidate-search-card.component.html
@@ -108,7 +108,8 @@
         <ng-template ngbNavContent>
           <app-candidate-general-tab
             [candidate]="candidate"
-            [editable]="false">
+            [editable]="false"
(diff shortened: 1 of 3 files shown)


camerojo (approved):
Approved

PR #1339, opened 2024-09-02
1321 selectively send required preview fields for show candidates

sadatmalik:
This PR:

- selectively sends candidate opportunity preview data required to show-candidates
- fixes a data bug by including candidate id in short candidate dtos because this field is used in candidate-opps component for fetching chats


camerojo (approved):
Approved - nice bug to fix!

PR #1346, opened 2024-09-04
1320 refine functionality that is available to employer access

camerojo:
With 66 changed files this looks scarier than it is.

Most of those files are deleted components that have been replaced by a single more generic component joi-data component - which pushes some functionality down into JobIntakeComponentBase.

There are around 18 files that have changed apart from the deleted ones - and most of them are simple component changes and new authorizations which restrict what an Employer Access Partner sees.

Issue #1320, opened 2024-08-27
Refine functionality that is available to Employer access 


camerojo:
- [x] In Job Intake for employer access Employer Details should be displayed differently to Employer
- [x] Make generic joi component rich/vs plain text input. 
- [x] Add help tool tips to all job fields 
- [x] Employer - Lists: can't Import candidates, Publish, Resolve tasks
- [x] Drop zombie database fields in salesforce_job_opp. recruitment_process and employer_cost_committment
- [x] Add new translation keys to Release processing - can be added once running in staging.

PR #1349, opened 2024-09-06
1255 only changes to a submission list should automatically create or modify candidate opportunities cases

### LINKED ISSUE #1255: Only changes to a submission list should automatically create or modify candidate opportunities (cases)
Naji reported a problem where he unintentionally created a whole bunch of opportunities by adding candidates to a list he had associated with a job opp.

This is too easy to do. The decision was that only submission lists should automatically create or modify cases.

In fact, unless we see that many people are using the ability to associated a job op with a list, we should consider only allowing job opps to be associated with submission lists (and maybe exclusion lists and suggested lists).

Changes
- Users can no longer associate lists with jobs. This not only happens automatically during job creation - in particular with the automatically generated submission list.
- Only submission lists automatically create a case when a candidate is added. Only submission lists automatically close a case when a candidate is removed. The "Update Cases" button only appears on submission lists.

camerojo:
Note that this PR is based on #1346 - issue #1320 

It should be reviewed after that PR is merged with staging. That way you will avoid also seeing all the changes for the base PR


sadatmalik (approved):
Approved.

PR #1350, opened 2024-09-09
1155 Confirm interest in International Recruitment

### LINKED ISSUE #1155: Mini Intake: Interest in International Recruitment
**Context:**
This task is complementary to the changes noted in issue #1153. 

In a meeting with Noura, Mariam, Zeynep, John, and Sadat on 14 June, it was agreed to add an additional initial question at the start of the intake process to re-confirm the candidate's interest in international employment opportunities.

**Details:**

1. **New Initial Question:**
   - Add a question at the beginning of the intake process: **"Are you still interested in international employment opportunities?"**
---
2. **Response Options:**
   - The candidate may answer with:
     - **Yes**
     - **No**
     - **Unsure**
---
3. **Guidance Notes for Interviewers:**
   - Include guidance notes for interviewers on how to respond based on the candidate's answer. Mariam and Noura to provide guidance note. John will approve the final wording.

camerojo:


ui/admin-portal/src/app/components/candidates/intake/avail-immediate/avail-immediate.component.ts
```diff
@@ -38,16 +42,44 @@ export class AvailImmediateComponent extends IntakeComponentBase implements OnIn
...
+    //If the candidate's status is withdrawn, assume that they are not interested.
+    if (this.candidate.status == 'withdrawn') {
+      interested = YesNo.No;
+    } else {
+      //If the candidate's status is not that they have withdrawn then compute their status from
+      //any response they have made to the availImmediate question.
+      //If they have got as far as answering that question - even with a "NoResponse" response,
+      //we can assume that they said that they were interested.
+      if (this.candidateIntakeData?.availImmediate &&
+        this.candidateIntakeData?.availImmediate != YesNo.NoResponse) {
```
sadatmalik, line 72: The logic is more cautious than the comment suggests. The code only considers them "interested" if they actively responded with a Yes or No. It will not consider them interested if they actively responded with NoRsponse.

This goes against the sense of the comment - guessing it's maybe a typo?

Should the check simply be:

if (this.candidateIntakeData?.availImmediate) {

Then, any response to the availability question will be deemed to express interest, which I think is the intent here.

sadatmalik (changes requested):

ui/admin-portal/src/app/components/candidates/intake/avail-immediate/avail-immediate.component.ts
camerojo: Good pick up - but it is the comment that is wrong.

The code is the way it is supposed to be.

The problem is that Interested is not stored and needs to computed. NoResponse is not a valid response. We can assume Interested only if they answered Yes or No.

sadatmalik (approved):
Approved.

Issue #1365, opened 2024-09-12
Browsing lists displays card details for Search with id = x instead of List with id = x


camerojo:
The problem is that the onChanges logic inside CandidateSourceComponent has changed.

<img width="1171" alt="Screenshot 2024-09-12 at 12 38 44 PM" src="https://github.com/user-attachments/assets/4bd504d5-2ac3-4df7-b46d-30b6ae111117">

PR #1366, opened 2024-09-12
1364 improve access to chats

### LINKED ISSUE #1364: Improve access to chats
As a partner associated with a job and cases
I need improved access to chats and cases

- [x] Short display of submission list should show case icon next to each candidate
- [x] Case display should show tab for destination/source chat

camerojo:
Also fixes bug #1365


sadatmalik (approved):
Approved.

PR #1380, opened 2024-09-17
1376 clarify partner chats when viewed by tbb who is both source partner and destination partner

### LINKED ISSUE #1376: Clarify partner chats when viewed by TBB who is both source partner and destination partner
Bug: Source partners can't edit their own user contact

PLUS...

As a TBB staff member looking at the partner chats 
I would like the chat descriptions to be clearer
because the partner chats screen is confusing when TBB is both source partner and destination partner for a case

@sadatmalik This is the confusion that you mentioned on our last team call.

Screenshot 2024-09-16 at 12 33 17 PM

camerojo:


sadatmalik (approved):
Approved.

PR #1395, opened 2024-09-23
1374 modify the way chats are displayed in case display grouping by candidate and not candidate

### LINKED ISSUE #1374: Modify the way chats are displayed in case display - grouping by candidate/not candidate
Two tabs: candidate chats, non candidate chats.

In each tab, multiple chats are displaying using chat descriptor components (#1373) in selection on left and selected chat on right (like the partner chats)

camerojo:


sadatmalik (dismissed):
Approved.

PR #1412, opened 2024-09-25
1408 mark chat as unread when posting

### LINKED ISSUE #1408: Mark chat as unread when posting
Currently as a admin portal user sending a chat,
the chat will be marked as unread even though I am the one sending the chat,
so this leads to confusion and extra effort as I have to click 'Mark as Unread' everytime I send a post so that I can keep my notifications up to date.

I have sent this chat, and you can see it has triggered a notification and requires the 'Mark as read' button to be checked to clear it after sending post.

cazcam34:


camerojo (dismissed):
Approved.

@cazcam34 I changed it to work a little differently - which also gets rid of the flashing MarkAsRead. Have a look at the code.

I will also push it into staging so that we can play with it on staging.

PR #1424, opened 2024-09-30
1423 show closed cases on submission lists no longer does anything

### LINKED ISSUE #1423: Show closed cases on submission lists no longer does anything
This is because the Show closed cases checkbox was moved out of the form group in the reformatting of the display.

A simple fix is to move the checkbox back into the form group. ie between the \<form\> block in the html.

But that messes up the layout. 

Screenshot 2024-09-30 at 10 26 33 AM

cazcam34:


camerojo (approved):
Approved

PR #1446, opened 2024-10-07
1362 bug with display of some icons next to each other in a chat post

### LINKED ISSUE #1362: Bug with display of some icons next to each other in a chat post
Screen shot from production. The icons should have been Australian flag and Kangaroo.

Screenshot 2024-09-11 at 9 46 18 AM

cazcam34:


ui/admin-portal/src/app/components/chat/create-update-post/create-update-post.component.ts
```diff
@@ -126,7 +126,7 @@ export class CreateUpdatePostComponent implements OnInit {
     this.emojiPickerVisible = false;
     const index: number = this.quillEditorRef.selection.savedRange.index;
     this.quillEditorRef.insertText(index, `${event.emoji.native}`, 'user');
-    this.quillEditorRef.setSelection(index + 2, 0);
+    this.quillEditorRef.setSelection(this.quillEditorRef.getLength(), 0);
```
camerojo: @cazcam34 - This is the right idea but I am not sure that it is replacing like for like.

The original code is inserting at the end of the current selection (it could be in the middle of a post).

Your code seems to always be inserting at the end of the whole data. 

Have you tested this when the emoji is being added in the middle of a post?

camerojo (changes requested):

ui/admin-portal/src/app/components/chat/create-update-post/create-update-post.component.ts
cazcam34: Good point - but it's working fine as it's inserting the emoji at the index (see insertText) but after the insert it's just making sure that the selector goes to the end of the string. Beforehand the selector was being put at the index + 2, and so if the emoji was longer it and another emoji was added without moving the selector it was inserting it at the wrong spot. @camerojo

camerojo (approved):

PR #1451, opened 2024-10-07
1450 keyword search in email and phone searches

### LINKED ISSUE #1450: Keyword search in email and phone searches
From Lea:

 I have been trying to search using the same candidate's email in the screenshot below, but it is not giving me any results. I've tried to search with another candidate's email and still nothing. As opposed to how fast it is now to search with TBB IDs. <!subteam^S06MQJ4GA8P> anyone else facing this issue?

[Slack Message](https://refugeejobsmarket.slack.com/archives/C42S4AJ73/p1728292960921779?thread_ts=1728292960.921779&cid=C42S4AJ73)

sadatmalik:
This PR:

- Fixes a bug in search by phone/email
- Fixes a breaking unit test (server-side) related to daylight savings

The changes in this PR were already pushed to production as a Hotfix.

### DIFF
--- a/server/src/main/java/org/tctalent/server/service/db/es/ElasticsearchServiceImpl.java
+++ b/server/src/main/java/org/tctalent/server/service/db/es/ElasticsearchServiceImpl.java
@@ -65,10 +65,10 @@ import org.tctalent.server.service.db.UserService;
 public class ElasticsearchServiceImpl implements ElasticsearchService {
 
   private static final String CANDIDATE_NUMBER_FIELD = "candidateNumber";
-  private static final String EMAIL_FIELD = "email";
-  private static final String EXTERNAL_ID_FIELD = "externalId";
+  private static final String EMAIL_FIELD = "email.keyword";
+  private static final String EXTERNAL_ID_FIELD = "externalId.keyword";
   private static final String FULL_NAME_FIELD = "fullName";
-  private static final String PHONE_NUMBER_FIELD = "phone";
+  private static final String PHONE_NUMBER_FIELD = "phone.keyword";
 
   private static final String COUNTRY_KEYWORD = "country.keyword";
   private static final String STATUS_KEYWORD = "status.keyword";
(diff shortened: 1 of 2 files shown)


server/src/main/java/org/tctalent/server/service/db/es/ElasticsearchServiceImpl.java
```diff
@@ -65,10 +65,10 @@
 public class ElasticsearchServiceImpl implements ElasticsearchService {
 
   private static final String CANDIDATE_NUMBER_FIELD = "candidateNumber";
-  private static final String EMAIL_FIELD = "email";
-  private static final String EXTERNAL_ID_FIELD = "externalId";
+  private static final String EMAIL_FIELD = "email.keyword";
+  private static final String EXTERNAL_ID_FIELD = "externalId.keyword";
```
camerojo, line 70: Not sure what the problem is but this still doesn't work well in production.

For example - see this production screen. Partial matches are not working.

Screenshot 2024-10-08 at 2 56 35 PM

Same goes for email addresses.

A search for yahoo.com.au or script doesn't come up with anything even though the email is

jadcpub-script@yahoo.com.au

camerojo (changes requested):
See comment.

Note that I think that a plain old 

`like "%script%"`

Works well on Postgres. Despite the wild cards at front and back. The email index on users is probably in memory and is fast.

server/src/main/java/org/tctalent/server/service/db/es/ElasticsearchServiceImpl.java
sadatmalik: This is now fixed and has been deployed to production as a hotfix. 

For partial phone numbers:

Screenshot 2024-10-08 at 11 39 40

And for partial email:

Screenshot 2024-10-08 at 11 41 05

Phone and email searches should use wildcard elasticsearch queries. Previously, they used prefix queries, which are inadequate and have now been corrected.

camerojo (approved):

cazcam34:
Merged by @cazcam34 as other branches were relying on this test fix to be in staging so we can merge the other PRs

PR #1466, opened 2024-10-10
1410 employer access users should get email notification of tc posts

### LINKED ISSUE #1410: Employer access users should get email notification of TC posts

camerojo:
Looks like a lot of files but quite a few files just have added documentation - no functional changes.
Then there are the 6 modified email template files.

Main structural change was moving the existing chat post notification code out of CandidateOpportunityService and into the new NotificationService.

A significant functional change was extending notifications to ALL users - not just candidate users (or employer access users). It is also changed so that users are NOT notified if they have marked the relevant chats as read. So normal admin users, who are keeping an eye on chats during the day and marking them as read, may not see any notifications.

Another change was how email templates receive data. Instead of leaking functional knowledge of things like Searches,  Chats, Jobs, Cases, Candidate etc down into the email template code - it has been made more generic by just passing a new kind of object EmailNotificationLink. Those generic links are created from Searches, Chats, Jobs etc in the domain specific code - which means that the templates can be much dumber.

One last item of note is that none of the tests require Spring to be running (they are not SpringBootTests). This means that they run much faster. This required some refactoring of code but was worth the effort I think.


sadatmalik (approved):
Approved.

PR #1467, opened 2024-10-10
Stats display of different candidate statuses should show draft statuses - which are normally ignored except in registration stats 

### LINKED ISSUE #1453: Stats display of different candidate statuses should show draft statuses - which are normally ignored except in registration stats
There are lot of other statuses which should be displayed.

Screenshot 2024-10-08 at 3 17 08 PM

The registrations stat display is another example of a display which reports on ALL statuses - not just the active ones.

The change should be easy. It should be just a matter of calling
<pre>
        selectSql += standardConstraints(candidateIds, sourceCountryIds, constraint, false);
</pre>

instead of

<pre>
        selectSql += standardConstraints(candidateIds, sourceCountryIds, constraint);
</pre>

in CandidateStatsServiceImpl.computeStatusStats

camerojo:
Also change to run new stats by default.


sadatmalik (approved):
Approved.

PR #1470, opened 2024-10-10
1413 missing localization in candidate portal

### LINKED ISSUE #1413: Missing localization in candidate portal
I checked from the [test portal](https://tctalent-test.org/candidate-portal) and found these small errors:

- [x] **Langauge section problem.**
  When the website first loads it shows everything corrects. When I go to another section like tasks and then come back to profile the error shows, when I reload it is fixed. 
   All okay:
Screenshot 2024-10-02 at 11 53 27 AM
  Error appears:
  
Screenshot 2024-10-02 at 11 54 11 AM
Screenshot 2024-10-02 at 11 55 23 AM

- [ ] Also, some parts are not going back to english when translating to Dari. 
 
Screenshot 2024-10-02 at 11 55 49 AM

EhsanEhrari:


ui/candidate-portal/src/app/components/common/candidate-language-card/candidate-language-card.component.ts
```diff
@@ -35,21 +35,29 @@ export class CandidateLanguageCardComponent {
...
+  // Watch for changes in the inputs and update the language name accordingly
+  ngOnChanges(changes: SimpleChanges) {
+    if (changes.language || changes.languages) {
+      this.getLanguageName();
+    }
+  }
 
-  getLanguageName(language) {
+  // Compute the language name once, when inputs change
+  getLanguageName() {
```
camerojo: I suggest a bit of refactoring - including changing the name of this method to lookupLanguageName.

"get" and "set" have very specific semantics associated with them. "get" implies that you are returning a value - which the previous implementation of the method did. However it no longer returns a value.  Instead it looks up a name given a language.

I would change the method to 

lookupLanguageName(language: CandidateLanguage): string

Then in the ngOnChanges code I would add:

this.translatedLanguageName = lookupLanguageName(this.language)

Then the implementation code inside the method just accesses the this.languages array rather than directly accessing the this.language and this.translatedLanguageName fields - which makes it more modular and more loosely coupled (and simpler to unit test)

Functionally no difference of course.

camerojo (changes requested):

ui/candidate-portal/src/app/components/common/candidate-language-card/candidate-language-card.component.ts
EhsanEhrari: @camerojo Thanks for the feedback.
Please have a look on new changes and let me know if there is something else.

camerojo (approved):

PR #1461, opened 2024-10-09
1417 chat reactions are not updating posts

sadatmalik:
This PR:

- adds multicast support for reactions on chat posts.

Currently in draft status while I complete testing in all identified scenarios.

PR #1472, opened 2024-10-10
1471 multicast chat reaction an alternative approach

sadatmalik:
This PR:

- Implements an alternative approach to publishing reactions based on feedback in #1461 
- I prefer the approach implemented in this PR; it is simpler and generalises to any kind of post update

In Draft: 

- Pending a decision on implementation approach 
- And then need to complete testing of scenarios identified in #1461


camerojo:
I agree - I think that this is the way to go. @sadatmalik

camerojo (approved):

PR #1478, opened 2024-10-13
1476 performance issue with launching new search

### LINKED ISSUE #1476: Performance issue with launching New Search
Launching the New Search screen has been becoming slower and slower. 

This is caused by initializing just one control ExclusionList. That control selects a list from a drop down of ALL lists. Currently FULL details of ALL lists is loaded at component start up - this is now around 1Mb of data on production and takes around 20 seconds.

Instead lists should be searched for back on the server based on typing in their names - like we do for the quick candidate searches by name, number, email or phone number.

camerojo:


ui/admin-portal/src/app/components/candidates/find-candidate-source/find-candidate-source.component.html
```diff
@@ -0,0 +1,19 @@
+<div class="input-group">
+  <ng-select
+    [items]="sources$ | async"
+    bindLabel="name"
+    [addTag]="true"
```
cazcam34: I don't think you want [addTag] here - you can remove it as it is default to false. It is allowing users to add a tag if their auto search doesn't show anything - it won't do anything but it's redundant I think.
Screen Shot 2024-10-14 at 3 52 05 pm

cazcam34: Just reminder I think you want to remove the addTag input @camerojo - it's redundant but could be confusing

camerojo: Nice one! Thanks. I didn't know what addTag meant. Definitely don't want it.

ui/admin-portal/src/app/components/util/candidate-name-num-search/candidate-name-num-search.component.html
```diff
@@ -23,6 +23,10 @@
        [editable]="false"
        (selectItem)="selectSearchResult($event, input)"
        [placeholder]="placeholderText"/>
+<!--We need this ng-template to do the highlighting (using ngb-highlight).
+But not sure why we need the t.split. Just t does OK. - JC
+See https://ng-bootstrap.github.io/#/components/typeahead/examples#template
+-->
```
sadatmalik, line 29: The split on ' ' means if a user searches for e.g. "Sara Mou", and the result is "Sara El Moussa", both "Sara" and "Mou" will be highlighted, even though they are non-continuous in the result string.

sadatmalik (approved):
Approved.

ui/admin-portal/src/app/components/util/candidate-name-num-search/candidate-name-num-search.component.html
camerojo: Thanks!

sadatmalik:
Great stuff - much better load time in all environments now!

PR #1493, opened 2024-10-17
fix: match linkedin regex

MuhannedNoman:
This PR updates the regular expression to correctly match LinkedIn URLs.

Server-side update: Previously, the server did not accept LinkedIn URLs without the "[www](http://www/)." prefix. Since URLs without "[www](http://www/)." still navigate correctly, I modified the server to make "[www](http://www/)." optional in the URL validation.

Client-side update: I enforced the use of "http" or "https" for LinkedIn URLs on the client side to ensure proper formatting and avoid potential security issues. Additionally, I added this validation check to the admin portal’s edit component to maintain consistency.

These changes ensure that LinkedIn URLs are handled correctly, whether they include "[www](http://www/)." or not, across both the client and server.


sadatmalik (approved):
Nicely done, Muhanned - first of many more!

Issue #1438, opened 2024-10-03
Clear selections when 'Clear Search' on new search


cazcam34:
Naji's bug seems to be replicatable by the following:
Run a new search A - save selections to list A.
Clear search A and run new search B,
Save new selections to list B.
**Problem:** list A selections are appearing in list B.

**Solution:** When clicking 'Clear Search' we need to make sure we are also clearing the previous selections so that they aren't being carried over in the new selections.

See Naji slack thread here: https://refugeejobsmarket.slack.com/archives/C42S4AJ73/p1727832334949069

cazcam34:
Because we are saving a search when saving candidates to a new list - then we even if we clear that search we are still on a saved search and that saved searches selections are always going to appear in our new search when saving to another list.

cazcam34:
Naji ran a search made selections and saved to list - cleared list. Then went back to the search, made a change to the search and selected new candidates. Saved to list - the old selections and new selections are added.

PR #1494, opened 2024-10-17
Only update default saved searches on search and mark selected for ap…

### LINKED ISSUE #1492: Additional selected candidates in selections when saving to list
Had a call with Naji to continue to discuss the additional candidates appearing in list when selecting. 

Found the source of the issue. As mentioned in another issue here () the define search filter is always saving to the 'default' search upon every 'Search' button click. Even if I am in a saved search screen and the filter is populated with my saved search filters, if I change those filters and click 'Search' I am updating and saving my default search filters not the saved search filters UNLESS i click 'Update Search' button then it saves to my saved search.

In addition to the above upon each 'Search' button click we also are marking the selected candidates for that search. Each user has a different amount of selections for a search. HOWEVER we are marking the selected candidates for the default search and not the saved search!!

cazcam34:
…propriate search requested


cazcam34:
PR for your review @camerojo - I added a couple of todos just to get your thoughts on those.

server/src/main/java/org/tctalent/server/service/db/impl/SavedSearchServiceImpl.java
```diff
@@ -361,29 +361,32 @@ public Page<Candidate> searchCandidates(SearchCandidateRequest request) {
...
+                //todo Need special method which only updates search part. Then don't need the above "no changes there" stuff
+                updateSavedSearch(savedSearch.getId(), updateRequest);
+            }
 
             //Do the search
             candidates = doSearchCandidates(request);
 
             //Add in any selections
-            markUserSelectedCandidates(savedSearchId, candidates);
+            // todo do we want to remove a selection when it get's removed from a search?
```
cazcam34: Related to discussion around selections for unsaved searches - if filters are changed and a selected candidate no longer shows up in search results, could it be unselected?

server/src/main/java/org/tctalent/server/service/db/impl/SavedSearchServiceImpl.java
```diff
@@ -361,29 +361,32 @@ public Page<Candidate> searchCandidates(SearchCandidateRequest request) {
         Page<Candidate> candidates;
         User user = userService.getLoggedInUser();
         if (user == null) {
+            // todo when is the user null?
```
cazcam34: I'm not quite sure when this would be the case - did give it a proper dig around for an answer but thought you might remember?

cazcam34:
Converted to draft as I have some comments for your review

cazcam34:
Maybe can undo the fix in #1438 where clicking Clear Search we are clearing the selections - now that it's much clearer to see the selections (with the total selected number on the show candidates component) I think we could remove this.

cazcam34:
Note to myself - see comment above re #1438 I think we can revert this change

camerojo (approved):
Approved - great work!

Issue #511, opened 2023-12-11
Convert WYSIWIG to ngx-quill from ngx-wig for job chats


cazcam34:
See here: https://www.npmjs.com/package/ngx-quill

cazcam34:
Existing sprint issue #446 is dependant on this - broken into it's own issue for sake of keeping it clean.

Issue #446, opened 2023-11-14
Uploads of files (images and other docs) to JobChat posts


camerojo:

cazcam34:
Store uploaded link in the html of the post - rather than in the database in a field.

cazcam34:
Going to move to ngx-quill from ngx-wig. Will create new issue.

cazcam34:
- [x] Create ChatUploads folder in job google folder - this is where we will store the post uploads
- [x] Rename the file with a prefix indicating the job chat id (ChatId_FileName)
- [x] Figure out how we give access to the file, tested briefly sharing with public google link but didn't seem to work. Tested with a regular google image link and it worked.
- [x] How can i return a file url string back in the api, that isn't saved to an object/database.

cazcam34:
Alternative to custom image handler https://www.npmjs.com/package/ngx-quill-upload

cazcam34:
Dont forget PDFs uploads
Test on all other chats (standardise the chat view)

PR #1503, opened 2024-10-21
1483 add clear format option to job opportunity intake joi section in text editor

EhsanEhrari:


cazcam34:
This looks great @EhsanEhrari! 
On a side note, I noticed was that we now have two different rich text editors ngx-wig (our original editor) and ngx-quill (our newer implementation which we used for chats due to more customisation and image uploading - see issues #511 and #446). Should we look to consolidate into the one text editor, which would have to be quill, at some point? fyi @camerojo

camerojo (approved):
Great work @EhsanEhrari - and well documented as well.

PR #1509, opened 2024-10-29
1498 support for background batch processing

### LINKED ISSUE #1498: Support for background batch processing
As Talent Catalog developers
we need to provide users with a way of submitting batch operations that can complete as low priority background tasks
so that those long compute intensive operations do not adversely affect other users of the TC.

For example Salesforce allows users to start potentially long import or export tasks - with the user being notified by an email when the task has completed.

The need for this is illustrated by this Slack conversation https://refugeejobsmarket.slack.com/archives/C42S4AJ73/p1729283700405699
where a user had been doing large search exports and list publications which probably were the causes of the TC CPU maxing out at 100% and causing Gateway timeouts for other users.

Note that the approach proposed here is different to the approach of low priority threads - where the idea is that higher priority threads can interrupt lower priority threads when they need to. Instead this approach works with repeated scheduled tasks - with the same task being scheduled to do a bit of work, then being scheduled again after a certain delay to pick where it left off.

camerojo:
The changes to SystemAdminApi are just a placeholder for an API which kicks off a background process.

Currently it does not do anything useful but the intention is just to show how a BackRunner could be instantiated there in the SystemAdminApi code.


sadatmalik (approved):
Approved.

PR #1520, opened 2024-10-31
1409 add search filter which specifies that candidates should appear in certain lists

### LINKED ISSUE #1409: Add search filter which specifies that candidates should appear in certain lists
As a TC admin user
I would like to restrict a search to candidates appearing (or not appearing) in certain lists
so that I can use lists like tags on candidates and then do searches for candidates which have certain combinations of tags.

From Mari Christy on Slack: https://refugeejobsmarket.slack.com/archives/C42S4AJ73/p1727159252390679

From above post:

> How can I search by lists, so if I wanted to pull all those who were #USCISPending or #USParole, and receiving English classes #CB-English?

camerojo:


sadatmalik (approved):
Approved.

PR #1523, opened 2024-10-31
Don't notify candidates who only got to prospect stage.

### LINKED ISSUE #1510: Restrict visibility of Candidates to chats in Prospect stage or for those who are ineligible for TC
Reported by Mariam - see https://refugeejobsmarket.slack.com/archives/C42S4AJ73/p1730214792734639

Added as hotfix

The resolution of this issue was:

- Candidate is not notified when they just become a Prospect for a job - and no chat with the candidate is created when they become a prospect.
- When candidates are marked as ineligible, they no longer see a Chat tab on the candidate portal

Note however the issue referred to below. We do want candidates to be notified if they move beyond prospect - for example when they are moved to the MiniIntake stage.

camerojo:
Don't display chat tab to candidates with status = ineligible.

### DIFF
--- a/server/src/main/java/org/tctalent/server/service/db/impl/CandidateOpportunityServiceImpl.java
+++ b/server/src/main/java/org/tctalent/server/service/db/impl/CandidateOpportunityServiceImpl.java
@@ -838,14 +838,17 @@ public class CandidateOpportunityServiceImpl implements CandidateOpportunityServ
                 " has been removed for the job '" + opp.getJobOpp().getName() +
                 "' with the reason " + newStage.getSalesforceStageName() + ".");
 
-        // AUTO CHAT TO PROSPECT CHAT
-        JobChat prospectChat = jobChatService.getOrCreateJobChat(JobChatType.CandidateProspect, null,
+        //Only post to candidate if they got past the prospect stage for this job
+        if (opp.getStage() != CandidateOpportunityStage.prospect) {
+            // AUTO CHAT TO PROSPECT CHAT
+            JobChat prospectChat = jobChatService.getOrCreateJobChat(JobChatType.CandidateProspect, null,
                 null, candidate);
-        // Create the chat post
-        ChatPost prospectChatPostRemoved = chatPostService.createPost(
+            // Create the chat post
+            ChatPost prospectChatPostRemoved = chatPostService.createPost(
                 autoPostRemovedFromSubList, prospectChat, userService.getSystemAdminUser());
-        // Publish chat post
-        chatPostService.publishChatPost(prospectChatPostRemoved);
+            // Publish chat post
+            chatPostService.publishChatPost(prospectChatPostRemoved);
+        }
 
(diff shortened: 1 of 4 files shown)


sadatmalik (approved):
Approved.

server/src/main/java/org/tctalent/server/service/db/impl/CandidateOpportunityServiceImpl.java
```diff
@@ -881,14 +885,17 @@ private void publishAddedToSubmissionListPosts(CandidateOpportunity opp) {
         autoPostAddedToSubList.setContent("The candidate " + candidateNameAndNumber +
                 " is a prospect for the job '" + opp.getJobOpp().getName() +"'.");
 
-        // AUTO CHAT TO PROSPECT CHAT
-        JobChat prospectChat = jobChatService.getOrCreateJobChat(JobChatType.CandidateProspect, null,
-                null, candidate);
-        // Create the chat post
-        ChatPost prospectChatPost = chatPostService.createPost(
+        // Note that we don't post to candidates until they get past the prospect stage
+        if (opp.getStage() != CandidateOpportunityStage.prospect) {
```
samschlicht, line 889: @camerojo I don't understand this condition check: the method `publishAddedToSubmissionListPosts()` is only called when the candidate is added to a submission list, at which point the stage is _always_ Prospect — so aren't lines 888-898 redundant?

camerojo: You are right - but I think that the publishAddedToSubmissionListPosts method is poorly named. 

In the JavaDoc for that method it says
"Publish posts for after a Candidate Opportunity is created (e.g. Added to submission list)"

Note the "eg". The method is only called from the generic createOrUpdateCandidateOpportunity method. And that method allows for the possibility of a candidate opportunity being created with any initial stage (through the supplied CandidateOpportunityParams) - not just prospect. 

You are probably right that possibly right now candidate opportunities are only created by adding to a submission list - and the stage defaults to prospect. That wasn't the case in the past, and may not always be the case in the future. The name of publishAddedToSubmissionListPosts making assumptions about how the code got to this point, when it is and should in fact be a completely generic method (like createOrUpdateCandidateOpportunity from where it is called).

Does that make sense? @samschlicht

samschlicht: Thanks, @camerojo, yes it does — although now the condition check assumes there's no chance an opp could be created at a stage _before_ Prospect (when we would also not want a post to be sent). 

I know in practice that will almost certainly be the case — it's just potentially a little confusing without some doc.

I know it's not quite so simple but what we would ideally have is `if (opp.getStage() > CandidateOpportunityStage.prospect) { `!

We're due a discussion on the issue that I picked up in the same area, so perhaps that's an opportunity to add some clarifying doc. Don't get me wrong, I don't think this is critical.

PR #1544, opened 2024-11-04
1439 autosaving search when saving to new list old behaviour that could be removed

### LINKED ISSUE #1439: Autosaving search when saving to new list, old behaviour that could be removed
As a TC admin user,
When saving selections to a new list from an unsaved search it is also saving the search with the same name as the new list,
This is confusing as the search could be mistaken for the list and perhaps is not the desired behaviour - or if so it needs some more explanation to user

**Updated:** See notes below for background - this issue became a pandora's box of other issues relating to the saved search screen. I have created new issues for those, this one is less urgent but perhaps worth more of a discussion. Do we want to continue autosaving a search when we are saving a selection to a new list? @camerojo suspects this was done in the early days to encourage people using saved searches, however now it seems unnecessary. Spoke with Naji and he didn't seem to be bothered about it, so it doesn't sound like it's useful for him but can't be sure about others.

cazcam34:


camerojo (approved):
Approved

PR #1609, opened 2024-11-06
Reduced wait time by approx. 85% and fixed Refresh button bug in this and counterpart components.

### LINKED ISSUE #1546: Bug: 'TBB Candidate Chats' not loading in prod
Noting that it works fine in staging, but currently in prod doesn't work for TBB:

I suspect it's a performance issue related to no. of candidates, because it works for SHARP in prod, albeit slowly:

And also works in staging for TBB, reasonably quickly:

samschlicht:


camerojo (approved):
Nice!

Issue #1614, opened 2024-11-07
Spring upgrade (including Elasticsearch upgrade)


camerojo:
The approach of this issue is to take the original code and convert the source according to the instructions for moving to Spring Boot 3. 

At the same time modifying the gradle build to follow a standard Spring Initializr for Spring Boot 3 given the dependencies we have.

This has taken over the work from #1619

camerojo:
https://www.elastic.co/guide/en/elasticsearch/client/java-api-client/current/index.html

https://www.elastic.co/guide/en/elasticsearch/client/java-api-client/current/searching.html

- [x] Latest Flyway does not support old versions of Postgres - eg 11.3. At start up get errror: org.flywaydb.core.api.FlywayException: Unsupported Database: PostgreSQL 11.3
- [x] Add WebSocketSecurityConfig back in again
- [x] Make sure WebConfiguration and WebSocketConfig and WebSocketConfig2 are unchanged from current set up
- [x] Updated Background code.
- [x] Document use of NativeQuery
- [x] Check whether name, email, phone lookup is still needed in ES
- [x] Wild card logic in Elastic search - can we go back to old Postgres code
- [x] Unit test new addTermQuery
- [x] Try and merge in staging
- [x] Add new Spring JavaMail #1617

camerojo:
### Dev changes

- Configure your Intellij to compile with -parameters option: https://github.com/spring-projects/spring-framework/wiki/Spring-Framework-6.1-Release-Notes (or we can be explicit about using Annotations to name all parameters)
- Won't work with Elasticsearch versions prior to 7.14
- To turn off Elasticsearch add xpack.security.enabled=false to the environment variables on command line or DockerCompose

camerojo:
### Bugs

- [x] List selection from drop down does not update after first display of options. Caused by Hibernate change https://stackoverflow.com/a/79194708/929968
- [x] Convert all Specification's to remove .getExpressions() code.
- [x] Check for disjunctions avoiding adding if empty
- [x] Bug issue: changing a stage from a closed stage back to an unclosed stage does not set (redundant) closed boolean field to false - stays as true.

camerojo:
Included in #1649

PR #1622, opened 2024-11-11
1619 write new version of elasticsearch interface

### LINKED ISSUE #1619: First attempt (cancelled) to Write new version of Elasticsearch interface
Spring Boot 3 no longer supports the high level REST interface: RestHighLevelClient - which is what we currently use.

See https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-3.0-Migration-Guide#elasticsearch-clients-and-templates

https://chatgpt.com/share/672e9017-9910-800f-8e5f-8e4dcacb7691

### Note

This approach attempted to convert the existing running code in Spring Boot 2 over to the new Elasticsearch API. 
This doesn't seem to be a good way to go.

Instead will rewrite the new Elasticsearch code in the Spring Boot 3 code - so these changes will be delivered directly in #1614

camerojo:
Draft only. Will be deleted.

### DIFF
--- a/server/build.gradle
+++ b/server/build.gradle
@@ -81,6 +81,8 @@ dependencies {
     implementation 'com.google.cloud:google-cloud-storage'
     implementation 'com.google.apis:google-api-services-sheets:v4-rev20210629-1.32.1'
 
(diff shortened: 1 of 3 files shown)


camerojo:
Addressed by #1614

PR #1623, opened 2024-11-11
Added new closed SF candidate opp stage for candidates mistakenly added to list

### LINKED ISSUE #1501: No SF stage for accidental additions to submission list
Despite precautions, candidates could still be accidently added to a submission list (therefore creating an SF candidate opportunity). We need an SF stage for this circumstance (e.g. Removed) so these candidate opps can be closed, and we want to filter candidates with this stage from the submission list so they appear to be removed as opposed to closed.

samschlicht:


ui/admin-portal/src/app/model/candidate-opportunity.ts
```diff
@@ -73,6 +73,7 @@ export enum CandidateOpportunityStage {
   jobOfferRetracted = "Closed. Job offer retracted",
   relocatedNoJobOfferPathway = "Closed. No job offer stream (Canadian pathway stream - based on skills not job offer)",
   trainingNotCompleted = "Closed. Training not completed",
+  candidateMistakenlyAddedToList = "Closed. Candidate mistakenly added to list"
```
camerojo: Can we rename this to something more related to jobs. 
I suggest

candidateMistakenProspect = "Closed. Candidate was mistakenly proposed as a prospect for the job."

Also can we modify the ordering.

Same with candidate portal version

camerojo (changes requested):

ui/admin-portal/src/app/model/candidate-opportunity.ts
samschlicht: Amended per your suggestions, @camerojo!

camerojo (approved):

PR #1630, opened 2024-11-13
Stage and logic changes implemented per issue description

### LINKED ISSUE #1629: New SF candidate opp stages
[Slack discussion](https://refugeejobsmarket.slack.com/archives/C42S4AJ73/p1730206359238559)

New candidate opp closing stages:
- [x] Job withdrawn
- [x] Job ineligible
 

New job closing stage logic:

- 'Job ineligible' automatically assigned when the job closing stage is one of
- [x] Ineligible employer
- [x] ineligible occupation
- [x] Ineligible region

- 'Job withdrawn' replaces 'Not fit for role' as automatic assignment when job closing stage is one of
- [x] Not interested
- [x] Too expensive
- [x] Too high wage
- [x] Too long

samschlicht:


ui/admin-portal/src/app/model/candidate-opportunity.ts
```diff
@@ -73,6 +73,8 @@ export enum CandidateOpportunityStage {
   jobOfferRetracted = "Closed. Job offer retracted",
   relocatedNoJobOfferPathway = "Closed. No job offer stream (Canadian pathway stream - based on skills not job offer)",
   trainingNotCompleted = "Closed. Training not completed",
+  jobWithdrawn = "Closed. Job withdrawn",
+  jobIneligible = "Closed. Job ineligible",
```
camerojo: The ordering of options is relevant because that is order that users will see the options displayed.

Instead of just adding them at the end it would be better to insert them in the list somewhere logical. Maybe next to the jobOfferRetracted. If no other logical applies, I like to go with alphabetical - so maybe in the order:

jobIneligible
jobOfferRetracted
jobWithdrawn

(and same change in candidate portal)

camerojo (changes requested):

ui/admin-portal/src/app/model/candidate-opportunity.ts
samschlicht: @camerojo good point, pushed a new commit with those changes!

camerojo (approved):

PR #1661, opened 2024-11-22
1649-merge-angular-and-spring-boot

camerojo:
Refactor so that search takes a nativeQuery and add method which can extract JSON from a naive query - make for more flexible logging.


camerojo:
Mistakenly put changes into this old branch

PR #1674, opened 2024-11-26
Implemented Next Step and Added By columns

### LINKED ISSUE #1637: Display username and partner on submission list entries
From Zeynep:

Could we consider adding functionality to see who added candidates to a submission list, ideally showing both the name and the partner organization? This would be really helpful for tracking actions since we have so many partners, and individuals within the same organization often work on the same submission lists. Having this information would make it easier to know who to contact if questions arise about a candidate. This feature would also benefit employer access by helping us distinguish candidates identified by employers from those identified by partners, which would support better management of next steps.

[Slack Message](https://refugeejobsmarket.slack.com/archives/C42S4AJ73/p1730914604891109)

samschlicht:
https://github.com/user-attachments/assets/551f3842-5c6b-448c-b09c-a1a18dfc0b9c


ui/admin-portal/src/app/services/candidate-field.service.ts
```diff
@@ -366,4 +399,32 @@ export class CandidateFieldService {
...
+   * @param candidate the given candidate for this table row
+   */
+  public getAddedByPartner(candidate: Candidate): string {
+    const candidateOppForThisList: CandidateOpportunity = this.findRelevantCandidateOpp(candidate);
+    return candidateOppForThisList.createdBy.partner.name;
+  }
+
+  // When displaying a submission list, will find the relevant candidate opp for given candidate.
+  private findRelevantCandidateOpp(candidate: Candidate): CandidateOpportunity {
+    return candidate.candidateOpportunities.find(
```
camerojo: Can this return null? I imagine it can - but I don't see how the calling code handles nulls/undefined

ui/admin-portal/src/app/model/candidate-field-info.ts
```diff
@@ -63,7 +65,9 @@ export class CandidateFieldInfo {
     let value;
     // Need to format field with the candidate object not value
     if (this.fieldPath === "ieltsScore" || this.fieldPath === "frenchAssessmentScoreNclc" ||
-      this.fieldPath === "latestIntake" || this.fieldPath === "latestIntakeDate") {
+      this.fieldPath === "latestIntake" || this.fieldPath === "latestIntakeDate" ||
+      this.fieldPath === "candidateOpportunities.nextStep" ||
```
camerojo: I know that it is just text "candidateOpportunities" - plural. This doesn't make sense to me. There is no "nextStep" for candidateOpportunities - there is a different next step for each opportunity.

I can't see in this PR how candidateOpportunities.nextStep is processed. Is that in another PR?

ui/admin-portal/src/app/components/candidates/show/candidate-source-base.ts
```diff
@@ -84,6 +84,10 @@ export class CandidateSourceBaseComponent {
   }
 
   protected loadSelectedFields() {
+    // Service and child components need to know the candidate source type
+    // (e.g. list (submission/regular), search) to set available fields
+    this.candidateFieldService.setCandidateSource(this.candidateSource);
```
camerojo: I don't understand the need for this. The call to candidateFieldService passes in the candidate source.

Also - this is not a good way to pass in state to a service - which can be used from multiple places. It would fail in a multithreaded environment. That may not be a problem in Angular which operates in a single thread I believe - but it is not good practice anyway.

ui/admin-portal/src/app/services/candidate-field.service.ts
```diff
@@ -27,12 +27,27 @@ import {
...
 
 @Injectable({
   providedIn: 'root'
 })
 export class CandidateFieldService {
 
+  // Dictates whether certain fields are available, e.g. Next Step and Added By, which require the
+  // source to be a submission list.
+  private candidateSource: CandidateSource;
+
```
camerojo: This "context" should be passed in with calls

camerojo (changes requested):
Let's have a chat about this when you have a chance.

ui/admin-portal/src/app/services/candidate-field.service.ts
samschlicht: Can a candidate be on a submission list but not have a candidate opp for the job that the sub list relates to? Perhaps I should put it in there as a safeguard anyway but it didn't seem necessary initially.

ui/admin-portal/src/app/model/candidate-field-info.ts
samschlicht: Because both these fields use a formatter, the field path is never used — but you're right, it's confusing. I've amended to null instead, which works fine. The next step processing is in this PR.

The existing code for using or disregarding the field path is a little hacky `CandidateFieldInfo.getValue()` - perhaps something we could mark for future attention. I think the formatter has been repurposed for something that its name doesn't quite describe.

ui/admin-portal/src/app/components/candidates/show/candidate-source-base.ts
samschlicht: Ah fair enough — it was a lot more code to pass the source without a service property, but I've read up and thought on it and fully take your point. Have pushed changes for your consideration.

ui/admin-portal/src/app/services/candidate-field.service.ts
samschlicht: 👍

ui/admin-portal/src/app/services/candidate-field.service.ts
camerojo: You are right that it shouldn't but always good to make your code bullet proof. 

Suppose that there is a bug where a candidate gets into a list when it shouldn't - it doesn't help if your code is throwing null pointer exceptions. That will make it much harder to track down the bug.

ui/admin-portal/src/app/model/candidate-field-info.ts
samschlicht: This is sorted now, per our discussion

ui/admin-portal/src/app/services/candidate-field.service.ts
samschlicht: Null handling added per discussion

ui/admin-portal/src/app/services/candidate-field.service.ts
```diff
@@ -406,15 +410,23 @@ export class CandidateFieldService {
...
+    return candidateOppForThisList === null ? '?' : candidateOppForThisList.createdBy.partner.name;
   }
 
   // When displaying a submission list, will find the relevant candidate opp for given candidate.
   private findRelevantCandidateOpp(candidate: Candidate, source: CandidateSource): CandidateOpportunity {
-    return candidate.candidateOpportunities.find(
+    const opp = candidate.candidateOpportunities.find(
       opp => opp.jobOpp.submissionList.id == source.id
     );
+    if (opp === null) {
```
camerojo: Just a note that you can use simple == with null. That has the advantage of working for both null and undefined.

As it is now an undefined opp will break your code.

camerojo (changes requested):
Very minor change - see comment

ui/admin-portal/src/app/services/candidate-field.service.ts
samschlicht: Great catch! Amended and resubmitted 👍

camerojo (approved):

PR #1677, opened 2024-11-26
1658 refactoring of some elastic search service methods

### LINKED ISSUE #1658: Refactoring of some Elastic Search Service methods
The addElasticTermsFilter currently has built in processing for an optional SearchType parameter - applying and/or/not to the result of the terms query.

This should be implemented as a sub query. It is not right the way it is - although it works for our current use of it.

camerojo:


sadatmalik (approved):
Nicely done - so much better!

PR #1701, opened 2024-12-01
1534 create a mechanism to import coupons to tc database

### LINKED ISSUE #1534: Create a mechanism to import coupons to TC database

EhsanEhrari:
Created a section in the TC settings for System Admins to import coupons as a CSV file from Duolingo dashboard into TC.


camerojo (changes requested):
Ignore previous comment

ui/admin-portal/src/app/services/duolingo-coupon.service.ts
```diff
@@ -0,0 +1,67 @@
...
+      headers: new HttpHeaders({ 'enctype': 'multipart/form-data' }),
+    });
+  }
+
+  /**
+   * Assign an available coupon to a candidate
+   * @param candidateId - ID of the candidate
+   */
+  assignCouponToCandidate(candidateId: number): Observable<any> {
+    return this.http.post<any>(`${this.apiBaseUrl}/${candidateId}/assign`, {});
```
camerojo: This is different from our standard way of handling return values and errors.

We try and avoid use of "any" - preferring to stay strongly typed where possible.

I know that this matches the server API call "{candidateId}/assign". That is really where the issue is. Our approach in the rest of our code is to return a value or throw an Exception - for example if there are no coupons available. The Angular code processes those exceptions and displays an error to the user.

I am not saying that is a better way that what you are doing. I am just pointing out that it is different and we need our code to be a consistent as possible - and also our API should be a consistent as possible.

As an example see the get method of the SavedListAdminApi. 

Exceptions thrown from API methods are handled in org/tctalent/server/util/dto/ErrorHandler.java on the server side, and ErrorInterceptor on the Angular side.

I would suggest throwing a NoSuchObjectException containing an appropriate error message (which will be displayed to the user) in the case where there are no available coupons.

In the normal case where a coupon has been assigned successfully you return the coupon details.

The same goes for the getCouponByCode API method.

camerojo (changes requested):
Great work - but see comments.

camerojo (dismissed):

camerojo (approved):

PR #1703, opened 2024-12-01
1529 use tc tasks to send the coupon to the candidate

### LINKED ISSUE #1529: Use TC tasks to send the coupon to the candidate
This task focuses on creating an interface where candidates can collect or request a coupon. Here, the user interface allows candidates to input information or interact with a form that lets them claim or generate a coupon if they're eligible.

MuhannedNoman:


ui/candidate-portal/src/app/components/profile/view/tab/tasks/task/candidate-task.component.html
```diff
@@ -33,6 +33,12 @@ <h5 class="text-alert">{{ 'TASKS.TASK.ABANDONEDDATE' | translate }} {{selectedTa
   </div>
 </div>
 
+<ng-container *ngIf="selectedTask.task.name === 'duolingoTest'">
+  <h5>You are receiving an invitation to take the Duolingo English Test for free! Accept the coupon below to get started.</h5>
+  <p>Your certified results will be sent to Talent Beyond Boundaries (TBB) after you complete the test.</p>
+  <a class="btn btn-sm btn-primary" target="_blank" [href]="'https://example.com/' + candidate.id">Accept Offer</a>
```
MuhannedNoman: Can someone point me on where can I add those to the translations? I couldn't find the translations files.

server/src/main/resources/db/migration/V1_325_add_dulingo_coupon_task.sql
cazcam34, line 1: Hey @MuhannedNoman! Just putting in a new Flyway to the Flyways branch and I saw your flyway 325. Just a small comment, so we use a strange but I think important format to the file name. You'll see we add two underscores after the number and before the title so it goes V1_325__(<-- two underscores here)add_duolingo_coupon_task.sql. It's super easy to miss but it has to be that way in order to run successfully as far as I know.

ui/candidate-portal/src/app/components/profile/view/tab/tasks/task/candidate-task.component.html
cazcam34: We don't actually translate task descriptions as they come from the database as fields. However since this is hard coded it could be translated. Translations are updated using translation keys and then we update those via the Settings on the Admin Portal. I can show you if you'd like just give me a slack @MuhannedNoman :)

server/src/main/resources/db/migration/V1_325_add_dulingo_coupon_task.sql
```diff
@@ -0,0 +1,20 @@
...
+ * FITNESS FOR A PARTICULAR PURPOSE. See the GNU Affero General Public License
+ * for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see https://www.gnu.org/licenses/.
+ */
+
+insert into task (admin, created_by, created_date, days_to_complete, description, name, optional, task_type,
+                  display_name)
+values (false, 25043, now(), 14, 'We are offering you a free Duolingo Test.', 'duolingoTest', false, 'Task',
```
cazcam34: Hi @MuhannedNoman, just while I'm on your PR I see you've used id 25043 for this created_by task field, what user does that id belong to? We have created past tasks using the System Admin user, if you see flyway # 167 for example you can see we do "(select id from users where username = 'SystemAdmin')," for the created_by field on other tasks.

ui/candidate-portal/src/app/components/profile/view/tab/tasks/task/candidate-task.component.html
cazcam34: Actually I've had a thought!! Maybe we'd have more flexibility if we made task descriptions rich text, so that we can add html to the flyway and it would display as rich text in the task? That would reduce the need to hard code text for particular tasks, which you needed to do here to add a button, and allow us to insert a link instead? What do you think? I originally wrote this tasks code before we used rich text, but now we use it in our chats and other descriptive fields, so perhaps we could use it here?

MuhannedNoman:
I am facing a side effect that I don't like here. 

The changes I made in:
`server/src/main/java/org/tctalent/server/service/db/impl/CandidateServiceImpl.java`
Lines 541–555.
Checks  if a candidate with a DuolingoTask has a coupon.
If they do, I use the email template located at:
`server/src/main/resources/mail/candidate-duolingo-coupon.html`,
inserting the coupons into the template.
I replace the current task description. For this, I added a setter function in:
`server/src/main/java/org/tctalent/server/model/db/TaskImpl.java`
`server/src/main/java/org/tctalent/server/model/db/task/Task.java`.
While this approach works, it updates the task description in the database with the latest template used which is a behaviour I don't like.

server/src/main/resources/db/migration/V1_333__add_duolingo_english_test_task.sql
```diff
@@ -0,0 +1,5 @@
+insert into task (admin, created_by, created_date, days_to_complete, description, name, optional, task_type,
+                  display_name)
+values (false, (select id from users where username = 'SystemAdmin'), now(), 14,
+        'We are offering you a free Duolingo Test.', 'duolingoTest', false, 'Task',
```
cazcam34, line 4: Where you've put task_type as 'Task' you are setting this task as a Simple Task type. If you have a look at other flyway examples of tasks you'll see that sometimes it'll be 'QuestionTask' or 'UploadTask'. Where it is just 'Task' that is task type is returned as a SimpleTask type. Not the best wording and the type should probably be more explicitly named 'SimpleTask'! See screenshot below from the Task interface where it returns task type SimpleTask.
Screenshot 2025-01-16 at 11 25 27 am

Because of this you aren't setting this task up as your new Duolingo task type, and so I'm finding it's not processing your code as it's not passing your if statements where you check if it's a Duolingo task type. Instead it is seen as a Simple task type and a Simple Task Assignment.

cazcam34: In saying this, I wonder if we can use the Simple Task type as you have created in the flyway - do we need a new task type for a Duolingo coupon task? I understand we need to insert the coupon variable into the task description, but perhaps we could find another way to get this value into the task description via rich text and insert the coupon via interpolation?

server/src/main/java/org/tctalent/server/service/db/impl/CandidateServiceImpl.java
```diff
@@ -589,6 +553,14 @@ private void populateTransientTaskAssignmentFields(TaskAssignment taskAssignment
 
         taskService.populateTransientFields(taskAssignment.getTask());
 
+        // If the task is a Duolingo coupon task, fetch the coupon content
+        if (taskAssignment instanceof DuolingoCouponTaskAssignmentImpl) {
```
cazcam34: I have assigned a duolingo task to my test candidate, however I am not passing this if statement as it seems the taskAssignment isn't a DuolingoTask type it's a Simple task type (see comment in Flyway file above).

server/src/main/java/org/tctalent/server/service/db/impl/CandidateServiceImpl.java
```diff
@@ -589,6 +553,14 @@ private void populateTransientTaskAssignmentFields(TaskAssignment taskAssignment
...
+            DuolingoCouponTaskAssignment duolingoCouponTaskAssignment = (DuolingoCouponTaskAssignmentImpl) taskAssignment;
```
cazcam34: I changed the if statement above to get past the check... I did 
`if (taskAssignment.getTask().getName().equals("duolingoTest"))`

But then I hit another error on this next line as it's not casting to the DuolingoCouponTaskAssignement, I think it could be also related to the task type being a simple task?
`java.lang.ClassCastException: class org.tctalent.server.model.db.TaskAssignmentImpl cannot be cast to class org.tctalent.server.model.db.DuolingoCouponTaskAssignmentImpl (org.tctalent.server.model.db.TaskAssignmentImpl and org.tctalent.server.model.db.DuolingoCouponTaskAssignmentImpl are in unnamed module of loader 'app')`

server/src/main/resources/mail/candidate-duolingo-coupon.html
cazcam34, line 1: I wonder could we also put this HTML (and then coupons/links) into the task description? 
We could use the existing Simple task type, and put this html into the task description which we can then display on the front end. 
Then instead of having a seperate duolingo task type, we could do a check by using the task name (duolingoTest) and if it's a duolingo task we could insert the unique coupons into the HTML on the backend before passing to the front end to display?
Love to hear your thoughts - I may be missing something as I am not as knowledgable in the Duolingo process.

server/src/main/resources/db/migration/V1_333__add_duolingo_english_test_task.sql
MuhannedNoman: @cazcam34 Yes you are correct I need to change the type here. 

As for not creating a new type, I initially tried using the description. But when I created a setter for it, the description of the task in the database kept getting the latest code sync in there witch is not how I want this to happen. That is why I created a new fake type to hold the data and not sync it with the database. 

But if there is way to use the simple task I am all for it 👍

server/src/main/resources/mail/candidate-duolingo-coupon.html
MuhannedNoman: Do you suggest having another property on the task it self to hold the new description? Or using the same description?

server/src/main/java/org/tctalent/server/model/db/TaskDtoHelper.java
```diff
@@ -53,8 +51,7 @@ public boolean ignoreProperty(Object o, String property) {
                 questionOnlyProperties.contains(property) &&
                     ! (o instanceof QuestionTask || o instanceof QuestionTaskAssignment) ||
                 uploadOnlyProperties.contains(property) &&
-                    ! (o instanceof UploadTask || o instanceof UploadTaskAssignment);
-
+                        !(o instanceof UploadTask || o instanceof UploadTaskAssignment);
```
camerojo, line 54: Minor one, keep an eye on any accidental indentations. Sometimes your Intellij may be doing it without you knowing, I can share my intellij config if you don't already have it as we try to have same config across team.

cazcam34: Minor one, keep an eye on any additional indentations that aren't standard. Sometimes your Intellij may be doing it without you knowing, I can share my intellij config if you don't already have it as we try to have same config across team.

server/src/main/java/org/tctalent/server/model/db/TaskDtoHelper.java
```diff
@@ -86,6 +83,7 @@ public static DtoBuilder getTaskDto(DtoType dtoType) {
             .add("name")
             .add("daysToComplete")
             .add("description")
+                .add("content")
```
cazcam34: Maybe just remove the extra indentation

server/src/main/java/org/tctalent/server/service/db/impl/SavedListServiceImpl.java
```diff
@@ -127,22 +111,23 @@ public class SavedListServiceImpl implements SavedListService {
...
-        CandidateOpportunityService candidateOpportunityService, ExportColumnsService exportColumnsService,
-        SavedListRepository savedListRepository,
-        DocPublisherService docPublisherService,
-        FileSystemService fileSystemService,
-        GoogleDriveConfig googleDriveConfig, PublicIDService publicIDService,
-        SalesforceService salesforceService,
-        SalesforceJobOppService salesforceJobOppService, TaskAssignmentService taskAssignmentService,
-        UserRepository userRepository,
-        UserService userService) {
+            CandidateRepository candidateRepository,
```
cazcam34: Just some extra indentation here that can be removed here and below to keep it standard

server/src/main/java/org/tctalent/server/service/db/impl/CandidateServiceImpl.java
```diff
@@ -261,6 +223,8 @@ public class CandidateServiceImpl implements CandidateService {
     private final TextExtracter textExtracter;
     private final EntityManager entityManager;
     private final PersistenceContextHelper persistenceContextHelper;
+    private final TemplateEngine htmlTemplateEngine;
```
cazcam34, line 226: This line caused a compile error for me, so I can't build your code. Did this happen for you too?:
Screenshot 2025-01-20 at 12 52 36 pm

server/src/main/java/org/tctalent/server/service/db/impl/SavedListServiceImpl.java
cazcam34: Just some more indentations you could undo to make your PR smaller (see line and see below)

cazcam34:
Thanks @MuhannedNoman for sharing with me and taking on some of my thoughts from Thursday! Unfortunately I can't build it due to the error in my code comment above. 
However, I was talking to @camerojo about what we discussed and he thinks we actually should improve the design of the tasks code in general, to make it more flexible for the Duolingo task and future tasks that may require some additional preparation (e.g. inserting unique data into the description). We can discuss further as a team on Tuesdays meeting. Great work though, this code was written 2 years ago and hasn't really been looked at since so you've done really well to figure something out and highlight the issues!!

server/src/main/java/org/tctalent/server/service/db/impl/CandidateServiceImpl.java
cazcam34: Sorry @MuhannedNoman I tried on a different computer and it worked - @camerojo remembers this error from the past and it's likely something to do with the upgrades we did to Spring boot 3 and the latest version of java.

server/src/main/java/org/tctalent/server/model/db/TaskDtoHelper.java
```diff
@@ -86,6 +86,7 @@ public static DtoBuilder getTaskDto(DtoType dtoType) {
             .add("name")
             .add("daysToComplete")
             .add("description")
+            .add("content")
```
camerojo, line 89: Don't need this now

server/src/main/java/org/tctalent/server/model/db/TaskImpl.java
```diff
@@ -49,6 +44,10 @@ public class TaskImpl extends AbstractAuditableDomainObject<Long> implements Tas
     private String name;
     private boolean optional;
 
+    @Transient
+    @Nullable
+    String content;
```
camerojo, line 49: No longer needed

server/src/main/java/org/tctalent/server/model/db/task/Task.java
```diff
@@ -87,6 +87,11 @@ public interface Task extends Auditable {
      */
     boolean isOptional();
 
+    @Nullable
+    String getContent();
+
+    void setContent(String content);
+
```
camerojo, line 94: No longer needed

server/src/main/java/org/tctalent/server/service/db/impl/CandidateServiceImpl.java
```diff
@@ -78,54 +78,12 @@
...
-import org.tctalent.server.model.db.RootRequest;
-import org.tctalent.server.model.db.SavedList;
-import org.tctalent.server.model.db.Status;
-import org.tctalent.server.model.db.SurveyType;
-import org.tctalent.server.model.db.TaskAssignmentImpl;
-import org.tctalent.server.model.db.UnhcrStatus;
-import org.tctalent.server.model.db.UploadTaskImpl;
-import org.tctalent.server.model.db.User;
-import org.tctalent.server.model.db.YesNoUnsure;
+import org.tctalent.server.model.db.*;
```
camerojo, line 81: It looks like your Intellij is not configured for our standard code style - which follows Google's standard Java style.

Here is a screen shot of my Intellij Java CodeStyle config.
I have just highlighted the Imports tab - but the whole thing is driven by selecting Google Style in the Scheme drop down.

Then we shouldn't get all these import changes.

Screen Shot 2025-01-22 at 4 15 26 pm
 config.

server/src/main/java/org/tctalent/server/service/db/impl/CandidateServiceImpl.java
```diff
@@ -589,6 +553,12 @@ private void populateTransientTaskAssignmentFields(TaskAssignment taskAssignment
 
         taskService.populateTransientFields(taskAssignment.getTask());
 
+        // If task is a duolingoTest task, fetch the coupon and add it to the instructions template
+        if (taskAssignment.getTask().getName().equals("duolingoTest")) {
+            String content = fetchCandidateDuolingoCoupon(taskAssignment);
+            taskAssignment.getTask().setContent(content);
+        }
+
```
camerojo, line 561: I don't think this is needed now.

server/src/main/resources/mail/candidate-duolingo-coupon.html
```diff
@@ -0,0 +1,108 @@
+<div>
+    <h4>Steps to take the Duolingo English Test (DET)</h4>
+    <p>Dear candidate,</p>
+    <b>
+        We hope you are doing well. You are receiving this email because we need you to take the Duolingo English Test
+        (DET).
+    </b>
```
camerojo, line 7: This is all good text. But now it will be displayed in the new DuoLingo tab of the candidate portal.

The notification email can be much simpler - just informing them that they can take the DET - directing them to login to their candidate portal for more details.

ui/admin-portal/src/app/components/candidates/view/tasks/view-candidate-tasks.component.ts
```diff
@@ -102,6 +105,31 @@ export class ViewCandidateTasksComponent implements OnInit, OnChanges {
 
   }
 
+  async assignDuolingoCouponTask() {
+    const duolingoTask = await this.taskService.listTasks().toPromise();
```
camerojo, line 109: This is not a pattern that we use. We don't use async because we don't need to when using Observables and subscribe. Look at other examples in our code where we are calling an API.

Similarly we very rarely use Promises - typically only when we are using an old library that requires us to use Promises.

ui/candidate-portal/src/app/model/task.ts
```diff
@@ -26,6 +26,7 @@ export interface Task extends Auditable {
   name: string;
   daysToComplete: number;
   description: string;
+  content: string;
```
camerojo, line 29: Not needed

camerojo (changes requested):
Hi @MuhannedNoman - I know that the new approach means that a lot of this PR is no longer relevant.

You may like to start fresh on on Issue #1825 where I assigned you.

But there is some stuff here that you might want to keep - so I have made a few comments that I hope will be helpful.

ui/admin-portal/src/app/components/candidates/view/tasks/view-candidate-tasks.component.ts
camerojo: Looking at this again @MuhannedNoman - I can see your thumbs up - but no changes that I can approve.

camerojo (commented):
Hi @MuhannedNoman - you mentioned that you made some changes in response to my requests - but looking again at this PR, I can't see your changes.

MuhannedNoman:
Closed in favor of #1833 and #1867

PR #1704, opened 2024-12-01
Redirect inactive partner URLs for branding serve and candidate assignment - now set from admin UI

samschlicht:


server/src/main/java/org/tctalent/server/service/db/impl/PartnerServiceImpl.java
```diff
@@ -212,6 +212,17 @@ public Page<PartnerImpl> searchPaged(SearchPartnerRequest request) {
...
+        if (request.getRedirectPartnerId() != null) {
+            PartnerImpl newPartner = manageRedirectPartnerAssignment(request.getRedirectPartnerId());
+            partner.setRedirectPartner(newPartner);
+
+            LogBuilder.builder(log)
+                .action("redirectInactivePartnerUrl")
+                .message("URLs identifying inactive partner " + partner.getName() +
+                    " will now redirect to " + newPartner.getName() + ".")
+                .logInfo();
+        } else partner.setRedirectPartner(null);
```
camerojo: Syntactically correct but usually not recommended.

Better to use braces for else.

See https://google.github.io/styleguide/javaguide.html#s4.1.1-braces-always-used

camerojo (changes requested):
Very trivial change

server/src/main/java/org/tctalent/server/service/db/impl/PartnerServiceImpl.java
camerojo:
```suggestion
        } else {
          partner.setRedirectPartner(null);
        }
```

camerojo (changes requested):

server/src/main/java/org/tctalent/server/service/db/impl/PartnerServiceImpl.java
samschlicht: Thanks, @camerojo, yea I can see how it would be less readable! Request change committed. 👍

camerojo (approved):

PR #1705, opened 2024-12-02
995 slow candidate search card loading times

### LINKED ISSUE #995: Slow candidate search card loading times
As a admin portal user,
When I am switching between candidates in a list and viewing them (candidate search card),
I want to speed the process and only fetch data that is editable via the search card.

In the main candidate profile, we need to fetch the latest data upon each tab change so we get the latest updated data. As this data is all editable! E.g. Fetch the latest educations/contact info/attachments whatever editable sections there are on the tab.

### Reduce API calls when viewing candidate search card
HOWEVER the candidate profile on the search card is NOT EDITABLE other than the candidate selected attachments selection. Therefore, we could reduce alot of API calls that are used in the main candidate profile as we don't need to fetch updated data when flicking between tabs. We only need to update the candidate selected attachments links, which could just bubble up via output events?? There is an 'editable' boolean input which we can use to determine if we may need to fetch an updated object, or if its not editable then we can just display what already exists on the candidate object (via the list of candidates).

Update: But we need to do the pagedSearch so that we can show the attachments with paging? Maybe we just display them without paging on the list to simplify.

cazcam34:


camerojo (approved):
Great work!

Issue #1441, opened 2024-10-03
Release 2.3.0 planning


camerojo:
As the tech team
we need to capture somewhere special manual tasks that have to be carried out for a release
so that we don't forget

Typically we will carry out these tasks progressively over time on staging, but all at once in production.

We should add the tasks as a task list as we carry them out on staging, then when we release into production, we just copy the task list, and mark them off as we complete them on production.

This type of issue starts in Review/QA so that we can always easily find it. It is only moved to Done when the production release is completed.

- Follow the [release runbook](https://github.com/Talent-Catalog/talentcatalog/wiki/Release#release-runbook)

# Staging task list
##  Pre-release:

- [x]  #1730 Add Duolingo API new environment variable to staging and production (Ehsan)

-  name: DUOLINGO_API_APISECRET,
-  value: fv7duvm328pkqz48

- [x] Create Terraform S3 bucket named tc-api-terraform-state for staging and prod
- [x] Confirm local AWS Terraform credentials
- [x] Create Mongo Atlas cluster for staging and prod
- [x]  Create Mongo database (tcapi) and collection (candidates) on staging and prod clusters
- [x] Update the Spring service config to connect to the Mongo Atlas URI (using Terraform) - staging and prod
- [x] Create tc-api TC user - staging and prod
- [x] Create tc-api search (include active, incomplete, pending, unreachable, employed)
- [x] Terraform the API service environment terraform init ;  terraform plan -out tfplan ; terraform apply "tfplan"

## Post-release:

- [x] Run Settings|AdminAPI set_public_ids - NOTE: Only needs to be run to completion once. It runs in the background and should not affect performance (configured to only use 20% of resources). It can take a couple of days to complete - but can be restarted. It does not change public_ids for entities that already have them.
- [x] Change the Duolingo Task description on production.  
- [x]  #1742 Run update SQL to populate the relocated_country_id with past relocated opps country id AND create candidate note
- [x] Run SystemAdminApi to add publicIds to all partners - set_public_ids
- [x] In Mongo Atlas -- Safe list the IP address from which the Spring service will be connecting (i.e. the NAT gateways public primary IP address - it will be visible from the VPC dashaboard sidebar)
[...]

PR #1720, opened 2024-12-08
1563 integrate the apiwebhook with tc to automatically collect test results

### LINKED ISSUE #1563: Integrate the API/Webhook with TC to automatically collect test results

EhsanEhrari:
This PR implements the integration of the Duolingo API with the Talent Catalog (TC) platform to collect the test results from Duolingo Dashboard. The following updates have been made:

**API Integration:**
- Implemented methods to fetch Duolingo dashboard results and verify test scores using the Duolingo API.
- Provided optional date filtering for dashboard results to enable targeted data retrieval.

**Service Layer:**
- Created service methods to communicate with the Duolingo API using WebClient.
- Handled error scenarios gracefully with proper logging and exception handling.

**Controller**:
        Added REST endpoints to expose the functionality:

- `/duolingo/dashboard-results`: Fetches test results from the dashboard.
- `/duolingo/verify-score: Verifies` a test score using a certificate ID and birthdate.

**Code Documentation:**

- Provided comprehensive JavaDocs for methods to ensure maintainability and clarity.


server/src/main/resources/application.yml
```diff
@@ -236,6 +236,12 @@ email:
   defaultEmail: ${EMAIL_DEFAULTEMAIL:noreply@talentcatalog.net}
   testOverrideEmail: ${EMAIL_TESTOVERRIDEEMAIL:}
 
+  # Duolingo API Configuration:
+  # Contains the API secret and base URL for interacting with the Duolingo English test API.
+duolingo:
+  api:
+    apiSecret: ${DUOLINGO_SECRET:}
```
camerojo, line 243: If you haven't already done it, can you add a connected issue to add this new environment variable to staging and production. It would be good to have a link to that issue in this issue - and also in the Release  2.3.0 Planning issue (#1441).

server/src/main/java/org/tctalent/server/configuration/DuolingoWebClentConfig.java
```diff
@@ -0,0 +1,68 @@
...
+
+/*
+ * DuolingoWebClientConfig:
+ * Configures a WebClient to interact with the Duolingo API, retrieving the API secret and base URL
+ * from application properties (duolingo.api).
+ */
+@Getter
+@Setter
+@ConfigurationProperties("duolingo.api")
+public class DuolingoWebClentConfig {
```
camerojo: Typo - Clent instead of Client

server/src/main/java/org/tctalent/server/configuration/DuolingoWebClentConfig.java
```diff
@@ -0,0 +1,68 @@
...
+    return ExchangeFilterFunction.ofRequestProcessor(clientRequest -> {
+      return Mono.just(
+          ClientRequest.from(clientRequest)
+              .headers(headers -> headers.setBasicAuth(apiSecret, ""))
+              .build()
+      );
+    });
+  }
+
+}
```
camerojo: This is absolutely correct but I would not mix configuration logic (and bean creation) with configuration property mapping (I know that we do in other parts of the TC code - but I don't think that is a good example)

For one thing it makes it difficult to unit test - because the configuration depends on running in Spring Boot.

Instead I would I would have two separate classes:

```
@ConfigurationProperties(prefix = "duolingo.api")
public class DuolingoProperties {
   ...
}

```
and

```
@Configuration
public class DuoLingoConfig {
    private final DuoLingoProperties properties;

    public DuoLingoConfig(DuoLingoProperties properties) {
        this.properties = properties;
    }

    @Bean
    public WebClient duolingoWebclient() {
       ...
    }
}

```
The second class that actually does the configuration can now be easily unit tested just by manually creating a DuoLingoProperties and using it to create a DuoLingoConfig.

I know that we have done configuration in different ways in different parts of the code - but I think this makes most sense. What do you think?

Possibly the DuoLingoProperties could be wired directly into the DuolingoServiceImpl instead of the webclient. So the actual webclient remains hidden. I am not sure if that is a good idea.

server/src/main/java/org/tctalent/server/response/DuolingoDashboardResponse.java
```diff
@@ -0,0 +1,115 @@
...
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see https://www.gnu.org/licenses/.
+ */
+
+package org.tctalent.server.response;
+
+import lombok.Data;
+/*
+ * DuolingoDashboardResponse:
+ * Represents the response object for the Duolingo Dashboard API. It contains various fields
```
camerojo, line 22: Can you add a link to the Duolingo API doc that this is taken from

server/src/main/java/org/tctalent/server/response/DuolingoVerifyScoreResponse.java
```diff
@@ -0,0 +1,43 @@
...
+ */
+
+package org.tctalent.server.response;
+
+import lombok.Data;
+
+/*
+ * The DuolingoVerifyScoreResponse class represents the response structure
+ * returned by the Duolingo API when verifying a test score. It contains
+ * various details about the applicant's test, including their name, test
```
camerojo, line 24: Link to Duolingo doc

camerojo (changes requested):
Looks great @EhsanEhrari. 

I have just made some suggestions. It would be good to standardize how we do configuration - which is a bit inconsistent at the moment. We could use this Duolingo code as our new standard.

server/src/main/java/org/tctalent/server/service/db/impl/DuolingoApiServiceImpl.java
```diff
@@ -0,0 +1,116 @@
+/*
```
camerojo, line 1: Just one more comment relating to the Duolingo database definition. (Sorry I missed this when you first submitted it).

`CREATE TABLE duolingo_coupon (
                         id BIGSERIAL NOT NULL PRIMARY KEY,
                         coupon_code VARCHAR(255) UNIQUE NOT NULL,
                         candidate_id BIGINT,
                         expiration_date TIMESTAMP NOT NULL,
                         date_sent TIMESTAMP,
                         coupon_status VARCHAR(50) NOT NULL
                         );
`

If you take a look at our other Flyways you will see that we have a standard of using timestamptz instead of just timestamp. Whether or not that is a good idea, it is our current standard.

Also when using references to other tables - eg candidate_id - it is our standard to be explicit about the table being referenced:

eg
  candidate_id bigint references candidate

camerojo (changes requested):

server/src/main/resources/application.yml
EhsanEhrari: Thank you @camerojo , Regarding add a link in release 2.3.0 Planning issue, Should I just add a the link to that issue as a comment or is there any other standard approach?

server/src/main/java/org/tctalent/server/configuration/DuolingoWebClentConfig.java
EhsanEhrari: I agree with you it seems much better and cleaner code, Where should I put the DuolingoProperties? should I create a separate folder for it or just put it inside configuration folder?

server/src/main/java/org/tctalent/server/service/db/impl/DuolingoApiServiceImpl.java
EhsanEhrari: Should I create a new migration for this? or just modifying it on migration 324?

EhsanEhrari: Or maybe creating a new SQL migration file that should include the modifications necessary to match the changes between the old and new versions of the duolingo_coupon table?

new migration maybe looks something like the below:

`ALTER TABLE duolingo_coupon
  ALTER COLUMN expiration_date TYPE TIMESTAMPTZ,
  ALTER COLUMN date_sent TYPE TIMESTAMPTZ,
  ADD CONSTRAINT fk_candidate FOREIGN KEY (candidate_id) REFERENCES candidate(id);`

camerojo (dismissed):

camerojo (approved):

PR #1748, opened 2024-12-15
1634 select unique identifier type for obfuscated ids

### LINKED ISSUE #1634: Implement unique identifier type for obfuscated IDs
For prototyping, uuid for obfuscated IDs will suffice.

However before proceeding beyond prototyping stages; care should be taken in selecting between UUIDs and, for example, Google style identifiers.

See:
https://stackoverflow.com/questions/19686430/is-google-drive-file-id-unique-globally
https://chatgpt.com/share/6733d084-5660-800f-a9df-1b920b9991d2

camerojo:
Adds new publicId field to Candidates, SavedList and SavedSearch. 
Adds to new entities when they are created. Exposes in DTOs.

See SystemAdminApi set_public_ids as one off way of setting publicIds on existing entities.

Runs in background.

### DIFF
--- a/server/src/main/java/org/tctalent/server/model/db/Candidate.java
+++ b/server/src/main/java/org/tctalent/server/model/db/Candidate.java
@@ -63,9 +63,10 @@ import org.tctalent.server.util.SalesforceHelper;
 @Table(name = "candidate")
 @SequenceGenerator(name = "seq_gen", sequenceName = "candidate_id_seq", allocationSize = 1)
 @Slf4j
-public class Candidate extends AbstractAuditableDomainObject<Long> {
+public class Candidate extends AbstractAuditableDomainObject<Long> implements HasPublicId {
 
     private String candidateNumber;
+    private String publicId;
 
     @Transient
     private Long contextSavedListId;
@@ -819,6 +820,14 @@ public class Candidate extends AbstractAuditableDomainObject<Long> {
 
     public void setCandidateNumber(String candidateNumber) {
         this.candidateNumber = candidateNumber;
+    }
+
+    public String getPublicId() {
+        return publicId;
+    }
+
+    public void setPublicId(String publicId) {
+        this.publicId = publicId;
     }
 
     /**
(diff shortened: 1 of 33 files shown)


server/src/main/java/org/tctalent/server/model/db/Candidate.java
```diff
@@ -63,9 +63,10 @@
 @Table(name = "candidate")
 @SequenceGenerator(name = "seq_gen", sequenceName = "candidate_id_seq", allocationSize = 1)
 @Slf4j
-public class Candidate extends AbstractAuditableDomainObject<Long> {
+public class Candidate extends AbstractAuditableDomainObject<Long> implements HasPublicId {
```
sadatmalik, line 66: Recommendation generated by Amazon CodeGuru Reviewer. Leave feedback on this recommendation by replying to the comment or by reacting to the comment using emoji.

This class references 52 other classes. By comparison, 99% of the classes in the CodeGuru reference dataset reference fewer. This indicates that this class is highly coupled with other classes. A class that is highly coupled with other classes is difficult to understand and its behavior might change unexpectedly when one of its referenced classes is updated. High coupling could also increase the integration test complexity, maintenance cost and technical debt. We recommend that you simplify this class or break it into multiple classes.

sadatmalik, line 66: Recommendation generated by Amazon CodeGuru Reviewer. Leave feedback on this recommendation by replying to the comment or by reacting to the comment using emoji.

This class contains 12 clusters of instance methods that do not have any accessed class members in common. For example, a cluster might have 2 methods that access only the class fields x and y, and another cluster might have 2 other methods that access only the class fields a and b. By comparison, 98% of the classes in the CodeGuru reference dataset contain fewer clusters. A high number of these clusters indicates low class cohesion. Classes with low class cohesion contain unrelated operations which make them difficult to understand and less likely to be used. We recommend that you simplify this class or break it into multiple cohesive classes.

sadatmalik (approved):

PR #1760, opened 2024-12-19
1739 rename tbbshortname to tcshortname

MuhannedNoman:
If you think this is ready to be merged, I can take out the duolingo migration and rearrange numbers.


camerojo (approved):
Looks good Mohanned!

PR #1774, opened 2025-01-07
1741 new relocated address fields updatable via candidate portal

### LINKED ISSUE #1741: New Relocated Address fields - Updatable via Candidate Portal
See issue #859 where fields were added

cazcam34:


cazcam34:
A broken test that seems to be due to the new year, so the age that is calculates (based on 1st Jan 90) has increased so it doesn't match the expected value. Fixed but increasing the age by a year, but I wonder if there is a better way to do this?
Screenshot 2025-01-07 at 3 23 47 pm


Screenshot 2025-01-07 at 3 19 00 pm

cazcam34:
One more broken test due to the new year, that I've fixed by changing the year. See below with both dates now being past (updated to 2026 instead of 2025). Will bring up in meeting tonight for another solution potentially:
Screenshot 2025-01-07 at 3 35 23 pm

camerojo (dismissed):

camerojo (approved):

PR #1804, opened 2025-01-13
Check that only employer job opps are used in SF new job link

### LINKED ISSUE #1795: Bug: user can create new job with SF candidate opportunity
The link that users provide will accept a candidate opportunity, which gets published as a Job and then can't be taken much further. Perhaps an unlikely event but liable to create work if it happens.

cazcam34:


camerojo (dismissed):

samschlicht:
Nice work, @cazcam34! I agree it's pretty obscure but I think it's definitely worth having - glad it wasn't too much code though. 🙂

camerojo (approved):

PR #1833, opened 2025-01-22
1589 add a button to manually trigger duolingo test communications

MuhannedNoman:


server/src/main/java/org/tctalent/server/service/db/impl/SavedListServiceImpl.java
```diff
@@ -747,6 +752,14 @@ public void associateTaskWithList(User user, TaskImpl task, SavedList list) {
...
         //Now assign tasks to candidates in list (if they do not already have the task actively assigned)
         Set<Candidate> candidates = list.getCandidates();
+        if (task.getName().equals("duolingoTest")) {
+            List<DuolingoCoupon> availableCoupons = couponService.getAvailableCoupons();
+            if (candidates.size() > availableCoupons.size()) {
+                throw new NoSuchObjectException(
+                    availableCoupons.size() + " coupons available, but you need "
+                        + candidates.size() + " coupons for your candidates list.");
+            }
+        }
```
camerojo: I don't think that this check belongs here.

Putting here builds a dependency between the SavedListService and the Coupon service. It also makes this method throw an Exceptions which it never did before (and is undocumented in the interface).

This code which is just an error check really would be much better placed in the caller: TaskAssignmentAdminApi.

That will add a dependency on CouponService to the TaskAssignmentAdminApi - but I think that is preferable to adding it to the already complex and core SavedListService.

server/src/main/java/org/tctalent/server/service/db/impl/TaskAssigmentServiceImpl.java
```diff
@@ -84,6 +88,9 @@ public TaskAssignmentImpl assignTaskToCandidate(
 
             default:
                 taskAssignment = new TaskAssignmentImpl();
+                if (task.getName().equals("duolingoTest")) {
+                    couponService.assignCouponToCandidate(candidate.getId());
+                }
```
camerojo: This creates a dependency between TaskAssignment and Coupons. See my comments on the SavedListServiceImpl above. 

I think a better approach which keeps services much more loosely coupled is to do two API calls when a button is pressed - one which assigns the coupon(s) and one which assigns the tasks.

Alternately - do it all through the DuoLingoCouponAdminApi. When a coupon is assigned, a task is created notifying the candidate(s).

It makes more sense for the DuoLingoCouponService to depend on the TaskService than to have the TaskService depend on the DuoLingoCouponService.

ui/admin-portal/src/app/components/candidates/view/tasks/view-candidate-tasks.component.ts
```diff
@@ -102,6 +105,37 @@ export class ViewCandidateTasksComponent implements OnInit, OnChanges {
...
+
+        if (!duolingoTaskId) {
+          this.error = 'Duolingo English Test task not found';
+          return;
+        }
+
+        const request: CreateTaskAssignmentRequest = {
+          candidateId: this.candidate.id,
+          taskId: duolingoTaskId,
+          dueDate: moment().add(2, 'weeks').toDate()
```
camerojo: DueDate can default to what is set in the task

ui/admin-portal/src/app/components/candidates/view/tasks/view-candidate-tasks.component.html
```diff
@@ -26,10 +26,17 @@
...
     <div class="d-flex flex-fill justify-content-between align-items-center">
       <span class="small font-italic">Comment?</span>
-      <div *ngIf="editable">
-        <button class="btn btn-primary btn-sm" (click)="assignTask()">
-          <i class="fas fa-plus"></i> Assign task
-        </button>
+      <div class="d-flex">
+        <div *ngIf="editable" class="mx-2">
+          <button class="btn btn-primary btn-sm" (click)="assignDuolingoCouponTask()">
+             Send Duolingo coupon
```
camerojo: I am not sure that this is the best place for the button. Users won't think to go to the Candidate's Tasks tab to assign a Duolingo coupon. I think we need a button somewhere more obvious on the main candidate screen - not hidden away in one of the tabs. Have the users given us any feedback where they would like the button?

camerojo (changes requested):
I have made a couple of suggestions around where the code should be.

But I am also unsure if this is the best place to have the assign coupon button. Is this where the users wanted the button located?

server/src/main/java/org/tctalent/server/service/db/impl/DuolingoCouponServiceImpl.java
```diff
@@ -172,6 +174,28 @@ public Candidate findCandidateByCouponCode(String couponCode) throws NoSuchObjec
...
+
+    if (availableCoupons.isEmpty() || candidates.size() > availableCoupons.size()) {
+      // Throw exception if no coupon are available, or if there are more candidates than coupons
+      throw new NoSuchObjectException("No available coupons for list ID " + list.getId());
+    }
+
+    for (Candidate candidate : candidates) {
+      //Assign coupon to candidate if they do not have one
+      List<DuolingoCoupon> coupons = couponRepository.findAllByCandidateId(candidate.getId());
+      if (coupons.isEmpty()) {
```
camerojo: What if candidate already was assigned a coupon a year ago? They failed. Now they are trying again. 

This code will make it impossible for them to be issued another coupon.

ui/admin-portal/src/app/components/candidates/view/tasks/view-candidate-tasks.component.ts
```diff
@@ -102,6 +106,43 @@ export class ViewCandidateTasksComponent implements OnInit, OnChanges {
 
   }
 
+  assignDuolingoCouponTask() {
+    this.duolingoCouponService.assignCouponToCandidate(this.candidate.id).subscribe(
+      () => {
+        this.taskService.listTasks().subscribe(
+          tasks => {
+            const duolingoTaskId = tasks.find(task => task.name === 'duolingoTest')?.id;
```
camerojo: I think it will all work better and be simpler if the task assignment is automated on the server end whenever a candidate is assigned a coupon.

Then all this code goes - moved to the DuolingoCoupon server code. Maybe in the DuolingoCouponAdminApi or in the DuolingoCouponService.

Then we just have a Assign Duolingo Coupon button - and the server code can handle the way the candidate is notified - task, email - or both.

By the way where does the email notification get sent in the current code?

server/src/main/java/org/tctalent/server/api/admin/DuolingoCouponAdminApi.java
```diff
@@ -118,5 +127,14 @@ public DuolingoCouponResponse getCouponByCode(@PathVariable("couponCode") String
     );
   }
 
+  // Endpoint to assign available coupons to a saved list candidate
+  @PostMapping("assign-to-list")
+  public void assignCouponToList(@Valid @RequestBody Long listId)
```
camerojo:
```suggestion
  public void assignCouponsToList(@Valid @RequestBody Long listId)
```

server/src/main/java/org/tctalent/server/service/db/DuolingoCouponService.java
```diff
@@ -105,4 +107,13 @@ public interface DuolingoCouponService {
    * @return an Optional containing the Candidate if found, otherwise empty.
    */
   Candidate findCandidateByCouponCode(String couponCode) throws NoSuchObjectException;
+
+  /**
+   * Assign available coupons to candidates in a saved list.
+   * <p/>
+   * This method assigns available coupons to candidates in a saved list.
+   *
+   * @param list the SavedList containing the candidates to whom coupons should be assigned.
```
camerojo, line 116:
```suggestion
   * @param list the SavedList containing the candidates to whom coupons should be assigned.
   * @throws NoSuchObjectionException if there are not enough coupons available
```

camerojo (changes requested):
Hi @MuhannedNoman - this is a big improvement. It looks a lot simpler, don't you think? And it is more loosely coupled with no extra dependencies being added to the code.

But I still have a few changes. The main one is that I think that the server should now decide how the candidate is notified of a coupon being assigned to them - not the Angular code. That will greatly simplify the Angular code. (I think I might have suggested earlier that Angular should have this logic - but now with your greatly simplified code, I think that the logic fits better on the server - and will lead us to the absolute simplest solution).

It also means that the new Angular button is simply to assign a coupon (not a task). Maybe two buttons: one to assign a coupon to a candidate, and one to assign coupons to a list of candidates. 

Thinking more about it, I think that the right place to add these buttons might be in Settings - but we need to run that past Zeynep.

Can we arrange a Zoom session to go over this together - and completely finish it off.

ui/admin-portal/src/app/components/util/duolingo-assignment/duolingo-assignment.component.ts
```diff
@@ -0,0 +1,47 @@
...
+  ) {
+  }
+
+  assignDuolingoCouponTask() {
+    this.duolingoCouponService.assignCouponToCandidate(this.candidate.id).subscribe(
+      () => {
+        this.candidateService.updateCandidate();
+      },
+      error => {
+        this.errorOccurred.emit(error);
```
camerojo, line 43: Not sure we need to output the error. The component can display any error itself - in its own conditional html beside or under the button. That html could display the "working..." or a busy icon like elsewhere in the code, then that displays "Done" or an error on completion.

ui/admin-portal/src/app/components/tasks/assign-tasks-candidate/assign-tasks-candidate.component.ts
```diff
@@ -99,30 +99,26 @@ export class AssignTasksCandidateComponent implements OnInit {
     if(task.name === 'duolingoTest') {
```
camerojo, line 99: Do we still need this code? If we have a specific button for assigning coupons.

ui/admin-portal/src/app/components/tasks/assign-tasks-list/assign-tasks-list.component.ts
```diff
@@ -117,31 +117,28 @@ export class AssignTasksListComponent implements OnInit {
     if(task.name === 'duolingoTest') {
```
camerojo, line 117: Use the button passing a list to avoid all this special "if task.name === duolingoTest" stuff

The big change here is that assigning a coupon automatically creates the task - rather than having a special task that also assigns the coupon. The task for them is to use the coupon. 

If a user is given a task to use a coupon and they haven't been assigned one, they can report back that they can't use a coupon because they haven't been assigned one.

camerojo (changes requested):
I think that this code can still be significantly simplified. In particular I don't see need to change the original AssignTasksCandidate, AssignTasksList components.

I would also prefer to move the button away from ViewCandidateTasks so that component can go back to its original state.

Let's talk with Zeynep about where to put the button. Given that it is an admin only function, I think that it could go in Settings in the existing DuoLingo Coupons tab.

Currently there is just code there where they can upload a coupon.

We could add code where they can select a candidate or a list to assign coupons to.

ui/admin-portal/src/app/components/util/duolingo-assignment/duolingo-assignment.component.ts
MuhannedNoman: The button placement doesn't allow that to be visible, that is why I have to out put it and make the parent component display that.

ui/admin-portal/src/app/components/tasks/assign-tasks-candidate/assign-tasks-candidate.component.ts
MuhannedNoman: Yes, as of now assigning a task via the traditional method won't grant you a coupon. You will get the instructions that you have a coupon but there won't be any assigned you. This code triggers the assignment of coupon and task. 

If we only use the buttons to assign coupons I can hide the option to assign the task alone since that doesn't make much since and will cause confusion.

ui/admin-portal/src/app/components/tasks/assign-tasks-list/assign-tasks-list.component.ts
MuhannedNoman: Just to double check, if I removed this code the admins might assign a task to the user without a coupon. 
Is that behaviour okay for now?

MuhannedNoman:
> I think that this code can still be significantly simplified. In particular I don't see need to change the original AssignTasksCandidate, AssignTasksList components.
> 
> I would also prefer to move the button away from ViewCandidateTasks so that component can go back to its original state.
> 
> Let's talk with Zeynep about where to put the button. Given that it is an admin only function, I think that it could go in Settings in the existing DuoLingo Coupons tab.
> 
> Currently there is just code there where they can upload a coupon.
> 
> We could add code where they can select a candidate or a list to assign coupons to.

I got a reply: 

> Hey Muhannad! I spoke to Mariam and we both agreed that this placement looks great and is what we had in mind. It is hidden yet accessible enough so that people won’t randomly use it but also can easily access it from a candidate’s profile.

camerojo (approved):
In the interests of getting this completed, if it works let's get it out there.

So approving it. Please go ahead and merge.

PR #1850, opened 2025-01-29
1820-investigate-usage-spikes

sadatmalik:
This PR:

- Adds unit tests for memory and cpu metrics to check micrometer gauges
- Updates the calculation of memory utilisation using aggregated gauge values for all memory spaces


camerojo (dismissed):
Nice work!

camerojo (approved):

PR #1858, opened 2025-01-30
1852 for clarity heatmaps to work we need url params for different tabs can we easily add url params using cached active tab value

### LINKED ISSUE #1852: For Clarity heatmaps to work we need URL params for different tabs. Can we easily add url params using cached active tab value
As a developer,
I would like to see where users are clicking on the TC despite it being a single page angular app,
So that I can analyse the users behaviour on each unique tab

To do this with clarity we need to differentiate between tabs by using the URL. We currently store the lastActiveTab as a browser cached value so that we can return to the last active tab. If we already calculate this value could we just put this easily into the URL via a param?

cazcam34:


camerojo (dismissed):
Nice work!

camerojo (approved):

PR #1861, opened 2025-01-31
Don't exclude candidates associated with non-active partners, but res…

### LINKED ISSUE #1732: Candidates belonging to any partner whose status not active are hidden from search results
doSearchCandidates uses this method to add default partner filter criteria to all searches - so no user will see a candidate assigned to a partner who is not a source partner or has an active status.

While true that no candidate should be assigned to such partners, I found 5 who are, hence they currently have no chance of appearing in search results:

Kyaw, ThantZin, 21kyawthantzin@gmail.com
Thantzin, Lurn, thantzinlurn@gmail.com
TUN, KHANT NYAR, tunkhantnyar790@gmail.com
Sabera,Sayeedi,saberasayeedi55@gmail.com
Khun Nge, Kyi, yairkyi1999.bit@gmail.com

First task is obviously to assign them to active partners, but I would also just put to the group whether this is entirely desirable behaviour? (I haven't fully thought it through at time of writing and I want to finish my current issue before I do.) If it is, then we might add least add some sort of failsafe mechanism so that no future candidate could get lost in this unlikely and unintended but still possible predicament.

cazcam34:
…trict other methods to active only partners

### DIFF
--- a/server/src/main/java/org/tctalent/server/service/db/impl/PartnerServiceImpl.java
+++ b/server/src/main/java/org/tctalent/server/service/db/impl/PartnerServiceImpl.java
@@ -162,6 +162,7 @@ public class PartnerServiceImpl implements PartnerService {
     public List<PartnerImpl> listSourcePartners() {
         SearchPartnerRequest request = new SearchPartnerRequest();
         request.setSourcePartner(true);
+        request.setStatus(Status.active);
         return search(request);
     }
 
--- a/server/src/main/java/org/tctalent/server/service/db/impl/PartnerServiceImpl.java
+++ b/server/src/main/java/org/tctalent/server/service/db/impl/PartnerServiceImpl.java
@@ -162,7 +162,6 @@ public class PartnerServiceImpl implements PartnerService {
     public List<PartnerImpl> listSourcePartners() {
         SearchPartnerRequest request = new SearchPartnerRequest();
         request.setSourcePartner(true);
-        request.setStatus(Status.active);
         return search(request);
     }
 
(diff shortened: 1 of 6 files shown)


server/src/main/java/org/tctalent/server/service/db/impl/PartnerServiceImpl.java
```diff
@@ -162,7 +162,6 @@ public List<PartnerImpl> listPartners() {
     public List<PartnerImpl> listSourcePartners() {
         SearchPartnerRequest request = new SearchPartnerRequest();
         request.setSourcePartner(true);
-        request.setStatus(Status.active);
```
camerojo, line 165: I don't think we can just remove this. If you check who calls this method some want to just see active partners which is what this method is supposed to do according to its doc in PartnerService:

     * Convenience method which just delegates to {@link #search(SearchPartnerRequest)} with an
     * appropriate request.
     * @return All active source partners
     */
    List<PartnerImpl> listSourcePartners();

It specifically is intended to return active partners. You can't change that without breaking something.

Instead I suggest renaming this method to listActiveSourcePartners - which makes it clearer.

Then add a new method on the service listAllSourcePartners - which will do a search where request.setStatus is not called.

camerojo (changes requested):
Nice picking up the chats not being created for inactive partners!

But we can't change the meaning of the existing service method - see my comments/

server/src/main/java/org/tctalent/server/service/db/impl/PartnerServiceImpl.java
cazcam34: Renamed to listActiveSourcePartners and created new method listAllSourcePartners

camerojo (approved):

PR #1867, opened 2025-02-05
Feat: add services tab to profile view

MuhannedNoman:
This will allow us to display services and show instructions for the duolingo coupon.


MuhannedNoman:
@hiba-machfej Can you take a look and let me know what you think?

hiba-machfej:
Hey @MuhannedNoman 
Thank you do much for your work! I like that it is similar to the task experience, which candidate users are already familiar with, so it won't be something completely new for them to adapt to.

I just have a few small comments after taking a closer look in the browser and checking the AWS Amplify preview link. Nothing urgent, and maybe they can wait until we have more services—you know better regarding this:

1- It might be useful to add extra fields to the table in the future if you find it necessary, such as the Duolingo coupon deadline or issue date.

2-Regarding the services icon on the mobile view:
Since most of the candidate portal users might be on mobile, I would suggest just reducing the margin between icons to around .5rem instead of 1 rem (I updated this using Chrome DevTools):
Screenshot 2025-02-09 at 4 24 21 PM
It currently looks like this on some mobile devices: 
Screenshot 2025-02-09 at 4 24 45 PM

ui/candidate-portal/src/app/services/duolingo-coupon.service.spec.ts
```diff
@@ -0,0 +1,45 @@
+import { TestBed } from '@angular/core/testing';
+
+import { DuolingoCouponService } from './duolingo-coupon.service';
```
EhsanEhrari: For better consistency with TC project’s style guide, could you please use the Google style formatting "intellij-java-google-style.xml" as mentioned in the readme for imports? Specifically, we aim to avoid extra spaces between curly braces and module names. This helps prevent IntelliJ from automatically removing the space and generating unnecessary commits when the file is opened.

ui/candidate-portal/src/app/components/profile/view/tab/services/duolingo-coupon/duolingo-coupon.component.ts
```diff
@@ -0,0 +1,45 @@
...
+  ngOnInit(): void {
+    this.fetchCoupons();
+  }
+
+  fetchCoupons() {
+    this.loading = true;
+    this.duolingoCouponService.getCouponsForCandidate(this.candidate.id).subscribe(
+      coupons => {
+        this.coupons = coupons.filter(coupon => coupon.duolingoCouponStatus === 'SENT').map(coupon => coupon.couponCode);
+        this.loading = false;
```
EhsanEhrari: You can use the DuolingoCouponStatus enum here instead of 'SENT'

EhsanEhrari (dismissed):

ui/candidate-portal/src/app/components/profile/view/view-candidate.component.ts
```diff
@@ -124,6 +127,11 @@ export class ViewCandidateComponent implements OnInit {
     this.fetchAllOpportunityChats();
   }
 
+  private setDuolingoTask(candidate: Candidate) {
+    this.duolingoTask = this.candidate?.taskAssignments.find(t => t.task.name === "duolingoTest" && t.status === Status.active);
+    console.log(this.duolingoTask)
```
EhsanEhrari, line 132: Can you please add the claimCouponButton task here as well, because we first assign the claimCouponButton task and then once a candiadte click on the button the claimCouponButton task will complete and new duolingoTest task will assign to candidate.

EhsanEhrari: @hiba-machfej Please ignore this comment now.

EhsanEhrari (approved):
All good to me!

PR #1871, opened 2025-02-06
Include audit stamp on SF Opp Next Step updates

### LINKED ISSUE #1856: TC Job and Candidate Opp Next Step change doesn't include audit stamp on SF
This means that on SF we wouldn't know who updated the Next Step. Not a showstopper but discovered in course of #1729 (which probably needs to be merged before this is tackled) and probably worth a quick fix, in my view.

samschlicht:


server/src/main/java/org/tctalent/server/service/db/impl/NextStepProcessingServiceImpl.java
```diff
@@ -0,0 +1,55 @@
...
+        // Some updates may be automated, so we attribute these to SystemAdmin
+        User userForAttribution = userService.getLoggedInUser();
+        if (userForAttribution == null) {
+            userForAttribution = userService.getSystemAdminUser();
+        }
+
+        String currentNextStep =
+            opp == null ? null : opp.getNextStep();
+
+        final String processedNextStep = auditStampNextStep(
```
sadatmalik, line 45: Recommendation generated by Amazon CodeGuru Reviewer. Leave feedback on this recommendation by replying to the comment or by reacting to the comment using emoji.

**Problem:**
 This line of code lacks validation when processing input data through the following parameter: &#39;nextStep&#39; (index: 1 | type: String). The parameter is exposed to external callers, because its enclosing class and method are publicly accessible. This means that upstream validation, if it exists, can be bypassed. Other validated parameters: &#39;opp&#39;. The same parameter type is validated here for example: talentcatalog/server/src/main/java/org/tctalent/server/service/db/impl/JobServiceImpl.java:430. Malicious, malformed, or unbounded inputs can cause unexpected runtime behavior or crashes, and can slow performance.

**Fix:**
 Add checks to ensure the validity of the parameter&#39;s value, such as testing it for nullness (for example, using the @NonNull annotation as [described in the Lombok library](https://projectlombok.org/features/NonNull)), emptiness, or equality. Or to prevent direct calls to it, [reduce the method&#39;s visibility](https://docs.oracle.com/javase/tutorial/java/javaOO/accesscontrol.html).

 Learn more about potential threats and guidance from the [Common Weakness Enumeration website](https://cwe.mitre.org/data/definitions/693.html) and the [OWASP Cheat Sheet series](https://cheatsheetseries.owasp.org/cheatsheets/Input_Validation_Cheat_Sheet.html).

camerojo: I found the Code Guru comment a bit hard to understand but I think that it is just complaining that you have annotated the opp parameter as nullable but haven't said anything about the nextStep parameter.

In fact if you look at the auditStampNextStep method call annotations, you will see that nextStep can be null. So I would add @Nullable to the nextStep parameter and I think that will keep Code Guru happy.

See suggestion above.

server/src/main/java/org/tctalent/server/service/db/impl/NextStepProcessingServiceImpl.java
```diff
@@ -0,0 +1,55 @@
...
+import org.tctalent.server.model.db.User;
+import org.tctalent.server.service.db.NextStepProcessingService;
+import org.tctalent.server.service.db.UserService;
+
+@RequiredArgsConstructor
+@Service
+public class NextStepProcessingServiceImpl implements NextStepProcessingService {
+    private final UserService userService;
+
+    public String processNextStep(@Nullable AbstractOpportunity opp, String nextStep) {
```
camerojo:
```suggestion
    public String processNextStep(@Nullable AbstractOpportunity opp, @Nullable String nextStep) {
```

camerojo (changes requested):
See changes which should keep Code Guru happy.

camerojo (approved):

PR #1880, opened 2025-02-08
1854 update coupon status when used or expired

### LINKED ISSUE #1854: Update Coupon Status When Used or Expired
Currently, coupon statuses are not updated after use or expiration. To improve tracking and prevent misuse, we need to implement logic that updates the coupon status:

- [x] 1. When a candidate uses a coupon, update its status to "Redeemed" so it cannot be reused.
- [x] 2. When a coupon expires, update its status to "Expired" to prevent further use.

EhsanEhrari:
This PR ensures that coupon statuses are properly updated to prevent misuse. Now:

- When a candidate uses a coupon, its status is set to "Redeemed", preventing reuse.
- When a coupon reaches its expiration date, its status is automatically updated to "Expired" to prevent further use.

This improves tracking and ensures that only valid coupons can be applied.


server/src/main/java/org/tctalent/server/api/admin/SystemAdminApi.java
```diff
@@ -356,6 +361,12 @@ public void updateCandidateExams() throws NoSuchObjectException {
     duolingoExamService.updateCandidateExams();
   }
 
+  @Scheduled(cron = "0 0 0 * * ?", zone = "GMT")
```
camerojo: I know that @samschlicht has put some scheduled tasks in here - but I have checked with him and there was a special reason and he will be moving them somewhere else in future.

SystemAdminApi is a REST Controller - designed to handle HTTP requests. 
See annotations:
@RestController
@RequestMapping("/api/admin/system")

If you look at other scheduled tasks in the code you will see that they are normally set in the relevant service. For example, see notifySearchWatchers in SavedSearchServiceImpl. In this case the relevant service would be DuolingoCouponService.

Can you move it to that service's implementation: DuolingoCouponServiceImpl (note that it does not have to be in the service interface. It just needs to be a public method in the implementation).

camerojo (changes requested):
I think that the scheduled task is better to be defined in the DuolingoService.

I am not sure why Sam has been scheduling tasks from inside SystemAdminApi. Maybe there is a reason relating to those particular tasks.

server/src/main/java/org/tctalent/server/api/admin/SystemAdminApi.java
EhsanEhrari: Thank you @camerojo for the requested changes.

camerojo (approved):

PR #1881, opened 2025-02-09
1789 add support for user email verification

hiba-machfej:


ui/admin-portal/src/app/components/account/verify-email/verify-email.component.ts
```diff
@@ -0,0 +1,170 @@
...
+
+  verifyToken(token: string) {
+    const request = { token: token };
+    this.userService.checkEmailVerificationToken(request).subscribe(
+        () => this.handleTokenVerificationSuccess(request),
+        (error) => this.handleTokenVerificationError(error)
+    );
+  }
+
+  handleTokenVerificationSuccess(request: { token: string }) {
```
camerojo: Methods like this one which are never used externally should be private.

ui/admin-portal/src/app/components/account/verify-email/verify-email.component.ts
```diff
@@ -0,0 +1,170 @@
...
+    console.log("before opening modal", this.state);
+    const modalElement = document.getElementById('verifyEmailModal');
+    if (modalElement) {
+      this.bootstrapModal = new bootstrap.Modal(modalElement);
+      this.bootstrapModal.show();
+    }
+  }
+
+  getUserEmail() {
+    this.userService.get(this.user.id).subscribe(
```
camerojo: You don't need the server to fetch the user (here and in other parts of this code). You already have the user details in "user"

ui/admin-portal/src/app/components/account/verify-email/verify-email.component.ts
```diff
@@ -0,0 +1,170 @@
...
+      }
+    }
+    this.router.navigate(['/jobs']);
+  }
+
+  sendVerifyEmail(token: string) {
+    this.state = 'loading';
+    const req: SendVerifyEmailRequest = new SendVerifyEmailRequest();
+    req.reCaptchaV3Token = token;
+    this.userService.get(this.user.id).subscribe(
```
camerojo: Why are you fetching the user?

ui/admin-portal/src/app/components/account/verify-email/verify-email.component.ts
```diff
@@ -0,0 +1,170 @@
...
+        modalBackdrop.remove();
+      }
+    }
+    this.router.navigate(['/jobs']);
+  }
+
+  sendVerifyEmail(token: string) {
+    this.state = 'loading';
+    const req: SendVerifyEmailRequest = new SendVerifyEmailRequest();
+    req.reCaptchaV3Token = token;
```
camerojo: Why is this field called reCaptchaV3Token?

server/src/main/java/org/tctalent/server/service/db/UserService.java
```diff
@@ -76,6 +80,14 @@ public interface UserService {
     void updatePassword(UpdateUserPasswordRequest request);
     void updateUserPassword(long id, UpdateUserPasswordRequest request);
 
+
+    // Generates an email verification token and sends a verification email.
```
camerojo: You should used standard JavaDoc comments like elsewhere in the code.
For example.

```suggestion
    /**
     * Generates an email verification token and sends a verification email.
     */
```

server/src/main/java/org/tctalent/server/service/db/impl/UserServiceImpl.java
```diff
@@ -522,6 +541,64 @@ public User getSystemAdminUser() {
         return findByUsernameAndRole(SystemAdminConfiguration.SYSTEM_ADMIN_NAME, Role.systemadmin);
     }
 
+    @Override
+    @Transactional(rollbackFor = Exception.class)
```
camerojo: rollbackFor is not needed here - what made you add it? Spring will automatically roll back when exceptions occur.

server/src/main/java/org/tctalent/server/request/user/emailverify/SendVerifyEmailRequest.java
```diff
@@ -0,0 +1,31 @@
...
+import lombok.Getter;
+import lombok.Setter;
+import lombok.ToString;
+
+@Getter
+@Setter
+@ToString
+public class SendVerifyEmailRequest {
+    private String email;
+    private String verificationToken;
```
camerojo: I don't see how verificationToken is ever used

server/src/main/java/org/tctalent/server/request/user/emailverify/SendVerifyEmailRequest.java
```diff
@@ -0,0 +1,31 @@
...
+    private String reCaptchaV3Token;
```
camerojo: I don't see where ReCaptchaTokens are used

ui/admin-portal/src/app/components/account/verify-email/verify-email.component.html
```diff
@@ -0,0 +1,100 @@
+<div class="modal fade" id="verifyEmailModal" tabindex="-1" aria-labelledby="exampleModalLabel" aria-hidden="true">
+    <div class="modal-dialog modal-dialog-centered">
```
camerojo: This is a very non standard way of doing modals in Angular.

Our standard approach is to use NgbModal. See lots of examples in our code - plus here https://ng-bootstrap.github.io/#/components/modal/examples

server/src/main/java/org/tctalent/server/api/admin/UserAdminApi.java
```diff
@@ -155,6 +178,9 @@ private DtoBuilder userDto() {
                 .add("usingMfa")
                 .add("mfaConfigured")
                 .add("partner", PartnerDtoHelper.getPartnerDto())
+                .add("emailVerified")
+                .add("emailVerificationToken")
```
camerojo: I don't understand why the Angular code needs to see or process the token or its issuedDate. All that can be managed by the server.

camerojo (changes requested):
Hi Hiba - your code looks very nicely and carefully constructed but I think that there are some approaches that could be different.
Let's arrange some time together to go through my suggestions.

server/src/main/java/org/tctalent/server/exception/InvalidEmailVerificationTokenException.java
```diff
@@ -0,0 +1,27 @@
...
+ * FITNESS FOR A PARTICULAR PURPOSE. See the GNU Affero General Public License
+ * for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see https://www.gnu.org/licenses/.
+ */
+
+package org.tctalent.server.exception;
+
+public class InvalidEmailVerificationTokenException extends RuntimeException {
```
camerojo: Why doesn't this extend ServiceException - like all our other Exceptions?

Same for VerifyEmailSendFailedException

server/src/main/java/org/tctalent/server/service/db/impl/UserServiceImpl.java
```diff
@@ -522,6 +541,64 @@ public User getSystemAdminUser() {
...
+    public void checkEmailVerificationToken(CheckEmailVerificationTokenRequest request) {
+        User user = userRepository.findByEmailVerificationToken(request.getToken());
+
+        if (user == null) {
+            throw new InvalidEmailVerificationTokenException();
+        } else if (OffsetDateTime.now().isAfter(user.getEmailVerificationTokenIssuedDate().plusHours(24))) {
+            throw new ExpiredEmailTokenException();
+        }
+
+        log.info("Email verification token is valid.");
```
camerojo: If the token verifies, shouldn't you set the user's email verified to true?

server/src/main/java/org/tctalent/server/service/db/impl/UserServiceImpl.java
camerojo: See also rollbackFor that you have used elsewhere

ui/admin-portal/src/app/app-routing.module.ts
```diff
@@ -224,6 +225,11 @@ const routes: Routes = [
       title: 'TC Reset Password'
     }
   },
+  {
+    path: 'verify-email',
```
camerojo: There is no need to process this through the Angular code - it can go straight to the server.

ui/candidate-portal/src/app/app-routing.module.ts
```diff
@@ -71,6 +72,11 @@ const routes: Routes = [
     canActivate: [AuthGuard],
     title: 'Edit Profile'
   },
+  {
+    path: 'verify-email',
```
camerojo: I don't think that the Angular code needs to intercept this - it can go straight to the server which can check the token and if it matches, mark the matching user's email as validated.

I understand why you did it this way - because it is like the Reset Password code that I referred you to. However that code is a bit different. With the reset password, when someone requests a password reset and their token matches you want them to be in the Angular code so that they can enter their new password.

However, with email validation there is no need for the Angular code to be involved in validating the email. If a correct token is supplied, we can just mark the email on the database as validated. That request can go straight to the server API - no need to go through Angular code.

The link that they click in the email they receive can be something like:

https://tctalent.org/api/admin/user/verify-email/<token\>

Let's work on this together. I think it will work out to be much simpler. And also interesting for you.

camerojo (changes requested):
More changes. I think you followed the Reset Password code that I referred you to a bit too closely. There is no need to get the Angular code involved at all in the email verification process. See my recent comments.

ui/candidate-portal/src/app/app-routing.module.ts
camerojo: The code changes will be fairly simple. Just add this as a new endpoint in UserAdminApi.java

<code>
    @GetMapping("/verify-email/{token}")
    public void verifyEmail(@PathVariable("token") String token) {
        //todo Validate token and look up user
        //todo If validated then user.setEmailVerified(true)
        System.out.println(token);
    }
</code>

And permit the new endpoint by editing SecurityConfiguraton.java to have a line like:

                .requestMatchers("/api/admin/user/verify-email/**").permitAll()

server/src/main/java/org/tctalent/server/api/admin/UserAdminApi.java
hiba-machfej: I initially didn’t send them, but later used them in a function to check if the user sent an email and if the token expired. This triggers the modal to say the email was already sent. However, there could be a better frontend approach, I will check.

server/src/main/java/org/tctalent/server/service/db/UserService.java
hiba-machfej: I'll make sure to fix this and use standard JavaDoc comments, Thank you!

server/src/main/java/org/tctalent/server/service/db/impl/UserServiceImpl.java
hiba-machfej: I was actually using the same approach that was used for the generateResetPasswordToken like in [here](https://github.com/Talent-Catalog/talentcatalog/blob/d5b5c758b50144924a2e3acc01a52573cc8fc3a3/server/src/main/java/org/tctalent/server/service/db/impl/UserServiceImpl.java#L580). I will update the code.

ui/admin-portal/src/app/components/account/verify-email/verify-email.component.html
hiba-machfej: I didn't realize the approach was so non-standard! Apologies, I’ll update it!

ui/admin-portal/src/app/components/account/verify-email/verify-email.component.ts
hiba-machfej: I'll update it and check for other methods missing the private keyword.

hiba-machfej:
I think I followed the Reset Password code too closely, as you mentioned. Thanks for the valuable comments! I'll review and improve the code.

server/src/main/java/org/tctalent/server/api/admin/UserAdminApi.java
hiba-machfej: I removed the token and its issuedDate.

server/src/main/java/org/tctalent/server/exception/InvalidEmailVerificationTokenException.java
hiba-machfej: The `ServiceException` was causing an error, and since I was just trying to get things done quickly, I ended up using `RuntimeException` as a workaround. But actually, all I needed to do was write `super("invalid_email_token", "The email token is not valid");` instead of `super("Invalid email verification token");`

hiba-machfej: I fixed this and updated it to `ServiceException`.

server/src/main/java/org/tctalent/server/service/db/impl/UserServiceImpl.java
hiba-machfej: fixed

server/src/main/java/org/tctalent/server/request/user/emailverify/SendVerifyEmailRequest.java
hiba-machfej: fixed

server/src/main/java/org/tctalent/server/request/user/emailverify/SendVerifyEmailRequest.java
hiba-machfej: fixed

server/src/main/java/org/tctalent/server/service/db/impl/UserServiceImpl.java
hiba-machfej: In the approach I was using I had the emailVerified set to true  in `verifyEmail`. So, it was like this:
- `sendVerifyEmailRequest` >  `checkEmailVerificationToken` > `verifyEmail`. I was trying to use the same approach to the password:
- `generateResetPasswordToke`n > `checkResetToken` > `resetPassword`.

hiba-machfej: Now, in the new simpler approach it is updated.

server/src/main/java/org/tctalent/server/api/admin/UserAdminApi.java
```diff
@@ -129,24 +129,20 @@ public void resetPassword(@RequestBody ResetPasswordRequest request) throws
...
-    }
-
-    @PostMapping(value="verify-email-token")
-    public void verifyEmail(@RequestBody VerifyEmailRequest request) {
+    @GetMapping("/verify-email/{token}")
+    public void verifyEmail(@PathVariable("token") String token) throws ExpiredEmailTokenException, InvalidEmailVerificationTokenException {
+        VerifyEmailRequest request = new VerifyEmailRequest();
+        request.setToken(token);
         userService.verifyEmail(request);
+        System.out.println(token);
```
camerojo: No reason to print token.

server/src/main/java/org/tctalent/server/service/db/UserService.java
```diff
@@ -80,12 +79,13 @@ public interface UserService {
...
+    /**
+     * Generates an email verification token and sends a verification email.
+     */
     void sendVerifyEmailRequest(SendVerifyEmailRequest request);
-    // Checks the validity of the email verification token.
-    void checkEmailVerificationToken(CheckEmailVerificationTokenRequest request);
-    // Verifies the user's email using the verification token.
+    /**
+     * Checks the validity of the email verification token and verifies the user's email.
+     */
```
camerojo, line 89: This does not say what happens if the verification fails. It throws an Exception which should be documented.

server/src/main/java/org/tctalent/server/api/portal/UserPortalApi.java
```diff
@@ -93,24 +93,18 @@ public void resetPassword(@RequestBody ResetPasswordRequest request) throws Inva
         userService.resetPassword(request);
     }
 
-    @PostMapping(value="verify-email")
+    @PostMapping("/verify-email")
```
camerojo: Why do we need a separate api entry on the portal.

server/src/main/java/org/tctalent/server/service/db/email/EmailHelper.java
```diff
@@ -171,7 +164,9 @@ public void sendVerificationEmail(User user) throws VerifyEmailSendFailedExcepti
             final Context ctx = new Context();
             ctx.setVariable("partner", partner);
             ctx.setVariable("displayName", displayName);
-            String verificationUrl = (user.getRole() == Role.user ? portalUrl : adminUrl) + "/verify-email?token=" + token;
+            String verificationUrl = "https://tctalent.org/api/" +
+                    (user.getRole() == Role.user ? "portal" : "admin") +
+                    "/user/verify-email/" + token;
```
camerojo: What is this used for in the email? We don't need this. A single api call is all that is needed to verify an email - either from a candidate or an admin. They are both users.

camerojo (changes requested):

server/src/main/java/org/tctalent/server/request/user/emailverify/VerifyEmailRequest.java
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * Request object for verifying email.
+ */
+@Getter
+@Setter
+public class VerifyEmailRequest {
+    private String token;
+
+    public String getToken() {
```
camerojo, line 30: Not needed? Plus toString

server/src/main/resources/mail/verify-email.html
```diff
@@ -0,0 +1,44 @@
...
+    <p style="margin: 20px 0;">
+        <a th:href="${verificationUrl}"
+           style="background-color: #49B6BD; color: white; text-decoration: none; padding: 12px 25px; border-radius: 5px; font-size: 16px; display: inline-block;">
+            Verify Email Address
+        </a>
+    </p>
+    <p style="font-size: 14px; color: #666;">
+        If you did not request this verification, please ignore this email.
+    </p>
+    <p style="margin-top: 30px; font-size: 12px; color: #999;">
```
camerojo: Probably leave this out

camerojo (approved):
Great work Hiba!

PR #1892, opened 2025-02-12
Fix: ensure consistent join type

### LINKED ISSUE #1887: Bug: 'Education Level (minimum)' filter error in staging
This one's in staging but not production, discovered while working on something else.

Select any option in the 'Education Level (minimum)' dropdown and run search:

`org.springframework.dao.InvalidDataAccessApiUsageException: Requested join fetch with association [maxEducationLevel] with 'left outer' join type, but found existing join fetch with 'inner' join type.`

MuhannedNoman:


server/src/main/java/org/tctalent/server/repository/db/CandidateSpecification.java
samschlicht, line 1: As noted in my issue comments, this approach handles the bug by replicating the LEFT JOIN used in the predicate builder for the code at the top of the class which enables sorting by column. 

However, this value is never null, so we don't need a LEFT JOIN - my preferred approach is to remove it from the predicate builder.

PR #1906, opened 2025-02-17
Fix: sf populate next step fields

### LINKED ISSUE #1847: Default Next Step for newly created Candidate Opp not reflected in SF
- Adding a candidate to a submission list creates a Candidate Opp in the TC with a Salesforce equivalent
- Next Step and Next Step Due Date are automatically populated
- These values are currently not populated in Salesforce upon record creation

MuhannedNoman:
The `candidateOppParams` parameter is being passed as null from the `addCandidateToList` function. To address this, I added a check to initialize it if it’s null. This ensures that Salesforce creation happens before adding objects to the database.

There’s another default initialization that performs a similar check in the `createOrUpdateCandidateOpportunity` function. However, I’m unsure if this logic is ever triggered because of the following line:
```java
CandidateOpportunity opp = findOpp(candidate, jobOpp);
```
This line usually returns a value, so the if check for initialization is never reached. That said, it might still be triggered by other workflows.

I didn’t refactor the initialization logic into a separate function because:

1. `createOrUpdateCandidateOpportunity` requires a `CandidateOpportunity` object.

1. `createUpdateCandidateOpportunities` requires a `CandidateOpportunityParams` object.

The differing parameter types make it impractical to consolidate the initialization into a single function.

### DIFF
--- a/server/src/main/java/org/tctalent/server/service/db/impl/CandidateOpportunityServiceImpl.java
+++ b/server/src/main/java/org/tctalent/server/service/db/impl/CandidateOpportunityServiceImpl.java
@@ -205,9 +205,18 @@ public class CandidateOpportunityServiceImpl implements CandidateOpportunityServ
                 sfJobOpp = salesforceJobOppService.updateJob(sfJobOpp);
             }
 
+            CandidateOpportunityParams oppParams =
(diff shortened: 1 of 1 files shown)


samschlicht:
See notes in #1847

PR #1923, opened 2025-02-26
Fix: remove roles from seeing candidate name info

### LINKED ISSUE #1919: Bug: Employer user sees candidate names in quick candidate search
I don't think this is a showstopper - we're only beta testing employer access at the moment - so I've marked it for 2.3 _and_ 2.4. 

Employer access is set up such that employer admins can only see candidate names once a candidate is past CV Review stage. So there's no practical application for candidate quick-search, and it compromises anonymity.

Worth checking for other instances, but this is List view and the Header:

MuhannedNoman:
Since the admin and partner admin roles don't get candidate names from the API I also removed them from the front-end.


sadatmalik:
Closing this PR because the issue that it is connected to has already been resolved by another PR: https://github.com/Talent-Catalog/talentcatalog/pull/1974/files

PR #1927, opened 2025-03-01
Add support for generating classes from the OpenAPI tc-api-spec, assu…

### LINKED ISSUE #1926: Add talentcatalog gradle task to generate and compile from OpenApi spec and import some generated classes
Set up the OpenAPI tc-api-spec as a Intellij module in the talentcatalog project. Then add a gradle openApiGenerate task to generate Java files into the talentcatalog  build/generated directory in the same way as we do for tc-api.

This enables the standard core TC code to share components generated from the OpenAPI. In particular this is needed for OpenAPI end points whose implementation needs to call back through the core TC API.
For example:

- OTA
- EOI
- 3rd party registration

camerojo:
…med to be present as module in the Intellij project.

### DIFF
--- a/server/build.gradle
+++ b/server/build.gradle
@@ -5,19 +5,23 @@
  * the terms of the GNU Affero General Public License as published by the Free
  * Software Foundation, either version 3 of the License, or any later version.
  *
- * This program is distributed in the hope that it will be useful, but WITHOUT 
+ * This program is distributed in the hope that it will be useful, but WITHOUT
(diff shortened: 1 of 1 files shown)


sadatmalik (commented):
Don't need this anymore since the api generated code has been promoted as an artefact in Maven central. Pull the build.gradle from staging and it should work.

camerojo:
No longer needed - see above

Issue #843, opened 2024-04-05
Move from UA to GA4 tags in admin portal


sadatmalik:
We introduced GA4 tags in the candidate portal in release 2.2.1

We need to do likewise in the admin portal - most likely with a new GA4 tag.

As part of this issue - also investigate the options for maintaining the UA history (noting that UA will be sunset later this year) in a data hub.

-----
All standard Universal Analytics properties will stop processing new hits on July 1, 2023, and 360 Universal Analytics properties will stop processing new hits on July 1, 2024. After that, you’ll be able to access your previously processed data in Universal Analytics for at least six months

https://blog.google/products/marketingplatform/analytics/prepare-for-future-with-google-analytics-4/

PR #1957, opened 2025-03-12
843 add GA4 tag to Admin portal

hiba-machfej:
This PR adds the new GA4 tag for Google Analytics to Admin Portal. It is related to this issue #843, and I tried to follow the same steps as #767.


sadatmalik (approved):

sadatmalik:
Hi @hiba-machfej just looking at this again, the googleAnalyticsId will also need to be added to environment.staging.ts

The PR has it in environment.ts and environment.prod.ts; and we more recently also have the environment.staging.ts file so that'll need updating as well.

PR #1969, opened 2025-03-14
1925 Implement OTA and EOI TC server APIs which support public API to core TC server

### LINKED ISSUE #1925: Implement OTA and EOI TC server APIs which support public API to core TC server
Offer to Assist API (also used for EOI job API)

Normal entity update API - with just a create method. All OTA requests are stored on the DB as Auditable entities (create and update dates)

For now, maybe we just display OTA requests to admins. For now they will be processed manually.

- [x] Database Flyway
- [x] Need to store partner (service provider) associated with ota in db. Pass partner into TC after looking up partner from authentication on creation of OTA? Public access partners need to be stored on internal TC (not public db - as we currently have (in ApiUser)
- [x] Call from tc-api
- [x] need to map authorized user to partner id - see todo in tc-api CandidateController
- [x] #1967

camerojo:


server/src/main/java/org/tctalent/server/api/admin/OfferToAssistAdminApi.java
```diff
@@ -0,0 +1,65 @@
...
+        OfferToAssist ota = offerToAssistService.createOfferToAssist(request);
+
+        //Create response
+        OfferToAssistCandidates201Response response =
+            new OfferToAssistCandidates201Response().toBuilder()
+                .offerId(ota.getPublicId())
+                .message("Your offer has been successfully recorded.")
+                .build();
+
+        return new ResponseEntity<>(response, HttpStatus.CREATED);
```
sadatmalik, line 63: Recommendation generated by Amazon CodeGuru Reviewer. Leave feedback on this recommendation by replying to the comment or by reacting to the comment using emoji.

You have an outgoing server response that uses potentially unsanitized external inputs. This can lead to Cross-site Scripting  (XSS) attack allowing an attacker to bypass access controls and execute malicious code remotely on clients. 

Mitigation : To prevent XSS attacks, make sure to sanitize and escape responses used in models, views, or controllers. You can use encoders such as `StringEscapeUtils.escapeHtml` or `org.owasp.html.HtmlSanitizer` for HTML content, and `URLEncoder` for URL contexts. See `org.owasp.esapi.Encoder` for escaping in other contexts such as HTML attributes, CSS data, DOM and embedded scripts.[Learn more](https://cheatsheetseries.owasp.org/cheatsheets/Cross_Site_Scripting_Prevention_Cheat_Sheet.html)

sadatmalik (approved):

PR #1981, opened 2025-03-18
1967 display otas in admin settings

### LINKED ISSUE #1967: Display OTAs in Admin settings
Note that this issue branches off #1925

cazcam34:
Table is in settings with a keyword search that searches across the partner, created by username, public id & reason. I thought those might be the most commonly searched fields, but can adjust the filters as needed. 
Screenshot 2025-03-18 at 12 56 40 pm

And table is responsive so scrolls across:
Untitled2


ui/admin-portal/src/app/components/settings/offer-to-assist/offer-to-assist.component.html
```diff
@@ -0,0 +1,78 @@
...
+
+      <tbody>
+      <tr *ngFor="let ota of results?.content">
+        <td>{{ota.id}}</td>
+        <td>{{ota.partner.name}}</td>
+        <td>{{ota.publicId}}</td>
+        <td>{{getDisplayEnumValue(ota.reason)}}</td>
+        <td>{{ota.additionalNotes}}</td>
+        <td>{{ota.createdBy.username}}</td>
+        <td>{{ota.createdDate}}</td>
```
sadatmalik, line 50: Since these are created directly by the TC service in the database when received from the API, there won’t be a createdBy.username. They also can’t be updated, so updatedBy and updatedDate will be null. Can we hide these three columns for now and display only createdDate?

cazcam34: Thanks @sadatmalik , I have removed the createdBy, updatedBy and updatedDate fields and the updated PR is ready for review.
Just as a side note, the createdBy field is being set in the create method by the logged in user. But I still removed it as I imagine it's not necessary to display in the settings.
Screenshot 2025-03-21 at 10 18 52 am

sadatmalik: Yeah I agree I think we don't have to display it, though no harm I guess if we were to. It will always be the same, i.e. it'll be the user that the TC API service is logged in with.

sadatmalik (approved):

PR #2032, opened 2025-04-04
1975 bug users can add candidates to lists they have no connection to

### LINKED ISSUE #1975: Bug: users can add candidates to lists they have no connection to
Currently as a user I can select a candidate and add them to an existing list that I have no connection to (see e.g. below - not mine, not shared with me, not starred by me).

This is particularly hazardous since we have a 'Replace list contents' checkbox — although thankfully not when a submission list is selected. 

In any case, someone with no connection to a job shouldn't be able to add a candidate to a sub list. Nor should they be able to add a candidate to someone else's list. In both cases they would be unable to fix their error, since they don't have direct access to the list.

cazcam34:
Employer partners -> can only select from their lists, partner's lists and shared lists
Everyone else -> can select from their lists, global lists and shared lists
Read Only -> can only select their lists


camerojo (approved):
Nice work!

PR #2039, opened 2025-04-07
2016 candidate portal support for changing password

### LINKED ISSUE #2016: Ability to require candidate user to change password following login. Candidate portal support for changing password while logged in.
Candidate entity should have a changePassword flag - which defaults to false but if true will require the candidate to change their password when they next log in.

This is needed in 3rd party registration where a candidate is automatically registered and assigned a temporary password which will have to be changed when they first log in.

It may be useful to have as general functionality though.

Currently the only way that a candidate can change their password is by clicking Forgot Password when they are logged out (on the login screen). This requests them to enter their email and an email is sent to them with a link which takes them to 'reset-password/token'. The Candidate Portal Angular code routes this (in app-routing.module.ts) to ChangePasswordComponent which extract the token, and then prompts them for a password and sends the password and token to the server (the token is used to look up the user).

Anyway - the point to note is that we already have a ChangePasswordComponent which we could reuse to allow a user to request a password change while already logged in. There will be no need for a token to identify the user. 

Once a password has been changed, the candidate's changePassword flag can be set to false.

When we have this functionality we may as well add a Change password menu item so a candidate user can change their password any time.

hiba-machfej:
This PR adds support for the changePassword flag on the Candidate model and implements related updates on both the server and frontend sides.

Server-side:
- Adds a changePassword boolean flag to the Candidate entity.
- Updates the password reset flow to set changePassword to false after the candidate successfully resets their password.

Frontend:
- When a candidate logs in and the changePassword flag is true, a modal prompts them to change their password.
- Adds an "Edit Password" button on the profile edit page, allowing users to update their password manually.

I used the same component we already have for ChangePasswordComponent.

### DIFF
--- a/server/src/main/java/org/tctalent/server/service/db/impl/UserServiceImpl.java
+++ b/server/src/main/java/org/tctalent/server/service/db/impl/UserServiceImpl.java
@@ -617,11 +617,12 @@ public class UserServiceImpl implements UserService {
         userRepository.save(user);
 
         /* Update changePassword */
-        if (user.getCandidate().isChangePassword()) {
+        if (user.getCandidate() != null && user.getCandidate().isChangePassword()) {
             Candidate candidate = candidateRepository.findById(user.getCandidate().getId())
                 .orElseThrow(
                     () -> new NoSuchObjectException(Candidate.class, user.getCandidate().getId()));
             candidate.setChangePassword(false);
+            candidateRepository.save(candidate);
         }
     }
 
--- a/server/src/main/java/org/tctalent/server/service/db/impl/UserServiceImpl.java
+++ b/server/src/main/java/org/tctalent/server/service/db/impl/UserServiceImpl.java
@@ -63,6 +63,7 @@ import org.tctalent.server.exception.ServiceException;
 import org.tctalent.server.exception.UserDeactivatedException;
 import org.tctalent.server.exception.UsernameTakenException;
 import org.tctalent.server.logging.LogBuilder;
+import org.tctalent.server.model.db.Candidate;
 import org.tctalent.server.model.db.Country;
 import org.tctalent.server.model.db.PartnerImpl;
 import org.tctalent.server.model.db.Role;
@@ -71,6 +72,7 @@ import org.tctalent.server.model.db.User;
 import org.tctalent.server.model.db.partner.Partner;
 import org.tctalent.server.repository.db.CountryRepository;
 import org.tctalent.server.repository.db.UserRepository;
+import org.tctalent.server.repository.db.CandidateRepository;
 import org.tctalent.server.repository.db.UserSpecification;
 import org.tctalent.server.request.LoginRequest;
 import org.tctalent.server.request.user.CheckPasswordResetTokenRequest;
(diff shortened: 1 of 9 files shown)


server/src/main/resources/db/migration/V1_352__add_change_password_field_to_candidate.sql
```diff
@@ -0,0 +1,17 @@
...
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE. See the GNU Affero General Public License
+ * for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see https://www.gnu.org/licenses/.
+ */
+
```
camerojo: Flyway's should not have the copyright notice. The reason is that it is messy to change Flyway files once they have been processed - so having copyright notices which are typically updated each year doesn't make sense on Flyway files.

server/src/main/java/org/tctalent/server/service/db/impl/UserServiceImpl.java
```diff
@@ -610,6 +615,14 @@ public void updatePassword(UpdateUserPasswordRequest request) {
         String passwordEnc = passwordHelper.validateAndEncodePassword(request.getPassword());
         user.setPasswordEnc(passwordEnc);
         userRepository.save(user);
+
+        /* Update changePassword */
+        if (user.getCandidate().isChangePassword()) {
+            Candidate candidate = candidateRepository.findById(user.getCandidate().getId())
```
camerojo, line 621: This will throw a null pointer exception for users that aren't candidates.

if (user.getCandidate() != null &&

server/src/main/java/org/tctalent/server/service/db/impl/UserServiceImpl.java
```diff
@@ -610,6 +615,14 @@ public void updatePassword(UpdateUserPasswordRequest request) {
...
+                .orElseThrow(
+                    () -> new NoSuchObjectException(Candidate.class, user.getCandidate().getId()));
+            candidate.setChangePassword(false);
+        }
```
camerojo, line 626: You haven't saved the updated candidate.

I think you need

candidateRepository.save(candidate)

camerojo (changes requested):

server/src/main/java/org/tctalent/server/service/db/impl/UserServiceImpl.java
hiba-machfej: done

server/src/main/java/org/tctalent/server/service/db/impl/UserServiceImpl.java
hiba-machfej: done

server/src/main/resources/db/migration/V1_352__add_change_password_field_to_candidate.sql
hiba-machfej: done

hiba-machfej:
Thank you so much for the PR review @camerojo. I updated all the changes.

camerojo (approved):

PR #2076, opened 2025-04-15
Reinserted job update setNextStep

### LINKED ISSUE #2074: Bug: Job Opp Next Step not updating
Here's a vid of the problem - it's happening in prod, too, so this'll need to be a hot fix:

https://github.com/user-attachments/assets/5c98a684-dd49-4027-a733-20efe9f79a4a

Here's the PR where it was removed, while refactoring the stage change chat notification code:
https://github.com/Talent-Catalog/talentcatalog/pull/1876/files#diff-de7c0d38405b838ea4362312cbeed08c9c018404cd41cd89be17da2573502830L830:~:text=if-,(,-request.getNextStep

samschlicht:


sadatmalik (approved):

camerojo (approved):
Looks good to me @samschlicht - many thanks.

What were your concerns?

samschlicht:
> Looks good to me @samschlicht - many thanks.
> 
> What were your concerns?

Thanks, @camerojo! Nothing major: your previous refactoring removed JobService's dependency on NextStepProcessingService and instead handled that in the OppNotificationService — now we have the same dependency in both services, and they both perform the same string processing on the Next Step.

It's the same as the equivalent Candidate Opp code and not too great a sin. Perhaps just not quite best-practice — but as I said in the meeting, arguably best to implement the simplest solution for now and refactor later if we really feel the need to refactor.

PR #2105, opened 2025-04-17
Remove DET button from side profile view (same as other buttons)

cazcam34:


camerojo (commented):
I can see that that the DuoLingo assignment button no longer displays but it looks like Assign now does display - why is that?

camerojo (changes requested):
This doesn't look right. Did you cut instead of copy the *ngIf line?

cazcam34:
@camerojo I'm not sure what you mean? This is what I see:
No buttons in side profile
Screenshot 2025-04-22 at 10 52 51 am

Both buttons in full profile
Screenshot 2025-04-22 at 10 52 43 am

What kind of user were you?

ui/admin-portal/src/app/components/candidates/view/tasks/view-candidate-tasks.component.html
```diff
@@ -26,11 +26,11 @@
     <span [ngClass]="{'custom-flex-40' : editable, 'custom-flex-75' : !editable}">Candidate Tasks: Ongoing</span>
     <div class="d-flex flex-fill justify-content-between align-items-center">
       <span class="small font-italic">Comment?</span>
-      <div class="d-flex">
+      <div *ngIf="editable" class="d-flex">
```
cazcam34, line 29: I moved the *ngIf="editable" to the parent div which encapsulates both buttons as opposed to just the assign task button

camerojo (approved):

sadatmalik (approved):

PR #2109, opened 2025-04-17
Add “HowHeardAboutUs” Enum & Update Candidate Source Options in Talent Catalog

### LINKED ISSUE #2030: Add “HowHeardAboutUs” Enum & Update Candidate Source Options in Talent Catalog
The current list of candidate source options in the Talent Catalog is outdated. Additionally, there is no existing enum in the API or TC service for categorizing these options. A new enum, HowHeardAboutUs, needs to be introduced.

The screenshot is taken directly from our production database with the currently selectable options.

EhsanEhrari:


camerojo (dismissed):

server/src/main/resources/db/migration/V1_354__update_survey_type.sql
```diff
@@ -0,0 +1,28 @@
+update survey_type
+set name = 'Friend or colleague referral'
+where name = 'From a friend';
+
+-- Merge 'Facebook - through an organisation' into 'Facebook'
+update survey_type
+set name = 'Facebook'
+where name = 'Facebook - through an organisation';
+
```
sadatmalik: I think you'll also need to mark its status inactive or it will appear twice in the "How did you hear about us?" drop down list?

sadatmalik (changes requested):

server/src/main/resources/db/migration/V1_354__update_survey_type.sql
EhsanEhrari: Hi @sadatmalik 
How will it appear twice I couldn't understand?
This is an active value, and I only merge the values.

sadatmalik: I mean here in the candidate portal drop-down:

Screenshot 2025-04-23 at 09 54 48

sadatmalik: Thinking about this some more. . . I think we can

(i) mark "Facebook - through an organisation" as an inactive status, and 
(ii) update candidates.survey_type_ids where candidates have selected "Facebook - through an organisation" to the active id for "Facebook"

sadatmalik (dismissed):

sadatmalik (dismissed):

sadatmalik (approved):

hiba-machfej (approved):

PR #2124, opened 2025-04-22
Removed unnecessary LEFT JOIN for maxEducationLevel

samschlicht:


EhsanEhrari (approved):
I think because of this filter on educationLevel.get("level"), 
LEFT JOIN is unnecessary unless we want to keep rows with null education levels.

sadatmalik (approved):

camerojo (approved):

PR #2131, opened 2025-04-22
2116 assess current test coverage identify improvements backend frontend

### LINKED ISSUE #2116: Assess current test coverage & identify improvements (backend & frontend)
Conduct a review of our current test coverage across both the backend and frontend codebases. 

The goal is to identify areas needing improvement, with the aim of reaching 70–80% test coverage in all key areas.

Actions:
- Asses what our current coverage is (backend / frontend)
- Recommend where we need to improve - aiming for 70-80% coverage
- Break down the work into smaller tickets

We will then organise a (partial) sprint for the whole team to work on completing these tickets together.

EhsanEhrari:


server/build.gradle
```diff
@@ -186,10 +187,24 @@ jib {
     }
 }
 
+jacoco {
+    toolVersion = "0.8.12" // Updated to latest version
```
sadatmalik, line 191: The comment seems a bit odd since we've not had a jacoco version before that is being updated?

```suggestion
    toolVersion = "0.8.12" 
```

sadatmalik (approved):
Left a minor suggestions but approved.

hiba-machfej (approved):

server/build.gradle
hiba-machfej: yes, double on that. It might be misleading to have "Updated".

PR #2133, opened 2025-04-23
2122 remove any null characters from linkpreviews

samschlicht:


server/src/main/java/org/tctalent/server/util/html/StringSanitizer.java
```diff
@@ -56,4 +61,40 @@ public static String replaceLsepWithBr(String input) {
...
+   * which are preserved.
+   * <p>
+   * If control characters are found and removed, a debug log message is generated with the input
+   * string (truncated to 500 characters if it's longer).
+   *
+   * @param input the string to sanitize, can be null
+   * @return the sanitized string with control characters removed, or null if the input was null
+   */
+  @Nullable
+  static String replaceControlCharacters(@Nullable String input) {
```
sadatmalik: Can you rename this to removeControlCharacters(...) instead of replace, which is closer in meaning to its purpose

```suggestion
  static String removeControlCharacters(@Nullable String input) {
```

server/src/test/java/org/tctalent/server/util/html/StringSanitizerTest.java
```diff
@@ -0,0 +1,71 @@
...
+ */
+
+package org.tctalent.server.util.html;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNull;
+
+import org.junit.jupiter.api.Test;
+
+public class StringSanitizerTest {
```
sadatmalik, line 24: Nice!

sadatmalik (dismissed):
Left a minor comment but approved.

server/src/main/java/org/tctalent/server/util/html/StringSanitizer.java
samschlicht: Great point, @sadatmalik! Oversight on my part - I was originally replacing with a question mark, like John's similar code in the Opportunity class, but ultimately opted for this approach. Resubmitted with your suggestion.

sadatmalik (approved):

camerojo (approved):

sadatmalik:
@samschlicht we'll go ahead and merge this into the patch release; however following conversation with John, can a removeNullCharacter method be used here as it's only the null character causing the related issue. Just noting here so it's not forgotten and we can talk it through next week.

PR #2151, opened 2025-04-24
2129 Fixing answers not displaying properly in Candidate Portal profile

### LINKED ISSUE #2129: Ready for CHERRY-PICKERS -- Bug: answers not displaying properly in Candidate Portal profile view
Found in production.

To replicate:
1. Add a language to your profile on the Candidate Portal 
2. Observe the resulting view:

NB: if you refresh the view, the languages display. However, log in fresh and they're gone again.

Video of that:

[Private Zenhub Video](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBeHZBQ1E9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--b77158fe0a915eb4b95975a2e3d0f32de0885db5/Screen%20sharing%20-%202025-04-22%203_08_48%20PM.mp4)

hiba-machfej:


samschlicht (dismissed):
Nice work, @hiba-machfej! Happy to approve but it might be worth also getting a review from @EhsanEhrari, since he commented this out in the first place - perhaps it was necessary to address another issue?

EhsanEhrari:
Hi @hiba-machfej,

If you uncomment the dropdown data for the candidate profile (occupations, countries, education majors, languages, language levels, and survey types) will immediately load when the component initializes, rather than waiting for a language change event to trigger the loading.
I think this line might be commented out to prevent a race condition or duplication, especially if the language is changed immediately after component initialization (as it does in the current code)
`setTimeout(() => this.languageService.changeLanguage(lang), 1000);
`
When you change the language, it triggers this subscription:
```
this.subscription = this.translateService.onLangChange.subscribe((event: LangChangeEvent) => {
  this.loadDropDownData();
});
```
So if you uncomment the initial this.loadDropDownData() line:

The dropdown data will be loaded once when the component starts.

Then again when the language changes (after 1 second).

This can cause unnecessary duplicate HTTP calls, or even data flickering in the UI as it loads twice quickly.

hiba-machfej:
> Hi @hiba-machfej,
> 
> If you uncomment the dropdown data for the candidate profile (occupations, countries, education majors, languages, language levels, and survey types) will immediately load when the component initializes, rather than waiting for a language change event to trigger the loading. I think this line might be commented out to prevent a race condition or duplication, especially if the language is changed immediately after component initialization (as it does in the current code) `setTimeout(() => this.languageService.changeLanguage(lang), 1000); ` When you change the language, it triggers this subscription:
> 
> ```
> this.subscription = this.translateService.onLangChange.subscribe((event: LangChangeEvent) => {
>   this.loadDropDownData();
> });
> ```
> 
> So if you uncomment the initial this.loadDropDownData() line:
> 
> The dropdown data will be loaded once when the component starts.
> 
> Then again when the language changes (after 1 second).
> 
> This can cause unnecessary duplicate HTTP calls, or even data flickering in the UI as it loads twice quickly.

Ah, thank you @EhsanEhrari for pointing that out. The issue was that `this.loadDropDownData();` wasn't being called unless the page was refreshed, so the dropdown data wasn't showing initially if there was no language change or refresh.

How about using this approach instead?

```if (lang) {
  setTimeout(() => {
    this.languageService.changeLanguage(lang);
  }, 1000);
} else {
  this.loadDropDownData(); // only load immediately if no lang change
}
```
This way, if there's no language change, the data is showing. It avoids unnecessary duplicate HTTP calls, but still supports direct reload when a language change is needed. or do you think there might be a better solution?

EhsanEhrari:
@hiba-machfej It avoids unnecessary duplicate HTTP calls but still needs a direct reload when a language change is needed.
I think another option is to load immediately only when lang param is not set.

something like this:
```
  const lang = this.route.snapshot.queryParams['lang'];

  if (lang) {
    setTimeout(() => this.languageService.changeLanguage(lang), 1000);
  } else {
    this.loadDropDownData();
  }
```
What do u think?

hiba-machfej:
> @hiba-machfej It avoids unnecessary duplicate HTTP calls but still needs a direct reload when a language change is needed. I think another option is to load immediately only when lang param is not set.
> 
> something like this:
> 
> ```
>   const lang = this.route.snapshot.queryParams['lang'];
> 
>   if (lang) {
>     setTimeout(() => this.languageService.changeLanguage(lang), 1000);
>   } else {
>     this.loadDropDownData();
>   }
> ```
> 
> What do u think?

@EhsanEhrari Yess🎉! I just realized I had the first line of the if condition missing when I wrote the comment above 🤦‍♀️
That is exactly what I have in the code now but in my comment before I missed the `if (lang) {` and I only have the else part of the condition.
Thank you Ehsan 🙏

EhsanEhrari (approved):

ui/candidate-portal/src/app/components/profile/view/tab/profile/candidate-profile.component.ts
```diff
@@ -78,9 +78,14 @@ export class CandidateProfileComponent implements OnInit {
...
-    )
+    if (lang) {
+      setTimeout(() => {
+        this.languageService.changeLanguage(lang);
+      }, 1000);
+    } else {
+      this.loadDropDownData(); // Only load immediately if no lang change
+    }
+
     // this.loadDropDownData();
```
samschlicht, line 89: @hiba-machfej 
Nice work! We need one approval from Sadat or John and one from the rest of the team, so you'll be able to merge this as soon as you have the former. It looks good to me, my only suggestion would be not to leave unnecessary commented-out code (L89) that could confuse someone in future.

sadatmalik (approved):

PR #2156, opened 2025-04-29
2020 genericise processing package + 2130 email alert failing syncs

### LINKED ISSUE #2130: Email alert for failing scheduled background syncs?
As a TC dev
I want an alert for a failing scheduled background sync (e.g. TC->SF candidate sync, SF->TC job/case sync)
So that I don't have to manually (and perhaps unnecessarily check the AWS logs)

To tech@talentbeyondboundaries.org ? ie. everyone on the team will see it.

samschlicht:


server/src/main/java/org/tctalent/server/util/background/BackRunner.java
```diff
@@ -51,6 +52,7 @@
 public class BackRunner<CONTEXT> implements Runnable {
     private TaskScheduler taskScheduler;
     private BackProcessor<CONTEXT> backProcessor;
+    private BackLogger backLogger;
```
camerojo, line 55: Some nice code in here using interfaces and factories - but it does add a lot of extra code to the the code base.

I think that a more flexible and less coupled approach would be to use the Observer pattern allowing the user to plug in a simple Listener. It is just an interface like BackLogger except that it does not make any assumptions about what code is executed. That means that none of the factories or logging code needs to be in this Background processing code. Then we don't need a util/background/logging package and classes. All of that can be external to the background processing code. Simpler, more flexible and more loosely coupled.

This is the approach used by Spring's standard Batch Processing - see org.springframework.batch.core.ChunkListener

camerojo (changes requested):

PR #2163, opened 2025-04-29
Create read only directive to disable form inputs

### LINKED ISSUE #2162: Create read only directive to disable form inputs
Read only users need to be able to see form inputs (intakes/visa checks/job intake) but not able to change the values. Adding a security end point to restrict read only caused alot of log outs, and it isn't clear that edits aren't allowed.
Create a directive that can easily be reused to apply the appropriate disabled state and CSS to the different inputs (different inputs require different attributes)

cazcam34:
Create directive that can be added at a parent level to disable all child inputs (ng-select, textarea, input, app-date-picker) and apply CSS if needed to keep consistent. Different input require different disabled attributes so needed to be done differently for each. Applied on the mini and full intakes, but can be applied on other areas we want to disable the inputs in, e.g. JOI, context notes, visa checks.

Directive can be applied at top of intake accordion:
Screenshot 2025-04-29 at 2 40 57 pm

Example of different kinds of inputs, all being consistently disabled:
Screenshot 2025-04-29 at 2 40 43 pm


cazcam34:
Putting in draft to finish adding the tests

cazcam34:
Using !isEditable as opposed to isReadOnly allows the inputs to be disabled for read onlys AND also for users who aren't allowed to edit a candidate due to not assigned to their partner for example.

sadatmalik (approved):

EhsanEhrari (approved):

PR #2197, opened 2025-05-14
2171 candidates not reassigned when current partner not operational in newly provided country location

### LINKED ISSUE #2171: Candidates not reassigned when current partner not operational in newly provided country location
Channel post from Suci:
Hi <@U3TVD430A> and <@U02NYRN57V5>,
I have noticed that some candidates who registered using our URL are not in Indonesia. If you extract data based on the URL, you will see that some refugees who registered using the CRS or YCWS URL are currently in China, their transit country, or even in their country of origin. For instance, refugees from Afghanistan registered directly from Afghanistan and are still there, as well as those from Pakistan and several countries in Africa.

I am not entirely sure how this happened, but we believe that due to the numerous information sessions we conducted both online and offline, as well as the various pamphlets we distributed, it seems they also shared the information with their relatives and friends outside Indonesia. Considering the link we provided is easy to remember, short, and also uses a barcode, this might have contributed to the situation.

[Slack Message](https://refugeejobsmarket.slack.com/archives/C42S4AJ73/p1746155475445169?thread_ts=1744181833.703889&cid=C42S4AJ73)

_Search for YCWS candidates not in Indonesia confirms 104._

samschlicht:
This PR makes the following changes:
- Only new registrants are automatically reassigned.
- If the currently assigned partner isn't operational in the given location, then we reassign to:
  - The auto-assign partner if there is one, or
  - the default source partner.


server/src/main/java/org/tctalent/server/request/candidate/UpdateCandidatePersonalRequest.java
```diff
@@ -51,4 +51,6 @@ public class UpdateCandidatePersonalRequest {
     private String unhcrNumber;
     @Enumerated(EnumType.STRING)
     private YesNo unhcrConsent;
+    /* REGISTRATION? */
+    private Boolean isRegistration;
 }
```
camerojo: Not sure I understand why you need this isRegistration

camerojo (commented):
Let's walk through this together - see comment

server/src/main/java/org/tctalent/server/request/candidate/UpdateCandidatePersonalRequest.java
samschlicht: @camerojo yep sounds good! In advance of that, I can quickly summarise the rationale:

- We don't want existing candidates to be reassigned when they update their location, because they may already have a relationship with their managing partner, irrespective of where they're now located. (That was your point, made in the issue comments, which I agree with.)
- Update profile and new registration use exactly the same path on the server, and I couldn't see a better way to identify new registrants than to pass this property from the front-end, where the origin is clearly identifiable via the parent component.

samschlicht: Discussed and resubmitted accordingly!

cazcam34 (dismissed):

cazcam34 (approved):

EhsanEhrari (approved):

camerojo (approved):
Nice work!

Issue #2218, opened 2025-05-23
Review of roles


cazcam34:
As discussed on the 22nd May team meeting when bringing up issue #2192, we identified that there are two different factors coming into roles currently, permissions and visibility. Currently we incorporate both into a role, and so it's hard to seperate. Admins have full visibility of fields but also have permission to create new users (and do other actions on the portal). So if we want a user to have these permissions BUT not have full field visibility and see contact details it's going to require some additional logic. Perhaps we can find a way to seperate these two factors (permissions/visibilites) so it'll be easier to hide/show fields and end points to users.

In terms of visibility, currently we are returning the following fields for admin user roles.

**Limited - public properties**
            "id",
            "status",
            "muted",
            "candidateNumber",
            "publicId",
            "allNotifications",
            "gender",
            "dob",
            "yearOfArrival",
            "additionalInfo",
            "mediaWillingness",
            "candidateMessage",
            "folderlink",
            "sflink",
            "selected",
            "createdDate",
            "updatedDate",
            "contextNote",
            "user",
            "candidateOpportunities",
            "candidateReviewStatusItems",
            "country",
            "nationality",
            "registeredBy",
            "regoPartnerParam",
            "regoReferrerParam",
            "regoUtmCampaign",
            "regoUtmContent",
            "regoUtmMedium",
            "regoUtmSource",
            "regoUtmTerm",
             "candidateCitizenships",
            "candidateDestinations",
            "candidateEducations",
            "candidateExams",
            "candidateLanguages",
            "candidateOccupations",
            "candidateSkills",
            "candidateJobExperiences",
            "candidateCertifications"

**Semi Limited - added on top of public properties above**
            "city",
            "state",
            "address1"
   

**Admin roles - all fields as determined by the DTO.**

Questions: 
- Do we want semi-limited roles to reflect the api fields? This means semi limited would see the publically available fields which are anonymised?
- However we currently hide the intakes from semi limited & limited users, so alot of these fields would be unnecessary as they wouldn't appear displayed on the front end. Perhaps we do want to show the intakes then in this case?

cazcam34:
Questions:

- Do we want limited/semi limited users to see candidate chats? Currently they can see candidate chats if the candidate is assigned to them. My thought is that candidate chats could contain confidential/personal information that we hide from these users (e.g. they can't see contact and personal information or attachments but things could be shared via chats)
[...]

PR #2219, opened 2025-05-23
2192 display key fields when logged in as a unchr partner user

### LINKED ISSUE #2192: Display key fields when logged in as a UNCHR partner user
From Zeynep:

Languages, experience, education, and other sections are currently hidden from all profiles for anonymized access users (i.e. UNHCR partner). Would be great to have them visible again.

cazcam34:


cazcam34:
@camerojo I've just pushed the logic I'm thinking for UNHCR and other partners as discussed last night. You'll see I've added another case for non source and non destination partners (UNHCR) which I've named Viewer Partner (we can change the name). I've kept it in draft as I just wanted to get your thoughts and we can discuss more here.

Basically if you look at the code before my changes you'll see that UNHCR wasn't seeing fields (only seeing public fields) as they weren't system admin, weren't default partner admins, and didn't have job opps assigned to them with candidates past the CV stage (see method isInFullyVisibleCandidates). Therefore they fell through the first if statement... then because the partner (UNHCR) did not match the candidate's partner, the field ignore was set to true. So all non public fields were ignored. Can see more about this in the zenhub issue #2192 

**What I've done here now based on our discussion is the following logic:**

- System admins/TBB admins can see all fields as requested (e.g. Extended DTO)
- Non source partners and non job creators (UNHCR) "Viewer Partners" admins should only see semi limited properties for all candidates - the semi limited properties have identifiable information (contact details/names etc) removed. I have also  now added the missing fields (candidate education/languages etc) to semi limited so this will fix the initial bug as raised in the issue.
- Source partner admins can see all fields as requested IF they are assigned to the same partner as the candidate, if source partner admins are NOT assigned to a candidate they can only see the public fields.
- Source partner semi limited users can see the additional semi limited properties on top of the public properties IF they are assigned to the same partner as the candidate.
- All other partners (destination partners) will see fields depending on their role.

What this PR does is fix the UNHCR bug, by displaying the information that they should see (without the personal details). In looking at this area of code I also added the necessary logic to make sure source partners only see what they should see and hopefully refactored it to make it easier to read.

server/src/main/java/org/tctalent/server/api/admin/PartnerAndRoleBasedDtoPropertyFilter.java
```diff
@@ -77,58 +77,98 @@ public boolean ignoreProperty(Object o, String property) {
...
+                ignore = true;
+                break;
+            case semilimited:
+                // Semi limited roles can see some additional properties to the public properties if they exist
+                ignore = semiLimitedExtraProperties == null
+                        || !semiLimitedExtraProperties.contains(property);
+                // However, source partner semi limited users can't see the extra semi limited properties if they aren't
+                // assigned to the same partner as the candidate.
+                if (partner.isSourcePartner() && isNotPartnerMatch(partner, candidatePartner)) {
+                    ignore = true;
```
camerojo, line 131: I don't think that this is right. SemiLimited source partners can see anonymous skills, experience and qualifications data. In addition they can just see some extra, semi private stuff - like any other semi limited role user - namely address details.

camerojo (changes requested):
I think that this is a move in the right direction but still not quite right - see my comment.

I am trying to understand when this all got broken. I think that will help - because it did all basically work. I will post some thoughts in the issue.

server/src/main/java/org/tctalent/server/api/admin/PartnerAndRoleBasedDtoPropertyFilter.java
cazcam34: @camerojo as we discussed in the meeting - this part of the switch statement for semi limited is to ignore fields only if the candidate isn't assigned to the user's partner. Same as the source partner admin code, where users can see allowed fields ONLY if the candidate is assigned to them. 
Above this if statement the ignore is set depending on if the field is included in the semi limited properties array of allowed fields. Then this is overwritten to be ignored if they aren't assigned the candidate.

cazcam34:
Just following up from our team meeting - this code logic is in the ignoreProperty part of the filter. It takes the DTO type (Extended, Preview, Minimal etc...) that is passed into the selectBuilder(DtoType) method. Then depending on the desired Dto type, we filter the properties depending on the users partner and role.

- Admins don't ignore any properties for all candidates
- Source partner admins don't ignore any properties for candidate's they are assigned to BUT if they aren't assigned the candidate they only see public properties and ignore all other properties.
- Semi limited users see the property if it's included in the semi limited property array for candidates they are assigned to BUT if they aren't assigned the candidate they only see public properties and ignore all other properties.
- Limited users only see public properties for all candidates so always ignore all other properties.
- ALSO  Viewer Partners - not source partners but not job creators/destination partners (such as UNHCR) see the equivalent of a semi limited user for ALL candidates regardless of their role. This gives them access to everything they need to see but restricts some fields that may be personal such as contact details.

cazcam34:
@camerojo So why did it used to work?!
As you mentioned it stopped working when working on speed improvements for the candidate search card (and in turn the candidate profile). 
Previously we would fetch the candidate data (languages/educations.. aka the fields that all of a sudden stopped appearing) by fetching them individually depending on which tab was displayed. So we wouldn't load all the candidate data in one go, but would call the language api for example and fetch all the languages for that candidate when the languages component was initialised. This was leading to multiple API calls when trying to view a candidate profile, and would load more API calls when the tab was changed and more components initialised. Because of this, any PartnerAndRoleBasedDtoPropertyFilter logic was overwritten as we additionally fetched data uniquely via their own apis (e.g. language admin api unfiltered) and we were just returning the data for each API call and using the front end logic to hide or display the data depending on the user.

In our December release when we introduced the speed improvements, we introduced the bug causing the above. To improve the speed I instead fetched all the required candidate profile data in one API call. Either GET /number/{{number}} from the Candidate Admin API when loading the full candidate profile view or POST /search-paged for saved list or saved search when loading a list/search view of candidate data to display in a search card. 
Then when tabs are changed and components initilised we are just populating the component with the already fetched data, and not making additional API calls as previously done. This improved the speed.
However in those API calls to fetch the ALL candidate data we used the CandidateBuilderSelector, using an Extended DtoType and applied the PartnerAndRoleBasedDtoPropertyFilter... however this filter wasn't working properly for source partner admins and is where the bug appeared. 

In the filter we were filtering out properties (only showing the public properties) if the user's partner isn't assigned the candidate (partner != candidatePartner). Except system admins & default partners with admin/partner admin roles.
OR if they are included in the isFullyVisibleCandidates list (these are assigned to job creator partners, includes candidates who are past CV review stage and in jobs managed by the job creator partner).
The problem was that UNHCR isn't a partner type that is assigned any candidates, so they were only seeing public properties for all candidates. The public properties doesn't include properties like languages/education etc that Zeynep flagged as missing.

Screenshot 2025-06-06 at 11 22 02 am

cazcam34:
So what this PR does is fix up the logic by having a isViewerPartner method which will cover partners such as UNHCR, and allow those partners to see the semi limited properties for all candidates. This is giving them a sort of 'medium' level visibility.
This list of semi limited properties is useful as it provides an additional layer of filtering between the DtoType level of properties, and the most restricted public properties. So it's an inbetween level which allows some restrictions of personal data (e.g. contact details) that might be allowed from the DTO type (e.g Extended), but still shows the additional fields such as languages/education/skills etc. which we may not want to show as public. 
Public properties is the most restricted level of fields for candidates, it is what users can see of candidates that they aren't assigned to. So I don't think we want to add these missing fields (languages/education/experience) to the public properties otherwise we are giving this level of access to all users for all candidates regardless of if they are assigned them or not and regardless of their role. Also they may have identifiable information in their job experience field for example.

I think perhaps in the future when we review roles, and perhaps break it up into 'permissions' and 'visibility', we might be able to provide better naming for these methods and lists of properties. Perhaps instead of 'semi limited properties' it could be 'medium level visibility properties' or something like that. But for now, essentially it is a list of properties that suits a 'medium level visibility' (semi limited is between an admin user and a limited user, so is a 'medium' level role). Hopefully that makes sense...

server/src/main/java/org/tctalent/server/api/admin/CandidateBuilderSelector.java
```diff
@@ -87,7 +87,17 @@ public class CandidateBuilderSelector {
...
+            "candidateCertifications",
+            "candidateCitizenships",
+            "candidateDestinations",
+            "candidateEducations",
+            "candidateExams",
+            "candidateLanguages",
+            "candidateOccupations",
+            "candidateSkills",
+            "candidateJobExperiences",
+            "candidateCertifications"
```
camerojo, line 92: This mistakes the meaning of the "candidateSemiLimitedExtraProperties". The "extra" in the name means that the properties named here are the **_extra_** properties that a semi limited partner can see - _**in addition to**_ the standard public properties ("candidatePublicProperties") that anyone can see. 

The other changes that you propose all make sense - I see your logic - but they cause other problems because you are using this field in a different way to the way it was designed to be used. That means that your changes don't work in other cases.

For example, set any partner to "limited" type and you will see that partners with a "limited" role will report to the user on the front end that a candidate didn't enter education or experience data - when in fact they did.

The solution is to leave the "candidateSemiLimitedExtraProperties" the way they were, but add the extra candidateEducations, candidateExperience etc to the "candidatePublicProperties".

camerojo (changes requested):
Hi Caroline - I finally got a chance to look at this again. I think that you were misinterpreting the meaning of "candidateSemiLimitedExtraProperties". See my comment there. That misinterpretation introduces another bug around what users with a "limited" role see. 

But if you move the new public fields where I suggest then I think we are close to fixing this.

Bottom line - EVERYONE can see the public properties - even the lowest level user role: "limited".
If a partner is associated with a candidate (or is the default source partner) then

- Semi limited role users can see a bit more (detailed address info)
- Admin role users can see everything 


The bug we have now is that limited partners **_can't_** see normal public properties (like education, experience etc)

server/src/main/java/org/tctalent/server/api/admin/CandidateBuilderSelector.java
cazcam34: Thanks for the comments @camerojo  - no I do understand that is the **extra** properties for a semi limited user, on top of the public properties. I mustn't of explained myself property. So on the same page with that.

I guess my question was do we want limited users to be able to see these additional (missing) properties - education, experience, languages. In my comment above I mention why I didn't think we should add them to public properties (it could contain identifiable information in experience for example). But after more thought I guess we have these fields in the public API so they should also be in the public properties. So will go ahead and do that - I'll also mention it in the meeting tonight

cazcam34:
> Hi Caroline - I finally got a chance to look at this again. I think that you were misinterpreting the meaning of "candidateSemiLimitedExtraProperties". See my comment there. That misinterpretation introduces another bug around what users with a "limited" role see.
> 
> But if you move the new public fields where I suggest then I think we are close to fixing this.
> 
> Bottom line - EVERYONE can see the public properties - even the lowest level user role: "limited". If a partner is associated with a candidate (or is the default source partner) then
> 
> * Semi limited role users can see a bit more (detailed address info)
> * Admin role users can see everything
> 
> The bug we have now is that limited partners **_can't_** see normal public properties (like education, experience etc)

Yes on the same page - my assumption was that we didn't want limited users to see these properties, it's extra information that may be unnecessary for a limited user and as I mentioned in my comment above due to perhaps having identifiable information in the experience for example. But since we have them in the public API, I can understand that they should be in the public properties here.

cazcam34:
@camerojo  Have added the fields to the public properties as requested, and I have taken the PR out of draft.
I have added some other comments to the issue 'Review of Roles' #2218 - one that is particularly relevant to this PR is perhaps removing 'candidateOpportunities' from the public properties. We currently hide the Jobs tab (candidate opportunites) on the front end for users that aren't admins. So semi/limited users are getting this data back from the server but aren't displaying it. Displaying candidate opportunities to limited users opens them up to knowing the candidate's first name due to the naming structure of the candidate opportunity.

camerojo (approved):

samschlicht (approved):
Nice work, @cazcam34! Intent is really clear here and solves our issue. Not a block to approval but we should try to stop referencing TBB in comments since their role may change.

PR #2221, opened 2025-05-27
2210 add missing tests for read only code

### LINKED ISSUE #2210: Add missing tests for read only code
A couple tests were missing from PR #2205 as suggested by Ehsan. Merged the existing PR as it had been approved and wanted to get the merge in, but will add these tests as a priority.

cazcam34:


cazcam34:
Hi @EhsanEhrari - I created a draft PR so you could maybe see where I am going wrong with writing the test for the read only directive on the visa check. I tried the overrideComponent method as suggested but I am not sure how to test when the directive isn't directly on the child but it's on the parent component, but it filters down through the child components. I am hitting a wall with this, so hopefully you can help. 
Alternatively as mentioned on Slack there are unit test on the directive component itself so perhaps that would be sufficient to test the functionality?

EhsanEhrari (approved):

camerojo (approved):

PR #2282, opened 2025-06-19
2259 Added main design system files

hiba-machfej:


ui/admin-portal/src/styles.scss
hiba-machfej, line 1: Looking at the google docs @cazcam34 I think to keep this clean you suggested to have this format right
```
src/
├── scss/
│   ├── theme/ (Put any theme specific variables here)
│   │   ├── _bootstrap-variables.scss
│   │   ├── _tc_theme.scss
│   ├── components/ (Unique existing component CSS  here)
│   │   ├── _table.scss
│   │   ├── _btn.scss
│   │   ├── _modal.scss
│   ├── custom.scss (Import all the above into this file)
├── styles.scss (Import the custom scss into this file - to keep the styles.scss really clean)

```

cazcam34: It was just a suggestion, maybe eventually if we could restructure like that?  But for now I think it's fine to keep it in styles scss, as it will keep working with our existing structure.

cazcam34 (approved):
Nice! I think these are named well, and we can always rename them once the team starts implementing the classes. Intellij has a good rename ability where it should be fairly easy to change the name if needed.

sadatmalik (approved):

PR #2283, opened 2025-06-19
2274 create button component

hiba-machfej:
The button component has two configurations which are size and type.

The selector is 
```html
<tc-button> Button </tc-button>
```
 and the default state is like this:
Screenshot 2025-06-19 at 3 09 31 PM

To change the style you can specify the size and the type like this:
```html
<tc-button size="sm" type="outline">Button</tc-button>
```

These are the available types and sizes configurations:

| Property  | Options                                                                                      |
|-----------|-----------------------------------------------------------------------------------------------|
| `size`    | `'xs'`, `'sm'`, `'default'`, `'lg'`, `'xl'`                                                   |
| `type` | `'primary'` (default), `'secondary'`, `'outline'`, `'plain'`                   |

The following image shows the different variants of the button:
Screenshot 2025-06-19 at 2 47 06 PM

To update any button we already have you can just replace the tag and remove the classes for example:
```html
<!-- Before -->
<button class="btn btn-sm btn-accent-3" (click)="refresh($event)">
  Refresh
  <i class="fas fa-sync" title="Refresh data"></i>
</button>

<!-- After -->
<tc-button (click)="refresh($event)">
  Refresh
  <i class="fas fa-sync" title="Refresh data"></i>
</tc-button>
        
```

I plan to rate the complexity and speed of creating and using components on a scale from 1 to 10, where 1 is very easy and 10 is very hard.
For the Button component, I’d rate it 2 for creation and 1 for reuse. it was fairly easy and straightforward overall.

Please feel free to test the button out and give me feedback.


ui/admin-portal/src/app/shared/components/button/button.component.ts
```diff
@@ -0,0 +1,26 @@
+import { Component, Input } from '@angular/core';
+
+@Component({
+  selector: 'app-button',
+  templateUrl: './button.component.html',
+  styleUrls: ['./button.component.scss'],
+})
+export class ButtonComponent {
+  @Input() size: 'xs' | 'sm' | 'default' | 'lg' | 'xl'  = 'default';
+  @Input() btnType: 'primary' | 'secondary' | 'outline' | 'disabled' | 'plain' = 'primary';
```
cazcam34: We often disable a button depending on a variable (e.g. form.isInvalid ) so we need to be able to set the btnType as disabled IF a condition is met or changed. Will this work here when using a [btnType] input? I know bootstrap buttons allowed us to pass in a [disabled] input which could be set depending on a condition. So could we use [btnType] to do the same? Or for easier use could it be worth having an @Input() disabled here?

ui/admin-portal/src/app/shared/components/button/button.component.spec.ts
cazcam34, line 1: We probably are going to need tests for these new components - the test should be fairly simple as they are very simple components but I wonder if we should be writing tests as we go?

ui/admin-portal/src/app/shared/components/button/button.component.scss
cazcam34, line 1: I'm surprised these .btn classes were able to be overwritten in this file here and not in the global scss file (_btn.scss), as bootstrap have these btn classes already (.btn, .btn-sm, .btn-primary etc). Since we are overwriting the bootstrap classes should we have these btn classes in the _btn.scss file? 
However if the classes seem to work here in the button components folder, then maybe its better here? Hmm, unsure curious what your thoughts are.
It's tricky as we have the new components classes and the overwriting bootstrap component classes, so it's tricky to know where to put things. That's why I'm surprised the classes were able to be overwritten here!

cazcam34 (dismissed):

ui/admin-portal/src/app/shared/components/button/button.component.scss
hiba-machfej: @cazcam34 Yes, CSS classes can be really confusing, especially when they come from an external library like Bootstrap, since CSS relies on the principle of specificity. I think even outside the global SCSS file, if a CSS rule is more specific than the Bootstrap one, it will take effect. However, we don't always know the exact CSS for each Bootstrap component, so it often comes down to testing what works.

I'm okay with both options, but I lean towards keeping the CSS in the same folder and only placing overrides in the global file when necessary. That way, if a CSS class is used elsewhere, keeping it local  in the button's file for example avoids unintentionally affecting global styles.

ui/admin-portal/src/app/shared/components/button/button.component.ts
hiba-machfej: Great point catching this Caroline. I will update this.

hiba-machfej:
Replaced the 'btnType' with 'type' I thought it is more intuitive. Updated the button to support the native [disabled] attribute and removed the '.btn-disabled' class.

ui/admin-portal/src/app/shared/components/button/button.component.scss
```diff
@@ -11,6 +11,15 @@
   transition: background-color 0.2s ease-in-out, box-shadow 0.2s;
   box-shadow: 0 1px 2px rgba(0, 0, 0, 0.05);
   gap: 0.5rem;
+
+  &[disabled] {
```
cazcam34, line 15: Ah this is really cool I didn't know you could write a class like this with an attribute!! Love it!

cazcam34 (dismissed):

caroline-ritchie (dismissed):

cazcam34 (dismissed):

EhsanEhrari (commented):
Looks great!
Just a couple of suggestions:

1. **Accessibility**:

Wouldn't it be better to support ariaLabel as an @Input() for cases where the button content is icon-only or not descriptive? This would improve screen reader support and make the component more accessible.

2. **Test Coverage:**

Consider adding a couple more unit tests to cover:

- Class application based on size and type inputs
- Disabled state functionality
- (If added) rendering of aria-label when passed

hiba-machfej:
> Looks great! Just a couple of suggestions:
> 
> 1. **Accessibility**:
> 
> Wouldn't it be better to support ariaLabel as an @input() for cases where the button content is icon-only or not descriptive? This would improve screen reader support and make the component more accessible.
> 
> 2. **Test Coverage:**
> 
> Consider adding a couple more unit tests to cover:
> 
> * Class application based on size and type inputs
> * Disabled state functionality
> * (If added) rendering of aria-label when passed

@EhsanEhrari I added the updates

cazcam34 (approved):

sadatmalik (approved):
Great!

PR #2284, opened 2025-06-19
2275 style form components

hiba-machfej:
As we continue building and styling our components, an approach we can use is to create reusable SCSS utility classes. This method allows us to style more complex third-party components like Bootstrap components without rebuilding them from scratch.

For example, ng-select comes with a lot of built-in features that would take time and effort to recreate. While it's possible to recreate it, we should consider whether the time investment is needed, especially when the component already works well.
```html
      <ng-select
        id="stage"
        [items]="stages"
        [multiple]="true"
        [closeOnSelect]="true"
        [searchable]="false"
        bindLabel="stringValue"
        bindValue="key"
        placeholder="Select Stage"
        formControlName="selectedStages">
       </ng-select>
```
Instead of fully replacing such components, we can use global SCSS utility classes, similar to how Caroline explored customizing Bootstrap. These reusable classes can be added selectively to components or their wrappers, ensuring that:

- We're only targeting and modifying the components we intend to.
- We're not introducing unintended side effects elsewhere in the app.
- We can apply consistent design across reused third-party components.

This approach gives us both control and flexibility: we can maintain consistent styling while saving time by reusing well-tested components that already exist in our app.

For example:
We put the classes in the `_form.scss`.
```
src/
├── scss/
│   ├── components/ (Unique existing component CSS  here)
│   │   ├── _form.scss
├── styles.scss 
```
We want to style the existing input on the left to the right. 
Screenshot 2025-06-19 at 11 03 53 PM
The code we have:
```html
      <div class="col-10">
        <input #searchFilter type="text" class="form-control" placeholder="Search..."
               aria-label="Search" formControlName="keyword"
               id="keyword">
      </div>
```
We updated it like this: 
```html
  <div class="col-10">
        <label class="input-label">Job Search</label> // <--- added input-label class
        <p class="input-description">Search the list of available Jobs for candidates. </p> // <--- added input-description class
        <input #searchFilter type="text" class="input-form form-control" placeholder="Search..." //<--- added input-form class
               aria-label="Search" formControlName="keyword"
               id="keyword">
      </div>
```

In a similar way the `ng-select` can be styled by adding the class:
```html
    <div class="mb-3 ng-select-container-styled"> // <--- added ng-select-container-styled to the parent component of ng-select
      <label class="input-label">Stage</label>
      <p class="input-description">  Select the current step in the partner engagement process. </p>

      <ng-select
        id="stage"
        [items]="stages"
        [multiple]="true"
        [closeOnSelect]="true"
        [searchable]="false"
        bindLabel="stringValue"
        bindValue="key"
        placeholder="Select Stage"
        formControlName="selectedStages">
      </ng-select>
    </div>
```
Screenshot 2025-06-19 at 11 15 04 PM

The final component looks like this 
Screenshot 2025-06-19 at 10 38 06 PM

### DIFF
--- a/ui/admin-portal/src/scss/components/_form.scss
+++ b/ui/admin-portal/src/scss/components/_form.scss
@@ -1,0 +1,97 @@
+@import '../support-colors';
+@import '../color-mixins';
+@import "../typography-mixins";
+
+.input-label {
+  @include label-md;
+  margin: 0.4rem 0;
+}
+
+.input-description {
+  @include text-color-caption;
+  @include label-sm;
+  color: color(gray, 500); // text-gray-500
+  margin: 0.4rem 0;
+}
+
+.input-form {
+  @include paragraph-md;
+  @include text-color-body;
+  border: 1px solid;
+  @include border-color-base;
+  border-radius: 0.4rem;
+  box-shadow: 0 1px 2px rgba(0, 0, 0, 0.03);
+  padding: 6px 12px;
+
+  &::placeholder {
+    @include text-color-caption;
+  }
+
+  &:hover {
+    border-color: color(gray, 300);
+  }
+
+  &:focus {
+    outline: none;
+    @include border-color-blue;
+    box-shadow: 0 0 0 2px color(primary, 500);
+  }
+
+  &:focus-visible {
+    outline: none;
+    @include border-color-blue;
+    box-shadow: 0 0 0 2px color(primary, 500);
+  }
+
+  &:disabled {
+    cursor: not-allowed;
+    @include background-color-soft;
+    @include text-color-caption;
+    @include border-color-light;
+  }
+}
+
+.ng-select-container-styled {
+  font-family: $font-lato;
+  .ng-select-opened.ng-select-top > .ng-select-container {
+    border-radius: 0.4rem 0.4rem 0 0;
+    border: 1px solid color(gray, 300);
+  }
+  .ng-select-container {
+    border: 1px solid color(gray, 300);
+    border-radius: 0.4rem;
+
+    .ng-value-container {
+      @include text-color-caption;
+    }
+  }
+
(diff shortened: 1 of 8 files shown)


hiba-machfej:
Looking at the naming of the issue I think Caroline maybe you also had similar thoughts? since all the other issues were create and this one was style?

cazcam34:
> Looking at the naming of the issue I think Caroline maybe you also had similar thoughts? since all the other issues were create and this one was style?

Yes exactly - I figured the form components (and ng-select) were already created so we would just be styling them as opposed to creating a new component. I think creating a new component would be unneccessary as it really would just be a parent component that contains the form/ng select components themselves anyways.

ui/admin-portal/src/scss/components/_form.scss
```diff
@@ -0,0 +1,97 @@
...
+
+  &:disabled {
+    cursor: not-allowed;
+    @include background-color-soft;
+    @include text-color-caption;
+    @include border-color-light;
+  }
+}
+
+.ng-select-container-styled {
```
cazcam34, line 54: nice work styling the ng-select component - it can be hard styling an external component and I've had troubles before. So this is great!!

cazcam34 (approved):

cazcam34:
@hiba-machfej do you think adding a wrapper class to these form inputs would be better? I know you have done it using classes as I was leaning that way but do you have a preference to using own classes?

hiba-machfej:
> @hiba-machfej do you think adding a wrapper class to these form inputs would be better? I know you have done it using classes as I was leaning that way but do you have a preference to using own classes?

I don’t mind either way, I’ve been making decisions based on what I thought would work best given the component structure and to keep the CSS code compact.

hiba-machfej:
I'm closing this draft PR, since we are going with the component approach.

PR #2331, opened 2025-06-30
2195 old completed candidate opps being mistakenly set to notfitforrole

### LINKED ISSUE #2195: Old completed Candidate opps being mistakenly set to NotFitForRole
See https://refugeejobsmarket.slack.com/archives/CAZ9NG4UF/p1747076222336749

@samschlicht Can we chat about this? See my suggested explanation for this on the above Slack link.

I suspect that the NotFitRole candidate opp changes may have been triggered by auto logic when associated jobs were set to HiringCompleted.

camerojo:
This puts more informative information into the candidate notes of when cases are auto closed - including the original case stages before they were closed.


samschlicht (approved):
Very nice! This will really help with any confusing closures.

sadatmalik (approved):

PR #2359, opened 2025-07-09
publishing lists that exceed the google sheet api size limits silently fails

sadatmalik:
This PR -

- checks that the number of candidates to be published does not exceed the sheet data range
- throws an exception for the user if it does
- adds a little logging for candidate folder creation (this can be time consuming and without logging it may lead to the wrong assumption that the publish has failed)


EhsanEhrari (approved):

camerojo (approved):

hiba-machfej (approved):

sadatmalik:
Merged into staging. Is this worth hot fixing?

PR #2360, opened 2025-07-10
2139 enhance test coverage for orgtctalentserverapiadmin package

### LINKED ISSUE #2139: Enhance Test Coverage for org.tctalent.server.api.admin Package
The JaCoCo report indicates 49% instruction coverage (7,609 instructions missed) for the `org.tctalent.server.api.admin` package, critical for admin API endpoints.
Goal: Increase instruction coverage to 80%.
Scope:
Write unit tests for admin API endpoints, focusing on untested methods and conditional logic.
Improve branch coverage for decision points in API logic.

EhsanEhrari:


camerojo (dismissed):

sadatmalik:
Hi @EhsanEhrari instead of changing the access modifier from private to package-private for this number of methods, would you have a look at these alternatives, courtesy of chatGPT, both are abstractions of Java reflection:

- Spring's ReflectionTestUtils

```
import org.junit.jupiter.api.Test;
import org.springframework.test.util.ReflectionTestUtils;
import static org.assertj.core.api.Assertions.*;

class MyServiceTest {
  @Test
  void privateCalc() {
    MyService svc = new MyService();
    // invoke private method “calculate”
    Object result = ReflectionTestUtils.invokeMethod(svc, "calculate",  42);
    assertThat(result).isEqualTo(84);
  }
}
```

- PowerMock / Mockito Whtebox

```
import org.junit.Test;
import org.powermock.reflect.Whitebox;
import static org.junit.Assert.*;

public class MyServicePowerMockTest {
  @Test
  public void testPrivateMultiply() throws Exception {
    MyService svc = new MyService();
    int result = Whitebox.invokeMethod(svc, "multiply", 3, 5);
    assertEquals(15, result);
  }
}
```

EhsanEhrari:
According to ChatGPT:

✅ Recommended: Spring’s ReflectionTestUtils
✔️ Why it’s better:
✅ Lightweight and easy to use with Spring Boot (no extra dependencies if you're already using Spring).

✅ Doesn’t require modifying test runners or classloading (unlike PowerMock).

✅ Actively maintained and works well with JUnit 5 and modern Spring applications.

✅ Compatible with Mockito, which is the preferred mocking library today.

❌ Limitations:
Cannot mock static or final methods/classes (but that’s a design smell anyway — try to avoid private static logic).

**⚠️ PowerMock / Whitebox (Deprecated & discouraged)**
❌ Why you should avoid it:
❌ PowerMock is mostly deprecated and not compatible with modern versions of JUnit (especially JUnit 5).

❌ Uses heavy bytecode manipulation and a custom classloader — can cause issues with Spring Boot, Mockito, and IDE test runners.

❌ Breaks test isolation and makes tests harder to maintain.

❌ Difficult to integrate with CI/CD pipelines (especially in newer Java versions).

✅ When is PowerMock useful?
Only if you're stuck with legacy code that:

Has unmodifiable private static methods or final classes/methods.

Cannot be changed/refactored.

sadatmalik (approved):

camerojo (approved):

PR #2441, opened 2025-08-15
2435 potential bug in namenumber quick search

### LINKED ISSUE #2435: Potential bug in name/number quick search
From Hanine:

I've noticed an issue in the TC search functionality. Some candidates don't appear when searched by name, but their profiles are found when searched by their emails. 

For example, searching for <https://tctalent.org/admin-portal/candidate/143946|Wissam Azzam> (in the Name or Number field) yields no results, but his profile appears when searched by his email address or TBB ID; could it be a bug?

[Slack Message](https://refugeejobsmarket.slack.com/archives/C42S4AJ73/p1755093993495729)

sadatmalik:
This PR -

- adds a full name search clause - first + last - for candidate fast search
- add the reverse search clause - last + first
- adds the pg_trgm extension to support trigram GIN indexing - efficient for this type of partial sub-string search
- adds forward and reverse full name indexes on users 

Note - This PR is a Hotfix candidate - the Flyways SQL can be manually run on prod safely, the Flyways migration script will not fail with the subsequent release.


cazcam34 (approved):
Looks great but noticed it looks like there are other flways for other branches in your PR so unsure if they need to go in first.

camerojo (approved):

PR #2473, opened 2025-08-28
2414 noisy same-value logging eliminated

### LINKED ISSUE #2414: Chats potentially being resubscribed
When loading at the Jobs tab on the admin portal, if you look in logs you'll see that the log 'Visible chats read update: false/true' is logged repeatedly. See screenshare of my screen for an example of this.

[Private Zenhub Video](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNTNtQ2c9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--8b64dc376a7fd72aab4bccf8a80af899676ab506/Screenshare%20-%202025-07-31%203_17_18%20PM.mp4)

What is odd is that it never appears consistent, sometimes the logs don't appear often and sometimes they appear hundreds of times. Also the number does seem to increase upon page reloads. It doesn't seem to be a problem for performance, however it does indicate that something fishy may be going on and it could be a potential memory leak in the long run. I did find this to be the case when I found this bug while I was working on a new reusable nav tabs component which was causing a resubscription from the tabs and so this log/process was reoccurring over and over and then causing a crash.

Discussing with John we wondered if on the server the subscription isn't being closed therefore multiple subscriptions are occurring?

The log occurs in **FilteredOppsComponentBase** and the **show-candidates-with-chat** components, where the code is pretty much duplicated. See example in FilteredOppsComponentBase here.

Higher up this is called in the chat service method combineChatReadStatuses where it takes each chat and checks the read status of that chat to determine if overall there are unread chats (at least 1) or if there are none.

To check for each chat status the getChatIsRead$ observable is called and here the API call getUserChatInfo is subscribed to checks the server for the status. 

Looking at the network logs I can see that this API call is made repeatedly also, and when I sort by URL i can see that the same chat is being fetched. Whats odd is not all chats are fetched twice. Unsure if this is irrelevant but thought I'd put here incase its helpful.

samschlicht:


camerojo (changes requested):
This looks great but I struggled a bit to understand what was going on when the comments are referring to a single chat per candidate - for example,  "Map of candidate id to chat" - when in fact there can be multiple chats per candidate id.

To help anyone else trying to get their head around this code could you update the comments to explain why there might be multiple chats per candidate.

camerojo (approved):

sadatmalik (approved):

PR #2478, opened 2025-09-01
2289 added the alert component

hiba-machfej:
This is how the alert component looks like now

Screenshot 2025-09-01 at 3 51 23 PM


cazcam34 (dismissed):
Nice!

camerojo (commented):
Chatting with Caroline, she explained that using SVG gives better performance. But in this area of our code I don't think that performance is really an issue so I would prefer to trade performance for simplicity by using standard icons from Fontawesome.

cazcam34 (approved):

camerojo (approved):

PR #2481, opened 2025-09-01
305 keep url on jwt token expired

sadatmalik:
This PR -

- Adds an http interceptor for auth expiry
- Redirects to /login with the target url passed in returnUrl
- The user is thus returned to the target url after re-authenticating
- Includes unit tests


camerojo (dismissed):
Nice work! I had assumed that we could do it all on the back end but makes sense that we also need this browser logic.

cazcam34 (dismissed):
Nice!!

PR #2509, opened 2025-09-12
2288 accordion   figma to code

### LINKED ISSUE #2288: Accordion - Figma to code
We currently use ngb-accordion (angular own implementation of bootstrap), however the version we are using is currently deprecated anyways so will need to look at an update.

Creating our own shared accordion will be best I think

cazcam34:
How it looks:
Screenshot 2025-09-12 at 1 09 34 pm

Added a primary hover color to the text - that can be changed.

Also added ability for an optional custom header to be added (eg. in intakes we have buttons in the accordion panels and display summary data).

As I mentioned in our slack channel, we are missing some font weights so despite the header using a label-md mixin and the body using a paragraph mixin, which should have different weights they appear the same. If we can get the additional font weights working in dev I think this will look great

And heres how it looks with a custom header
Screenshot 2025-09-12 at 3 40 35 pm

### DIFF
--- a/ui/admin-portal/src/app/shared/components/accordion/accordion-item/tc-accordion-item.component.html
+++ b/ui/admin-portal/src/app/shared/components/accordion/accordion-item/tc-accordion-item.component.html
@@ -1,6 +1,10 @@
 <div class="tc-accordion-item">
   <a class="tc-accordion-header" (click)="toggle()">
-    {{ title }}
+    <!-- If exists custom header projected using attribute selector -->
+    <ng-content select="[custom-header]"></ng-content>
+
+    <span>{{ header }}</span>
+
     <tc-icon size="sm" color="gray">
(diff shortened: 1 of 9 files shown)


hiba-machfej (approved):
All look great 🙌 I was wondering if we could remove the arrow icon for the custom header, as I remember we had this style where a simple button would open up the accordion (or maybe add type button where a button can open it up).

cazcam34:
> All look great 🙌 I was wondering if we could remove the arrow icon for the custom header, as I remember we had this style where a simple button would open up the accordion (or maybe add type button where a button can open it up).

Ah yes sorry that's a bit confusing in my custom header example, the + button is meant to be like in the intake where we have Citizenships and that button adds a Citizenship to the accordion content. As opposed to a button that opens the accordion. 
Screenshot 2025-09-16 at 10 03 03 am

So the arrow is still there to indicate the open status of the accordion panel

camerojo (approved):

PR #2515, opened 2025-09-16
2513 list import should recognize publicids as well as candidate numbers

camerojo:


camerojo:
Hotfix has pulled in too many changes

PR #2530, opened 2025-09-17
2525 add TcDatePicker - only have TcDateRangePicker

### LINKED ISSUE #2525: Need TcDatePicker as well as existing TcDateRangePicker
Currently the new shared date range picker only emits an event on date picked, but to keep consistent with our usage of the date pickers in reactive forms we need to allow for a form control to be passed in.

Update!: This is actually how our current date range picker already works, as we need to handle the two dates provided by the range picker and split that into two seperate form fields.

cazcam34:
I initially thought the date range picker needed to allow form control input, however the new shared TcDateRangePicker component works as intended (same as our current date range picker). However I did then realise we need a shared TcDatePicker with the new style also.
I moved the updated tc style for the date pickers to a global file, as both the date picker and date range picker need to use the same styles so put it in a global to avoid double ups in each components scss file. Also in a global file I could overwrite the ngbDatePicker bootstrap classes which is needed in the restyle too (without using a deprecated :ng-deep).
Screenshot 2025-09-17 at 3 43 56 pm


hiba-machfej (dismissed):
The angular test is failing but other than that all looks good.

camerojo (approved):

samschlicht (approved):

PR #2531, opened 2025-09-17
2295 added the loading component

hiba-machfej:
Added the loading component based on a component we already have I updated it to include both page and section according to what we need. It looks like this now:

Screenshot 2025-09-17 at 2 46 48 PM


samschlicht (dismissed):
Great work!

cazcam34 (dismissed):

samschlicht (approved):

cazcam34 (approved):

PR #2536, opened 2025-09-19
Fix a bug with the error AuthExpiryInterceptor.

### LINKED ISSUE #2520: Angular reporting of thrown errors no longer seems to work
This should display a sensible error. You can see the data sent back in the response but it is not being passed back to the Angular ErrorInterceptor.ts

Also reported in #2506

The easiest way to demonstrate this bug is to log in with a bad password. Instead of the expected error message, you get the same odd error which is the ErrorInterceptor.ts not handling nulls properly. That should be fixed (by replacing !== null with != null, which will also handle undefined). However the real problem is that the error response is not being passed into the code.

Image

camerojo:
It was not rethrowing the error correctly. It was using RxJS version 7+ syntax while we are currently on RxJS V6

Improved handling of undefined in ErrorInterceptor.

Removed unused Router

### DIFF
--- a/ui/admin-portal/src/app/services/auth-expiry.interceptor.ts
+++ b/ui/admin-portal/src/app/services/auth-expiry.interceptor.ts
@@ -1,9 +1,10 @@
-import { Injectable } from '@angular/core';
+import {Injectable} from '@angular/core';
 import {
-  HttpRequest,
-  HttpHandler,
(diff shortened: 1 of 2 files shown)


samschlicht (approved):
Nice 👍

sadatmalik (approved):
good catches!

PR #2558, opened 2025-09-30
2497 create robust stage badge colour retrieval functions

### LINKED ISSUE #2497: Create robust stage badge colour retrieval functions
Caroline has made a start in #2437 on a function to retrieve the appropriate colour for a given stage's badge display.

We both agree that the current version - which checks the string stage title (see below) - is fine as a placeholder (it works and the call structure makes sense) is not robust enough for future stage additions and edits. Working suggestion is to add badge colour to the stage enum. Will need to be done for both Jobs and Cases.

cazcam34:
Rather than set a colour for each stage enum, I thought it would make more sense to just group the stages by appropriate/similar stages. See the way I've grouped the stages to set the colours, and let me know your thoughts.
Screenshot 2025-09-30 at 3 29 10 pm
Screenshot 2025-09-30 at 3 29 05 pm

And how it appears as an example:

### DIFF
--- a/server/src/main/java/org/tctalent/server/api/admin/JobAdminApi.java
+++ b/server/src/main/java/org/tctalent/server/api/admin/JobAdminApi.java
@@ -245,6 +245,8 @@ public class JobAdminApi implements
             .add("jobCreator", shortPartnerDto())
             .add("skipCandidateSearch")
             .add("stage")
+            .add("closed")
+            .add("won")
             .add("starringUsers", shortUserDto())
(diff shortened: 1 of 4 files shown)


cazcam34:
How it appears in action:
Screenshot 2025-09-30 at 3 06 32 pm

samschlicht (approved):
Great idea to approach it this way, @cazcam34! Approved 👍

hiba-machfej (approved):

PR #2565, opened 2025-10-01
2563 update the TC tab to support the *ngIf

hiba-machfej:


camerojo (approved):

cazcam34 (approved):
Thanks for finding this!

Issue #2551, opened 2025-09-24
TcButton component missing colour variations


cazcam34:
Currently the tc-button component only has a type input which allows for ```primary, secondary, outline, plain```.

However this doesn't give us the option to have other brand colours such as green, red, gray etc (eg. how we use btn-success, btn-danger).

Also we only get the one type of outline button (which is a primary color), would be good to have flexibility in what colour out outline buttons can be.

Suggest separating type into two inputs:
```
type: solid, outline, plain
color: primary, secondary, gray, green, red etc. (Similar to how tc-icons are set up with colour variants)
```

PR #2566, opened 2025-10-01
2551 - 2562 update the TC button component

hiba-machfej:
This PR fixes the issue of the TcButton Disabled state which still triggers on click once. #2562 . It also works on adding the different variations for the button which are mentioned in issue #2551 .


hiba-machfej:
@cazcam34 I noticed in the issue #2551 you mentioned about separating type into two inputs type and color, which totally makes sense I was just worried that it might affect the work that is already done and might require re-fixing if the primary, secondary were used as type not color, what do you think?

camerojo (dismissed):

cazcam34:
> @cazcam34 I noticed in the issue #2551 you mentioned about separating type into two inputs type and color, which totally makes sense I was just worried that it might affect the work that is already done and might require re-fixing if the primary, secondary were used as type not color, what do you think?

@hiba-machfej yes I think that's fine, it's only early days that we are using the buttons so maybe if you fix the button to have type and colour and then create a new issue that's "implement updated tc-button" or something and one of us can just fix up the places we have already used this button

cazcam34 (dismissed):
I approved incase you want to do a separate issue with the type & color inputs, or you can just add to this PR and we can reapprove again later

hiba-machfej:
Hey @cazcam34.
This is how the button components look like, what do you think? 
Screenshot 2025-10-03 at 2 01 29 PM

cazcam34 (approved):
Looks awesome!! Love it.

camerojo (approved):

cazcam34:
Merged on behalf of @hiba-machfej

PR #2573, opened 2025-10-02
2528 add new inactive candidate status for candidates who relocate to safety independent of the tc

### LINKED ISSUE #2528: Add new inactive Candidate status for candidates who relocate to safety independent of the TC
This was raised by Hanine here https://refugeejobsmarket.slack.com/archives/C42S4AJ73/p1756979092740999

It covers candidates who registered on the TC but manage to move to a safe destination independent of TC or partners.

Until now, the best alternative for such candidates was to mark them with status "ineligible" based on the fact that they are now in a safe location with a durable solution. But this is a bit of an extended use of the ineligible status which is usually set soon after a candidate registers based on whether they are eligible for support when they registered. In this case we are typically dealing with a candidate who was in fact eligible but who have successfully relocated without help. It seems useful to have a status that records that.

Suggested name for new status: relocatedIndependently

Hanine suggested an automation when a candidate's location is set to a destination country and we have no other information to go on. For example, if a candidate has no currently active opportunity.

camerojo:


server/src/main/java/org/tctalent/server/service/db/impl/CandidateServiceImpl.java
```diff
@@ -3206,29 +3210,46 @@ public boolean deleteCandidateExam(long examId)
...
-                && countryReq != afghanistanCountryId && countryReq != ukraineCountryId) {
-                newStatus = "ineligible";
+    /**
+     * Checks for automatic status changes based on country and nationality changes
+     * @param countryReqId Requested country id
+     * @param nationalityReqId Request nationality id
+     * @param candidate Candidate
+     * @return new status or null if no change
+     */
+    private CandidateStatus checkForAutomaticStatusChanges(
```
sadatmalik, line 3220: Recommendation generated by Amazon CodeGuru Reviewer. Leave feedback on this recommendation by replying to the comment or by reacting to the comment using emoji.

The cyclomatic complexity of this method is 12. By comparison, 98% of the methods in the CodeGuru reference dataset have a lower cyclomatic complexity. This indicates the method has a high number of decisions and it can make the logic difficult to understand and test. We recommend that you simplify this method or break it into multiple methods.

sadatmalik (dismissed):

PR #2576, opened 2025-10-02
Fix: sql clause fixed per error message suggestion - worked a charm

### LINKED ISSUE #2532: Bug: potential duplicates filter doesn't work with new search functionality
Found in dev/staging:
Candidate Search > select a value for Potential Duplicate > server throws the following error.

```
Caused by: org.postgresql.util.PSQLException: ERROR: column "potentialduplicate" does not exist
  Hint: Perhaps you meant to reference the column "candidate.potential_duplicate".
```

Works fine when you check the box 'Use old way of searching'

samschlicht:


camerojo (approved):
Nice one Sam! My bad - but so much easier than Criteria isn't it?!!!

hiba-machfej (approved):

PR #2577, opened 2025-10-02
2323 create task and additional functionality for relocating family member travel doc upload and data entry

### LINKED ISSUE #2323: Create Task and additional functionality for relocating family member travel doc upload and data entry
See parent issue for background.

Noting that we already record family members at intake and subsequently add them as relocating dependents on candidate opportunities (choosing from the options defined at intake).

EhsanEhrari:


server/src/main/resources/db/migration/V1_376__add_candidate_family_members_form.sql
```diff
@@ -0,0 +1,12 @@
+insert into candidate_form (name, description)
+values ('FamilyDocForm',
+        'Enter details of your eligible relocating family members.');
+
+
+insert into task (name, display_name, task_type, description, days_to_complete,
+                  created_by, created_date, candidate_form_id)
+values ('familyDocFormTask','Enter details of your eligible relocating family members',
```
camerojo: Display name should be shorter than this. This is more like a description. This name will appear in the drop down of tasks to be completed. I suggest "Complete relocating dependant data"

server/src/main/java/org/tctalent/server/model/db/FamilyDocForm.java
```diff
@@ -0,0 +1,73 @@
...
+  private static final String FAMILY_MEMBERS_JSON = "familyMembersJson";
+  private static final String NO_ELIGIBLE = "noEligibleFamilyMembers";
+  private static final String NO_ELIGIBLE_NOTES = "noEligibleNotes";
+
+  @Override
+  public String getFormName() {
+    return "FamilyDocForm";
+  }
+
+  @JsonProperty("familyMembersJson")
```
camerojo: Every Spring entity is automatically converted to Json with the same property name as the Java property. Normally you only need to specify @JsonProperty if you want the Json property to have a different name to the Java property.

ui/candidate-portal/src/app/components/form/family-doc-form/family-doc-form.component.ts
```diff
@@ -0,0 +1,248 @@
...
+
+      firstName: [value?.firstName ?? '', [Validators.required, Validators.maxLength(100)]],
+      lastName: [value?.lastName ?? '', [Validators.required, Validators.maxLength(100)]],
+      dateOfBirth: [value?.dateOfBirth ?? '', [Validators.required]],
+      gender: [value?.gender ?? 'other', [Validators.required]],
+      countryOfBirth: [value?.countryOfBirth ?? '', [Validators.required]],
+      placeOfBirth: [value?.placeOfBirth ?? ''],
+
+      // NEW fields
+      dependantHealthConcerns: [(value as any)?.dependantHealthConcerns ?? null],
```
camerojo: Why prefix these fields with "dependant"? You don't prefix other fields - eg dependantFirstName.

The publishing code that I wrote depends on dependant fields having the same names are the fields for the primary applicant. That way dependant data can be displayed in the same published columns as the corresponding fields for the primary applicant.

ui/candidate-portal/src/app/model/form.ts
```diff
@@ -44,3 +45,39 @@ export interface TravelDocFormData {
   travelDocIssueDate: string;
   travelDocExpiryDate: string;
 }
+
+export interface FamilyMemberDoc {
```
camerojo, line 49: Won't this be the same as the same structure for the primary applicant? Do we need a special interface for it?

camerojo (changes requested):
Looks really good Ehsan! I just have added a few comments.

Ideally we want dependants to have exactly the same data fields as the primary applicant.

ui/candidate-portal/src/app/model/form.ts
EhsanEhrari: Good point, you're right that it's quite similar to the primary applicant's structure.
For now, it includes a few extra fields specific to the family context (like relationship, healthConcerns, etc.), but we’ll refactor the entire interface model once all related epics are completed to consolidate and avoid duplication.

camerojo (approved):
Thanks Ehsan

samschlicht (approved):

PR #2590, opened 2025-10-07
2580 on admin portal and maybe on candidate portal you cannot clear a phone number field once the country area code has been entered

### LINKED ISSUE #2580: On admin portal (and maybe on candidate portal) you cannot clear a phone number field once the country area code has been entered
There is no way to clear a country code once it has been entered.

Image

cazcam34:
Now a form can be saved if there's just a country code, also the form can be saved if the phone number is removed (to remove the number)


camerojo (approved):
Thanks!

hiba-machfej (approved):

PR #2591, opened 2025-10-07
2584 extract skills from text   eg job description

### LINKED ISSUE #2584: Extract skills from text - eg Job Description

camerojo:
This is just a proof concept for a  potential Python service for extracting skills from text. That code is in SpacySkillsExtractor.py. 

SpacyPOC.py is just some code for playing around with embedding - which will come later. It is a bit of a play area, nothing more and will probably eventually be deleted.

This PR also contains an updated README.md related to Python.

There are no functional TC changes - so no risk of merging into staging.

### DIFF
--- a/server/src/main/python/SpacyPOC.py
+++ b/server/src/main/python/SpacyPOC.py
@@ -15,6 +15,8 @@
 """
 SpacyPOC.py
 
+Proof of concept for using spaCy to extract skills from a CV and a job description.
+
 Prerequisites
 -------------
 0. The assumption is that you have installed a virtual environment in ~venv
--- a/server/src/main/python/SpacyPOC.py
+++ b/server/src/main/python/SpacyPOC.py
@@ -33,15 +33,18 @@ to specify a python interpreter running in the virtual environment.
    $ python3 SpacyPOC.py
 """
 
+"""
+NOTE: On an old Intel Mac I had to restrict the version on numpy to < 2
+although it worked on a new Mac with the latest numpy.
+"""
 
 
 # pip install spacy sentence-transformers pgvector psycopg2-binary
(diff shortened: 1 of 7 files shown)


server/src/main/python/SpacyPOC.py
```diff
@@ -0,0 +1,115 @@
...
+candidate_units = cv_sentences + cv_skills
+job_units = [job_description] + jd_skills  # query as a whole + explicit skills
+
+# --- Embed
+cand_vecs = embed_texts(candidate_units)
+job_vecs = embed_texts(job_units)
+
+# --- Score: job → candidate (max or mean over candidate chunks)
+scores = util.cos_sim(job_vecs[0], cand_vecs).cpu().numpy().flatten()
+top_k_idx = scores.argsort()[::-1][:5]
```
sadatmalik, line 111: Recommendation generated by Amazon CodeGuru Reviewer. Leave feedback on this recommendation by replying to the comment or by reacting to the comment using emoji.

When slicing sequential data types (for example, string, list, or tuple), we recommend that you don&#39;t use all three parameters (lower bound, upper bound, and stride). This can be difficult to read and understand. If all three parameters are required, considering doing two separate assignments (for example, one to stride and one to slice) or using `islice` from the itertools built-in the module. If you use stride, do not use a negative stride value without lower and upper bounds. For example,

```
nums = list(range(0,25))
### instead of this:
nums[5:19:3]
### do this:
temp = nums[5:19]
temp[::3]
### or do this:
itertools.islice(nums, 5, 19, 3) 
```

samschlicht (approved):
Nice, interesting stuff!

sadatmalik (approved):

PR #2618, opened 2025-10-20
2589 download of autogenerated cv not working for candidate 289669 on production

EhsanEhrari:
The issue was caused by unnormalized or unescaped Unicode characters (such as smart quotes ’) and raw HTML content within candidate job experience fields.
During PDF generation, these characters made the XHTML invalid, which caused Flying Saucer’s XML parser to throw a SAXParseException (“well-formed character data” error).
To fix this, I updated the cleanCandidateJobDescriptions method to normalize and sanitize text fields before PDF rendering.
Normalized Unicode punctuation in role, companyName, and description to prevent invalid XML characters.


hiba-machfej (approved):

camerojo (approved):
Nice one!

PR #2667, opened 2025-10-30
2545 UI rollover radio button tc component

### LINKED ISSUE #2545: UI rollover: radio button tc component
Not sure how many places we use this (whether we need to make a standalone component or a one-time fix) but it's in the select-list modal that is opened from show-candidates

cazcam34:


samschlicht (approved):
Great stuff!

camerojo (approved):

PR #2694, opened 2025-11-04
2326 create task and additional functionality for relocating family member rsd evidence and refugee id upload and data entry

EhsanEhrari:


samschlicht (dismissed):

camerojo (approved):

camerojo (changes requested):
Looking at the mismatch of names

server/src/main/java/org/tctalent/server/model/db/FamilyRsdEvidenceForm.java
```diff
@@ -0,0 +1,50 @@
...
+
+/**
+ * Candidate portal form that captures refugee status documentation for
+ * relocating family members. The payload is stored as JSON in a candidate
+ * property so it can flexibly represent the per-member data set and any
+ * associated uploads.
+ */
+public class FamilyRsdEvidenceForm extends CandidateFormInstanceHelper {
+
+  private static final String FAMILY_RSD_JSON = "familyRsdEvidenceJson";
```
camerojo: It is up to you to be consistent with the property names. Here you name this property as "familyRsdEvidenceJson"

But in the Property definition below you name the property differently.

I think that the above line should be:
private static final String FAMILY_RSD_JSON = "FAMILY_RSD_EVIDENCE_INFO";

server/src/main/resources/db/migration/V1_384__add_family_rsd_evidence_form.sql
```diff
@@ -0,0 +1,16 @@
...
+
+insert into task (name, display_name, task_type, description, days_to_complete,
+                  created_by, created_date, candidate_form_id)
+values ('familyRsdEvidenceFormTask', 'Family RSD Evidence',
+        'FormTask', 'Submit refugee status details evidence for each relocating family member.',
+        7, (select id from users where username = 'SystemAdmin'),
+        now(), (select id from candidate_form where name = 'FamilyRsdEvidenceForm'));
+
+insert into candidate_property_definition (name, label, definition, type)
+values ('FAMILY_RSD_EVIDENCE_INFO',
```
camerojo, line 13: Here you name the property as FAMILY_RSD_EVIDENCE_INFO

camerojo (changes requested):

camerojo (approved):

samschlicht (approved):

Issue #2634, opened 2025-10-23
UI Rollover inconsistencies to review


cazcam34:
Thought I'd create an issue where we can put all the inconsistencies that have naturally come up when we are working on similar but different components. We can add screenshots to the comments and each comment can be it's own inconsistency for review?

cazcam34:
### Header of a candidate opp & job opp
These should be as consistent as possible and follow the same layout. 
- header sizes
- subheader location (above or below main title header)
- icons, in dropdown or alongside? If we decide dropdown we should make consistent with searches/lists/candidates also.
- Breadcrumb, i suggest removing any breadcrumb as it's always at a very shallow level and not necessary in my opinion. 
<img width="2000" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBM1BSQ3c9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--46a99cb2df63058ff7943683bc7ebf060be072de/Screenshot%202025-10-23%20at%202.44.58%E2%80%AFpm.png" alt="Screenshot 2025-10-23 at 2.44.58 pm.png" />
<img width="2000" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBM1RSQ3c9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--43728e3e7c67c4c5a52ee9bf9b402518aabdd494/Screenshot%202025-10-23%20at%202.44.43%E2%80%AFpm.png" alt="Screenshot 2025-10-23 at 2.44.43 pm.png" />

cazcam34:
### Icons in tables
- color? I like the gray (or alternate to primary) as I think it leaves the focus on the primary elsewhere
- size? I think the size sm seems to be quite small and might be trickier to view/click, so I suggest a bit bigger or inherited from the font size
[...]

PR #2702, opened 2025-11-05
2696 Unifying Icons in tables and adding striped columns

hiba-machfej:
This is an issue mentioned in the UI Rollover inconsistencies to review #2634 

> Icons in tables
color? I like the gray (or alternate to primary) as I think it leaves the focus on the primary elsewhere
size? I think the size sm seems to be quite small and might be trickier to view/click, so I suggest a bit bigger or inherited from the font size

According to the notes above I made the icons small and added an alternate color the secondary. I thought of adding gray but then I thought maybe it might be too much gray in the page. 
Screenshot 2025-11-05 at 7 40 14 AM

### DIFF
--- a/ui/admin-portal/src/app/shared/components/table/tc-table.component.scss
+++ b/ui/admin-portal/src/app/shared/components/table/tc-table.component.scss
@@ -8,6 +8,16 @@
   border-radius: 12px;
   padding: 32px 0 48px;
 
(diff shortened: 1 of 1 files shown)


samschlicht (approved):
I like it!

EhsanEhrari (approved):

PR #2744, opened 2025-11-13
fix: discarded directive, now uses disabled input on textarea

### LINKED ISSUE #2601: Bug: (UI rollover) 'disabled' input to tc-textarea not working
When `[disabled]="true"` is passed to the tc-textarea component, it isn't working.

Once fixed, uncomment this test, which should work: 
`candidate-shareable-notes-component.spec.ts#it('should disable textarea if not editable'`

samschlicht:


ui/admin-portal/src/app/components/util/candidate-shareable-notes/candidate-shareable-notes.component.ts
```diff
@@ -61,16 +61,13 @@ export class CandidateShareableNotesComponent extends AutoSaveComponentBase
...
+      this.form.controls['shareableNotes'].patchValue(
+        this.candidate.shareableNotes,
+        { emitEvent: false }
+      );
     }
   }
 
-  isEditable(): boolean {
-    return this.authService.isEditableCandidate(this.candidate);
-  }
```
samschlicht, line 75: This is not required - the auth check takes place in the parent and is passed to the component as the `editable` input. Visibility is also handled by the parent. The component isn't used in any context where the parent doesn't handle the candidate.

hiba-machfej (approved):

camerojo (approved):

Issue #2778, opened 2025-11-24
Candidate search results: make Dependants column sortable?


samschlicht:
Since this is a @Transient property it has to be differently handled in our new search functionality in order to be sortable.

`CandidateSearchUtils.buildOrderByClause()` would be a good place to start — the SQL clause will need to compute the count and then order by it, instead of just inserting the column (which obviously doesn't exist).

Might also be a good opportunity to review Latest Intake and Latest Intake Date, which are also computed and non-sortable (particularly given we do now store mini and full intake dates in the candidate table).

Labelling this as a question for now: I don't see it as a showstopper and we may not want to complicate our search code just for this reason. We also may find that the intake columns are appreciated by users as they are.

PR #2780, opened 2025-11-24
2768 bug sort search results by max education level error

### LINKED ISSUE #2768: Bug: sort search results by max education level error
1. Submit any new search
2. Attempt to sort by 'Highest level of edu' column

Works with old way of searching but not the new. Suggest checking all columns for similar issue.

```
org.hibernate.exception.SQLGrammarException: JDBC exception executing SQL [select distinct candidate.id,max_education_level.level from candidate left join users on candidate.user_id = users.id left join null where candidate.status in ('active','incomplete','pending','unreachable') and candidate.id not in (select candidate_id from candidate_saved_list where saved_list_id = 8026) and users.partner_id in (204,3,1,205,207,6,5,10025,7,10030,10024,10034,10031,8) order by max_education_level.level ASC,candidate.id DESC fetch first ? rows only] [ERROR: syntax error at or near "null" Position: 129] [n/a]
```

samschlicht:


ui/admin-portal/src/app/services/candidate-field.service.ts
```diff
@@ -168,7 +168,7 @@ export class CandidateFieldService {
       new CandidateFieldInfo("Legal status", "residenceStatus", null,
         this.residenceStatusFormatter, null, true),
       new CandidateFieldInfo("Dependants", "numberDependants", null,
-        null, null, true),
+        null, null, false),
```
samschlicht, line 171: Noting that it's possible to make this column (which references an @transient property) sortable, but I felt the level of complexity was outside the scope of this bug fix: cases where a user requires it will be rare indeed, and this close to a release it didn't seem like a great use of my time and that of the PR reviewers. Better just to turn off sortability for the time being.

Created https://github.com/Talent-Catalog/talentcatalog/issues/2778 for future consideration.

server/src/main/java/org/tctalent/server/util/CandidateSearchUtils.java
```diff
@@ -301,7 +303,16 @@ private static boolean isSortedByRank(Sort sort) {
             propertyName = "candidate." + propertyName;
         }
 
-        propertyName = propertyName.replace("user.", "users.");
+        // Special-case replacements
+        Map<String, String> specialCases = Map.of(
+            "user.", "users.",
+            "maxEducationLevel.", "educationLevel."
```
camerojo, line 309: Not sure why this special case is needed. Why don't we use the correct field name in the sort?

The reason for the user to users replacement is because of old legacy tech debt where table "users" is the only table in the database that uses the plural in the table name. We didn't want to push that non standard anomaly all the way up into user code.

samschlicht: @camerojo the correct field path is actually `maxEducationLevelId`, but we don't want to order by that, we want to order by the level. It might be worth taking a look together, the issue is that the code as it's currently written will use the term before a full-stop (when there is one) in a field path and add that to the joins for the search query, e.g. `partner.name`, `users.email` — so in this case we need to interject the name of the table we need to join with to get the level.

It's something we might want to review but my priority for this release was just to fix the bug.

camerojo: OK - the intention was that you could sort by maxEducationLevel.level but maybe that doesn't work.

camerojo (approved):
Approved with comment - maybe minor issue to resolve in the future

hiba-machfej (approved):

PR #2811, opened 2025-11-26
2797 'Show published sheet' inline heading icon

hiba-machfej:


samschlicht (approved):
Thanks, @hiba-machfej !

camerojo (approved):

PR #2827, opened 2025-11-28
2801 train to hire add comments field to all four forms tasks

### LINKED ISSUE #2801: Train-to-Hire: add comments field to all four Forms Tasks
The regular comment field isn't viewable from List view or in the published sheet, so let's add it as a form field and candidate property to make admins' lives easier and potentially important comments unmissable.

Secondary: for clarity, if it's a Forms Task, the regular comments field displayed at the bottom of any Task should only become visible when the abandonment checkbox is checked.

camerojo:
You were right, @samschlicht, there is no reason not to rename the task names to add a prefix. (There is a bit of complexity renaming forms which is what I was thinking of - but task renaming is just a database update, as you said).

So I have added a standard prefix to all TTH properties and tasks. This will make selection of tasks in task assignment, and selection of properties in list publishing much clearer and simpler.


samschlicht (approved):
Great work, @camerojo. I won't lie, this one doesn't look like it was much fun!

PR #2899, opened 2025-12-10
2880 keyword text search should include candidate general notes

EhsanEhrari:


server/src/main/java/org/tctalent/server/model/db/Candidate.java
```diff
@@ -2612,6 +2612,9 @@ public void updateText() {
             .filter(CandidateAttachment::isCv)
             .map(CandidateAttachment::getTextExtract)
             .collect(Collectors.joining(" || "));
-        this.text = combinedJobText + " || " + combinedCvText;
+        String notesText = getShareableNotes();
+        this.text = Stream.of(combinedJobText, combinedCvText, notesText)
+            .filter(s -> s != null && !s.isBlank())
```
sadatmalik, line 2617: Approved.

Just a small suggestion: since we're already using Spring, we can simplify the filter by using StringUtils::hasText, which is a nice built-in way to handle these kinds of checks. For example:

.filter(StringUtils::hasText)

sadatmalik (approved):

server/src/main/java/org/tctalent/server/model/es/CandidateEs.java
```diff
@@ -266,6 +269,8 @@ public void copy(Candidate candidate, TextExtracter textExtracter) {
         this.externalId = candidate.getExternalId();
         this.candidateNumber = candidate.getCandidateNumber();
 
+        this.shareableNotes = candidate.getShareableNotes();
+
         this.miniIntakeCompletedDate = candidate.getMiniIntakeCompletedDate() == null ?
             null : candidate.getMiniIntakeCompletedDate().toInstant().toEpochMilli();
         this.fullIntakeCompletedDate = candidate.getFullIntakeCompletedDate() == null ?
```
camerojo, line 276: Just noting that this CandidateEs code relates to Elastic Search which we are phasing out - but not a problem to add this in I guess.

camerojo (approved):

server/src/main/java/org/tctalent/server/model/es/CandidateEs.java
EhsanEhrari: Yeah, I thought it's worth adding it as well.

server/src/main/java/org/tctalent/server/model/db/Candidate.java
EhsanEhrari: Since it is approved by 2 people, I will note it in future PRs edditing it.

PR #2903, opened 2025-12-10
2878 Implement intake refresh

### LINKED ISSUE #2878: Intake refresh - refresh buttons generally
See recent post from Yara Issa - https://refugeejobsmarket.slack.com/archives/C42S4AJ73/p1765189487893509

She does a mini intake then goes straight to doing a full intake and she has to do a refresh to see the mini intake changes she has just done on the full intake. (Noting that this is a little unusual - the whole reason for separating mini intake from full intake is that they are normally filled out one long after - eg weeks or months - the other.)

Maybe there is a more elegant way to resolve this - but the easy solution, given to her in reply to the post, is to refresh the browser.

The problem with refreshing the browser as a solution is that it is a very expensive operation - it reloads ALL the Angular code up from the server - not just the data we want refreshed. On some screens we get around this by having a special refresh button - which just reloads the data in the page.

I propose that we add that button to the full intake page - and maybe systematically to many other pages. It could be a button that is always there when appropriate (and visible, not hidden away in a menu).

- [ ] Add refresh button to full intake page
- [ ] Make refresh button (eg on search results) always visible - not hidden in dropdown menu

hiba-machfej:


sadatmalik:
Hi @hiba-machfej  — following the team meeting, just sharing an option John surfaced via GPT that would work across browser tabs.

GPT’s best recommendation was the **BroadcastChannel API**. (This may well align with the approaches you were already considering.)

Example Angular usage:

```ts
// cross-tab.service.ts
import { Injectable } from '@angular/core';

@Injectable({ providedIn: 'root' })
export class CrossTabService {
  private channel = new BroadcastChannel('my-app-channel');

  send(message: any) {
    this.channel.postMessage(message);
  }

  listen(callback: (data: any) => void) {
    this.channel.onmessage = (ev) => callback(ev.data);
  }
}
```

Usage:

```ts
constructor(private crossTab: CrossTabService) {
  this.crossTab.listen(msg => console.log('Received from another tab:', msg));
}

someAction() {
  this.crossTab.send({ type: 'LOGOUT' });
}
```

This handles same-origin tabs. For multiple users working mini/full intakes from different browsers, as John suggested, reverting to the idea of adding a refresh button on the intakes is the simplest fallback.

ui/admin-portal/src/app/components/util/intake/IntakeComponentTabBase.ts
```diff
@@ -128,6 +130,12 @@ export abstract class IntakeComponentTabBase implements OnInit {
    */
   examLabels: { [key: string]: string } = {};
 
+  /**
+   * Keeps track of which accordion panels are currently open.
+   * Used to restore the open panels when the candidate intake data is refreshed or reloaded.
+   */
```
sadatmalik: Nice touch!

sadatmalik (commented):
Really nice work on the cross-tab service so far @hiba-machfej  — it’s a neat approach and gets us most of the way to the behaviour we want. I had one small UX thought: it’s more about _how_ the user experiences the update rather than the underlying mechanism.

Right now the auto-refresh will fire immediately when another tab updates the intake. I wonder if this could be disruptive if the user is mid-edit — especially with unsaved changes — as an unexpected refresh may interrupt their flow or, in a race condition, even cause input to be lost if the data hasn’t been saved yet.

A proposed adjustment: Instead of auto-refreshing, we add a "Refresh" button on the intake tabs, along the lines of @camerojo 's initial suggestion. When new data is available, the button (or its text) could visually change — colour shift, badge, subtle highlight, etc. — signalling that updated data exists and that the user can pull it in when ready. 

This adjustment mirrors for example GitHub’s pattern during PR reviews: when the underlying source changes, GitHub surfaces an orange “Refresh” indicator, letting the reviewer know the content is out of date without automatically reloading the page.

Same cross-tab sync benefits, but with user control and no surprise refreshes.

What do you think?

hiba-machfej:
> Really nice work on the cross-tab service so far @hiba-machfej — it’s a neat approach and gets us most of the way to the behaviour we want. I had one small UX thought: it’s more about _how_ the user experiences the update rather than the underlying mechanism.
> 
> Right now the auto-refresh will fire immediately when another tab updates the intake. I wonder if this could be disruptive if the user is mid-edit — especially with unsaved changes — as an unexpected refresh may interrupt their flow or, in a race condition, even cause input to be lost if the data hasn’t been saved yet.
> 
> A proposed adjustment: Instead of auto-refreshing, we add a "Refresh" button on the intake tabs, along the lines of @camerojo 's initial suggestion. When new data is available, the button (or its text) could visually change — colour shift, badge, subtle highlight, etc. — signalling that updated data exists and that the user can pull it in when ready.
> 
> This adjustment mirrors for example GitHub’s pattern during PR reviews: when the underlying source changes, GitHub surfaces an orange “Refresh” indicator, letting the reviewer know the content is out of date without automatically reloading the page.
> 
> Same cross-tab sync benefits, but with user control and no surprise refreshes.
> 
> What do you think?

Thank you for the suggestion Sadat, I'll start working on it!

hiba-machfej:
Hey @sadatmalik 
can you check the new approach in the PR?

[Private Zenhub Video](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNTFqREE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--0cff8f9c88e3d00db8f50fa46b395954ce94218b/new%20apprach%20--.mp4)



The refresh button is always visible. 
From one browser to another if there is any update it will change the button style and notify the user to refresh the data. 
From one intake tab to another inside the same browser tab it just updates the data.

sadatmalik:
Video looks great @hiba-machfej, thank you! I’ll go through the PR tomorrow.

ui/admin-portal/src/app/services/cross-tab-sync.service.ts
```diff
@@ -0,0 +1,64 @@
...
+
+
+/**
+ * Channel name used to communicate candidate intake updates
+ * between different browser tabs or windows.
+ */
+const CHANNEL = 'tc-candidate-intake-updates';
+
+@Injectable({ providedIn: 'root' })
+export class CrossTabSyncService {
```
camerojo, line 28: This is great - and important new functionality for the TC. I think it needs some more document with links to the Broadcast Channel API, what it does and why we need it.

It seems that Broadcast Channel API is not an Angular thing but a general API supported by all browsers. I think the best definitive doc on it is here https://developer.mozilla.org/en-US/docs/Web/API/Broadcast_Channel_API.
We should have a link to that doc in the doc for this service.

camerojo (approved):
Well done getting this done Hiba!

As I mention in my other comment, the Broadcast Channel API is an important API beyond just Angular. In 2.5.0 I think it will be worth us building a more general Broadcast Channel Service so that we can make full use of this API.

I will create an issue for it.

sadatmalik (approved):

PR #2910, opened 2025-12-11
2886 search by multiple candidate numbers

### LINKED ISSUE #2886: Search by multiple candidate numbers
See this Slack conversation with Mari:
https://refugeejobsmarket.slack.com/archives/C42S4AJ73/p1764959038935349

The idea is to add a search filed that would support retrieval of multiple candidates by their candidate numbers.

EhsanEhrari:


sadatmalik (dismissed):
I think the placement of the candidate-numbers field is fine for now.

When we take on the UX redesign epic, the search screen will be reworked anyway, so that will be a time to revisit the positioning of all fields, possibly including this one.

@hiba-machfej — the search screen redesign might be a good entry point when we kick off that epic.

server/src/main/java/org/tctalent/server/service/db/impl/SavedSearchServiceImpl.java
```diff
@@ -1859,6 +1860,12 @@ private Page<Candidate> doSearchCandidates(SearchCandidateRequest searchRequest)
         // Modify request, doing standard defaults
         addDefaultsToSearchCandidateRequest(searchRequest);
 
+        // Handle candidate numbers
+        Page<Candidate> candidatesByNumbers =
+            searchByCandidateNumbers(searchRequest, excludedCandidates);
+        if (candidatesByNumbers != null) {
+            return candidatesByNumbers;
+        }
```
camerojo: This ignores the possibility of combining this filter with other filters.

server/src/main/java/org/tctalent/server/service/db/impl/SavedSearchServiceImpl.java
```diff
@@ -1924,6 +1931,60 @@ private Page<Candidate> doSearchCandidates(SearchCandidateRequest searchRequest)
...
+            PageRequest.of(request.getPageNumber(), request.getPageSize())
+        );
+
+        // 2) Remove excluded candidates
+        List<Candidate> filtered = page.getContent().stream()
+            .filter(c -> !excludedCandidates.contains(c))
+            .collect(Collectors.toList());
+
+        // 3) Maintain input order (important for user experience)
+        Map<String, Integer> orderByNumber = new HashMap<>();
```
camerojo: Search sorts are controlled by clicking on the results column. 

You can't assume that the user wants the results sorted by candidate number.

server/src/main/java/org/tctalent/server/service/db/impl/SavedSearchServiceImpl.java
```diff
@@ -1924,6 +1931,60 @@ private Page<Candidate> doSearchCandidates(SearchCandidateRequest searchRequest)
...
+        // 4) Return as a new paginated result
+        return new PageImpl<>(
+            filtered,
+            request.getPageRequest(),
+            filtered.size()
+        );
+    }
+
+
+
```
camerojo: Our searches translate into SQL and it is the SQL logic that returns the results which will be a combination of all the filters involved in the search. We don't have logic like the searchByCandidateNumbers  method above - ie with separate methods for each kind of filter.

The real search logic now takes place in SavedSearchServiceImpl.extractJoinAndWhereSQL. See, for example, the search logic in there for filtering by candidate status.

<pre>
       // STATUS SEARCH
        if (!ObjectUtils.isEmpty(request.getStatuses())) {
            String values = request.getStatuses().stream()
                .map(Enum::name).map(val -> "'" + val + "'").collect(Collectors.joining(","));
            ands.add("candidate.status in (" + values + ")");
        }
</pre>

Your candidate number filter will look a bit like that.

(You can ignore code in CandidateSpecification.java. That is the old way of searching that we will be removing altogether - probably in 2.5.0)

camerojo (changes requested):
See other comments.  But another major feedback is that this code doesn't support saving of searches. Saved Searches are saved - including all the filter values so that individual searches can be named and reused.

See the SavedSearch entity SavedSearch.java.

server/src/main/java/org/tctalent/server/service/db/impl/SavedSearchServiceImpl.java
EhsanEhrari: @camerojo , thanks a lot for the detailed feedback that was very helpful.

You’re absolutely right: my initial approach with a separate searchByCandidateNumbers path was incorrect and didn’t fit the existing search architecture, especially around saved searches, joins, and SQL-driven filtering. Sorry about that.

I’ve now removed that logic and reworked candidate number filtering to follow the same pattern as the other filters. The candidate numbers are handled directly in extractJoinAndWhereSQL, so they:

participate in the main SQL search,

combine correctly with all other filters.

New changes have been pushed to the PR.
Please let me know if you’d like the candidate number filter to behave differently.

Thanks again for the guidance.

server/src/main/java/org/tctalent/server/service/db/impl/SavedSearchServiceImpl.java
```diff
@@ -1826,6 +1827,26 @@ List<Long> getIdsFromString(String listIds){
...
+    List<String> getStringListFromString(String csv) {
+        return csv != null
+            ? Stream.of(csv.split(","))
+            .map(String::trim)
+            .filter(StringUtils::hasText)
+            .collect(Collectors.toList())
+            : null;
+    }
+
+
```
camerojo: These are quite nice general purpose string utilities. How about extracting them out into a nice little reusable utility - a bit like, for example, org/tctalent/server/util/EnumHelper.java.

StringHelper?

server/src/main/java/org/tctalent/server/service/db/impl/SavedSearchServiceImpl.java
```diff
@@ -2267,6 +2229,19 @@ private String extractJoinAndWhereSQL(SearchCandidateRequest request,
...
+                .filter(StringUtils::hasText)
+                .map(String::trim)
+                .map(val -> val.replace("'", "''"))
+                .map(val -> "'" + val + "'")
+                .collect(Collectors.joining(","));
+
+            ands.add("candidate.candidate_number in (" + values + ")");
+        }
+
+
```
camerojo: Can this call standard string utilities - see comment above

server/src/main/java/org/tctalent/server/service/db/impl/SavedSearchServiceImpl.java
```diff
@@ -1757,6 +1758,7 @@ private SearchCandidateRequest convertToSearchCandidateRequest(SavedSearch searc
 
         // Check if the saved search countries match the source countries of the user
         List<Long> requestCountries = getIdsFromString(search.getCountryIds());
+        searchCandidateRequest.setCandidateNumbers(getStringListFromString(search.getCandidateNumbers()));
```
camerojo: Can you move this line up so that it belongs with similar lines above?

At the moment it is located under the 
//Check ... source countries of user 
comment - which is confusing. It doesn't have anything to do with that.

camerojo (changes requested):
Looks great. Just a couple of really minor changes!

camerojo (dismissed):
Looks good!

sadatmalik (dismissed):

hiba-machfej (dismissed):

hiba-machfej (approved):

camerojo (approved):

PR #2914, opened 2025-12-12
2906 Fix the search paging control and allow tables to have top and bottom pagination

hiba-machfej:


ui/admin-portal/src/app/components/search/define-search/define-search.component.html
```diff
@@ -907,7 +907,7 @@
                        [pageNumber]="pageNumber"
                        [pageSize]="pageSize"
                        [searchRequest]="searchRequest"
-                       [manageScreenSplits]="false"
+                       [manageScreenSplits]="true"
```
sadatmalik, line 910: Nice catch @hiba-machfej 

Though I think there may be a bug here. No matter whether I set manageScreenSplits to true or false, show-candidates picks up the coded default of true.

I'll add a screenshot from my debugging to the PR comments

sadatmalik (approved):

ui/admin-portal/src/app/components/search/define-search/define-search.component.html
hiba-machfej: Thank you Sadat!  
I added a few screenshots to the related issue too.

samschlicht (approved):

PR #2916, opened 2025-12-13
Add spacing between left panel and candidate card in split view layout

sadatmalik:
Tiny PR that adds spacing between left panel and candidate card in split view layout consistent with Bootstrap’s layout rhythm


samschlicht (dismissed):

camerojo (dismissed):
Approved but is it OK to put a comment in the css file - along the lines of the comment that you have made on the PR about the purpose of this change. It would help someone like me who is not a css or bootstrap expert understand it.

PR #2917, opened 2025-12-14
2901 tth relationship to candidate field not available to be published from list

### LINKED ISSUE #2901: TTH: relationship to candidate field not available to be published from list
Nice-to-have because at present dependants aren't able to move via this program. But it's anticipated they may be and this would definitely be a nice field to have available to be published.

The candidate gives the relationship in the document info forms tasks, the field just isn't defined so as to be publishable.

EhsanEhrari:


samschlicht (approved):

samschlicht (dismissed):

ui/candidate-portal/src/app/model/form.ts
```diff
@@ -52,7 +52,6 @@ export interface TravelInfoFormData {
  * into each column based on the field key or property name of the column.
  */
 export interface RelocatingDependant {
-  relationship?: DependantRelations;
   dependantRelationOther?: string;
```
camerojo: Maybe we also need a property for dependantRelationOther? Otherwise we won't be able to display it.

eg TTH_IT$OTHER_RELATIONSHIP_TO_PRIMARY_APPLICANT

camerojo (changes requested):
Looks good. I think we probably need one more property - see comment.

ui/candidate-portal/src/app/model/form.ts
EhsanEhrari: Nice catch!

EhsanEhrari: Done!

camerojo (approved):
Great1

sadatmalik (approved):

PR #2919, opened 2025-12-15
Allow default sort direction

### LINKED ISSUE #2918: Allow overriding of default sort order of candidate results by column
In particular, the default sort order of the Rank column should be descending (ie best matches first).

Equally default sort order for candidate number should be descending (most recent matches first).

camerojo:
Default sort for Rank and Candidate Number is descending

### DIFF
--- a/ui/admin-portal/src/app/components/candidates/show/show-candidates.component.ts
+++ b/ui/admin-portal/src/app/components/candidates/show/show-candidates.component.ts
@@ -538,8 +538,8 @@ export class ShowCandidatesComponent extends CandidateSourceBaseComponent implem
     this.doSearch(true);
   }
 
-  toggleSort(column: string) {
-    super.toggleSort(column);
+  toggleSort(column: string, sortDirection: string = 'ASC') {
+    super.toggleSort(column, sortDirection);
     this.doSearch(true);
   }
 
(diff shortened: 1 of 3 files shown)


ui/admin-portal/src/app/components/candidates/show/show-candidates.component.ts
```diff
@@ -538,8 +538,8 @@ export class ShowCandidatesComponent extends CandidateSourceBaseComponent implem
     this.doSearch(true);
   }
 
-  toggleSort(column: string) {
-    super.toggleSort(column);
+  toggleSort(column: string, sortDirection: string = 'ASC') {
```
samschlicht: @camerojo wondering if this variable should now be named `defaultSortDirection` — otherwise the template call is a bit confusing, i.e. on first reading there's an implication that however many times you click you'll get descending order. I see how it works in the superclass but it took a little investigation. Not a biggie so I'll approve anyway.

samschlicht (dismissed):

sadatmalik (dismissed):

ui/admin-portal/src/app/components/candidates/show/show-candidates.component.ts
camerojo: Good catch - I'll change it. Thanks @samschlicht

PR #2941, opened 2025-12-24
1286 tuning stage 2 of candidate results projecting to summary dtos instead of entities

### LINKED ISSUE #1286: Tuning, stage 2, of candidate results - projecting to summary DTO's instead of entities

camerojo:


server/src/main/java/org/tctalent/server/service/db/impl/CandidateDtoServiceImpl.java
```diff
@@ -0,0 +1,235 @@
...
+
+        if (!unprocessedIds.isEmpty()) {
+            LogBuilder.builder(log)
+                .action("CandidateDtoService.loadByIds")
+                .message("Candidate JSON missing after Redis + Postgres cache + recompute. " 
+                    + "requestedIds=" + ids + ", unprocessedIds=" + unprocessedIds 
+                    + ", versions=" + versions)
+                .logError();
+            throw new NoSuchObjectException(
+                "Candidates not found with ids: " + badIds
```
sadatmalik: Think this should log unprocessedIds, not badIds?

sadatmalik (dismissed):

sadatmalik:
Great stuff!

Approved with just one tiny comment

server/src/main/java/org/tctalent/server/service/db/impl/CandidateDtoServiceImpl.java
camerojo: Nice catch! Thanks!

Issue #25, opened 2023-05-17
Create a shared component library for portals


sadatmalik:
JC mentioned there is an opportunity for a shared ng component library - needs investigation / recommendation.

Issue #1896, opened 2025-02-13
Standardize on a single way of processing Fontawesome icons in Angular code


camerojo:
Currently the admin-portal and candidate-portal process Fontawesome icons differently.

We should standardize in the admin-portal way.

### Different package.json packages

Admin-portal 
    "@fortawesome/fontawesome-free": "^6.2.0",

Candidate-portal
    "@fortawesome/angular-fontawesome": "^0.10.2",
    "@fortawesome/fontawesome-free": "^6.2.0",
    "@fortawesome/fontawesome-svg-core": "^6.2.0",
    "@fortawesome/free-solid-svg-icons": "^6.2.0",

The angular-fontawesome package adds some extra support around using icons but it doesn't seem like we need it. 
https://github.com/FortAwesome/angular-fontawesome

sadatmalik:
**Reminder**: Once Font Awesome is standardised, re-check the candidate-portal accordion in the registration file upload to check that the open/close chevron icons render correctly. (See [#2952#discussion_r2664622296](https://github.com/Talent-Catalog/talentcatalog/pull/2952#discussion_r2664622296))

samschlicht:
- [x] remove unnecessary Angular FontAwesome dependencies from CP
- [x] update CP templates to match AP convention 
i.e. `<fa-icon icon="external-link-alt"></fa-icon>` -> `<i class="fas fa-spinner fa-spin">`
- [x] check all icons incl. above-noted

sadatmalik:
Resolved conflicts and merged.

sadatmalik:
Confirming that after merging #1896 the open/close accordion arrows are now appearing:

<img width="708" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBdys1REE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--5ca4c2a46054f2316eebcb961ef74f15980212c2/Screenshot%202026-01-27%20at%2010.58.37.png" alt="Screenshot 2026-01-27 at 10.58.37.png" />

PR #2952, opened 2026-01-02
2951 remove ng bootstrap accordion dependency in candidate portal

sadatmalik:
This PR replaces the use of ngb-accordion in the candidate-portal registration upload UI with the existing custom accordion components (tc-accordion and tc-accordion-item).

@ng-bootstrap/ng-bootstrap v16 is required for Angular 17 and no longer supports ngb-accordion. This PR should be reviewed and merged before the Angular upgrade proceeds.

The minimum required accordion components have been copied from the admin-portal into a shared module in the candidate-portal, following the same implementation pattern already in use there.

There are more robust approaches for sharing common component libraries across portals; a separate spike ticket (#25) exists in the backlog to explore the longer-term solution.

### DIFF
--- a/ui/candidate-portal/src/app/shared/components/accordion/accordion-item/tc-accordion-item.component.html
+++ b/ui/candidate-portal/src/app/shared/components/accordion/accordion-item/tc-accordion-item.component.html
@@ -1,0 +1,17 @@
+<div class="tc-accordion-item" [class.open]="isOpen">
+  <a class="tc-accordion-header" (click)="toggle()">
+    <!-- If exists custom header projected using attribute selector -->
+    <ng-content select="[custom-header]"></ng-content>
+
+    <span>{{ header }}</span>
+
+    <tc-icon size="sm" color="gray">
+      <i *ngIf="!isOpen" class="fa-solid fa-chevron-down"></i>
+      <i *ngIf="isOpen" class="fa-solid fa-chevron-up"></i>
+    </tc-icon>
+  </a>
+
+  <div class="tc-accordion-body" *ngIf="isOpen">
+    <ng-content></ng-content>
+  </div>
+</div>
--- a/ui/candidate-portal/src/app/app.module.ts
+++ b/ui/candidate-portal/src/app/app.module.ts
@@ -17,6 +17,7 @@
 import {BrowserModule} from '@angular/platform-browser';
 import {NgModule} from '@angular/core';
 import {AppRoutingModule} from './app-routing.module';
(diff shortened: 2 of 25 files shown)


ui/candidate-portal/src/app/shared/components/accordion/accordion-item/tc-accordion-item.component.html
```diff
@@ -0,0 +1,17 @@
+<div class="tc-accordion-item" [class.open]="isOpen">
+  <a class="tc-accordion-header" (click)="toggle()">
+    <!-- If exists custom header projected using attribute selector -->
+    <ng-content select="[custom-header]"></ng-content>
+
+    <span>{{ header }}</span>
+
+    <tc-icon size="sm" color="gray">
+      <i *ngIf="!isOpen" class="fa-solid fa-chevron-down"></i>
```
hiba-machfej, line 9: @sadatmalik Just one small note here: in the admin portal we are using `Font Awesome via CSS`, while in the candidate portal we are using `@fortawesome/angular-fontawesome`. Because of this difference, the open/close accordion arrows are not appearing.

This is not a blocking issue, so I approved the PR to avoid affecting the rest of the Angular version update. I will create a separate issue and pick it up to add proper icon support here so the arrows appear correctly in the accordion. Alternatively, if you prefer, I can also push one more commit to fix this directly.

hiba-machfej (approved):

EhsanEhrari (approved):

ui/candidate-portal/src/app/shared/components/accordion/accordion-item/tc-accordion-item.component.html
sadatmalik: Thanks @hiba-machfej — as discussed in the team meeting, this was intentional as we want to keep the component code identical across the admin and candidate portals, which will make a future shared module easier.

The missing chevrons come down to differences in how Font Awesome is configured between the two portals. We should standardise on the admin-portal approach — there’s already an issue in the current sprint for this (#1896), which you might want to pick up.

Once that’s resolved, the chevrons should appear consistently in the candidate portal as well. I’ve added a note to #1896 to remind us to re-check this.

sadatmalik: Confirming that after merging #1896 the open/close accordion arrows are now appearing:

Screenshot 2026-01-27 at 10 58 37

PR #2981, opened 2026-01-09
2949 remove transient url field from candidate attachment replacing with redefinition of location field

### LINKED ISSUE #2949: Remove transient url field from Candidate Attachment replacing with redefinition of location field
The CandidateAttachment entity has a transient field called url which is calculated from other fields as follows:

<pre>
    public String getUrl() {
        if (type == AttachmentType.file) {
            url = "https://s3.us-east-1.amazonaws.com/files.tbbtalent.org/candidate/" 
+ (migrated ? "migrated" : candidate.getCandidateNumber()) + '/' + location;
        } else {
            url = location;
        }
        return url;
    }
</pre>

Having the hardcoded url path in there is not good and inflexible.

Instead, we should modify all location's in the database with a Flyway to convert them all to urls using the above logic.

Then there is no more need for url - the front end code can treat all locaton values as urls.

camerojo:
This turned out to be nastier than I thought but it is good to have it done.

We have not uploaded attachments to S3 since 2020, when we started uploading to Google Drive. So code in there for uploading to S3 is never used. So I removed it. If we do start uploading to S3, as planned, we will do it completely differently anyway.

We used to have a transient computed field called url - which was widely used, computed from the database "location" field. My Flyway converts ALL db "location" fields to urls. Then, for simplicity and to minimise the number of tricky changes, I deleted the transient "url" field and then mapped the db "location" field on to an entity field called "url".


EhsanEhrari (dismissed):

server/src/main/java/org/tctalent/server/model/db/CandidateAttachment.java
```diff
@@ -48,13 +46,10 @@ public class CandidateAttachment extends AbstractAuditableDomainObject<Long>  {
     private String name;
 
     /**
-     * For links {@link AttachmentType#link} and Google Docs
-     * {@link AttachmentType#googlefile}, the associated url.
-     * <p>
-     * For S3 files {@link AttachmentType#file}, it is the unique filename
-     * generated on S3.
+     * The attachment's url.
```
sadatmalik, line 49:
```suggestion
     * The attachment's url.
     * 
     * Historical note: This field maps to the "location" database column for backward 
     * compatibility. Prior to 2026, this column stored relative S3 paths for file-type 
     * attachments. Migration V1_400 converted all paths to full URLs.     
```

sadatmalik (dismissed):
Approved 👍 — small suggestion to add a bit more documentation around CandidateAttachment.url.

PR #3030, opened 2026-01-26
3007 Prevent candidate detail panel opening on icon click

### LINKED ISSUE #3007: Clicking on candidate icon in results opens card. Shouldn't. Click should be consumed. I think it used to be.
In search results (or a list of candidates), clicking on a particular candidate normally opens a display on the right displaying details for that candidates.

However, if you click on an icon or link (for example clicking on the little candidate icon to see the candidate page) the current code also displays the right hand candidate display, as well as activating the link.
I think it should just follow the link without displaying the right hand display. I think that this is how it used to work.

In other words the link should consume the click, rather than passing it on, triggering the right hand display.

hiba-machfej:
Added click event propagation prevention to the candidate number/icons cell so links and icons no longer trigger row selection, while clicks elsewhere on the row still open the candidate detail card

### DIFF
--- a/ui/admin-portal/src/app/components/candidates/show/show-candidates.component.html
+++ b/ui/admin-portal/src/app/components/candidates/show/show-candidates.component.html
@@ -584,7 +584,7 @@
                 </tc-icon>
               </div>
             </td>
-            <td>
+            <td (click)="$event.stopPropagation()">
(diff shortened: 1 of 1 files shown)


camerojo (approved):
Nice!!!

samschlicht (approved):

PR #3112, opened 2026-02-23
Handle candidate update on attachment modal dismissed

sadatmalik:
This PR updates attachment upload handling so that the attachment list is refreshed in the event that the attachment upload modal is dismissed.


EhsanEhrari (approved):
A small improvement, you could consider using .finally() instead of duplicating the updateCandidate() call in both then and catch, just to simplify the logic.

hiba-machfej (approved):

PR #3142, opened 2026-03-05
3135 bug load saved search throws incorrectresultsizedataaccessexception

EhsanEhrari:


samschlicht (dismissed):
Nice work, @EhsanEhrari! Feel free to contact Josue when this is merged - his request (see link in ZH issue) surfaced the bug and I'm sure he'd appreciate knowing you've dealt with it. 🙂

hiba-machfej (dismissed):

server/src/main/java/org/tctalent/server/service/db/impl/SavedSearchServiceImpl.java
```diff
@@ -954,7 +957,23 @@ public SavedSearch removeWatcher(long id, UpdateWatchingRequest request) {
...
+             * Avoid duplicates:
+             * We rely on the database unique index:
+             *   saved_list (saved_search_id, created_by) WHERE saved_search_id is not null
+             *
+             * If two requests race, Postgres will reject the second insert.
+             * In that case we re-fetch and return the row created by the first request.
+             */
+            try {
+                savedList = savedListRepository.save(savedList);
+            } catch (DataIntegrityViolationException ex) {
```
sadatmalik: Don't think this exception throws here (on save)

See:
https://stackoverflow.com/questions/41362841/spring-data-cannot-catch-dataintegrityviolationexception-in-service-layer

server/src/main/java/org/tctalent/server/service/db/impl/SavedSearchServiceImpl.java
```diff
@@ -954,7 +957,23 @@ public SavedSearch removeWatcher(long id, UpdateWatchingRequest request) {
...
+                //Another request created the selection list at the same time.
+                savedList = savedListRepository.findSelectionList(id, userId)
```
sadatmalik: If an exception does throw, don't think this will work with the poisoned session - good to  put in an integration test to verify 

See:
https://stackoverflow.com/questions/5109175/hibernate-session-is-invalidated-after-constraintviolationexception

Also a 400 InvalidRequestException (below) doesn't seem right - maybe a 500 - InternalSystemException or IllegalStateException ?

server/src/main/java/org/tctalent/server/repository/db/SavedListRepository.java
```diff
@@ -99,4 +99,8 @@ where sl.public_id in (:publicListIds)
      */
     Optional<SavedList> findByPublicId(String publicId);
 
+    @Query(value = "select pg_advisory_xact_lock(:savedSearchId, :userId)", nativeQuery = true)
+    void lockSelectionListKey(@Param("savedSearchId") long savedSearchId,
+        @Param("userId") long userId);
+
```
camerojo: As we discussed in our call, it is interesting to see what you did here, but I think it is not needed if there is a constraint added to the database to make the selection list unique by search and createdBy.

Once that constraint is in place any attempt to add a second selection list will fail and throw an exception which the user will see. That should help us understand exactly how this can happen and maybe come up with another way of preventing it.

camerojo (changes requested):

server/src/main/java/org/tctalent/server/repository/db/SavedListRepository.java
EhsanEhrari: Yes @camerojo , that constraint will be added as direct db SQL execution using manual db scripts. I am going to revert it back to it's original state.

camerojo (dismissed):

sadatmalik (dismissed):

sadatmalik (approved):

camerojo (approved):

Issue #3129, opened 2026-02-26
Write API allowing Angular authorization services to populate canViewChats method from server user policy


camerojo:
### Note
Depends on #3124

Currently the authorisation services for candidate portal and admin portal have a hard coded canViewChats method which means that NO users can see or subscribe to chats. 

For our testing purposes we have support on our server in ChatPolicy where certain test users can see and subscribe to chats. This is so we can continue testing chat code even though it is generally disabled.

The Angular code should know from the server to check whether a particular user (candidate or admin) should be able to see or create chats.

camerojo:
- [x] Instead of separate API call to server, return canViewChats on login response from server
- [x] Do on admin portal login
- [x] Do on candidate portal login

PR #3191, opened 2026-03-16
3180 re enable policy terms display and acceptance for grn

camerojo:
Note that this depends on #3129. Merge that in first.


ui/admin-portal/src/app/model/terms-info-dto.ts
```diff
@@ -16,7 +16,9 @@
  */
 
 export enum TermsType {
-  CANDIDATE_PRIVACY_POLICY,
+  GRN_CANDIDATE_PRIVACY_POLICY,
+  TBB_CANDIDATE_PRIVACY_POLICY,
+
   DATA_PROCESSING_AGREEMENT
```
sadatmalik: Should this be OPC_STANDARD_DATA_PROCESSING_AGREEMENT ?

(To match the server side enum)

server/src/main/java/org/tctalent/server/service/db/impl/UserServiceImpl.java
```diff
@@ -123,26 +128,8 @@ public class UserServiceImpl implements UserService {
...
-        this.countryRepository = countryRepository;
-        this.passwordHelper = passwordHelper;
-        this.authenticationManager = authenticationManager;
-        this.tokenProvider = tokenProvider;
-        this.authService = authService;
-        this.emailHelper = emailHelper;
-        this.partnerService = partnerService;
-    }
+    @Value("${tc.instance-type}")
+    private TcInstanceType tcInstanceType;
```
sadatmalik: I might move this to a  @ConfigurationProperties -- same pattern as you used for e.g. TcSkillsExtractionServiceProperties

And then constructor inject the properties

server/src/main/java/org/tctalent/server/service/db/impl/UserServiceImpl.java
```diff
@@ -123,26 +128,8 @@ public class UserServiceImpl implements UserService {
     @Autowired
     private CodeVerifier totpVerifier;
```
sadatmalik: Since you've refactored to using @RequiredArgsConstructor you could remove the @Autowired here as well I think for completeness. Making it final like the others will constructor inject it.

sadatmalik (commented):
Looks solid @camerojo 

There's a potential runtime bug I think (in the enum) 

Other than that just a couple of minor comments.

It's picked up some merge conflicts along the way, but otherwise looking good.

ui/admin-portal/src/app/model/terms-info-dto.ts
camerojo: Great catch!

server/src/main/java/org/tctalent/server/service/db/impl/UserServiceImpl.java
camerojo: Another good catch! Thanks.

server/src/main/java/org/tctalent/server/service/db/impl/UserServiceImpl.java
camerojo: I think that the application.yml properties needs a bit of a rethink before using @ConfigurationProperties more effectively. I am not sure I like where I have put instance-type.

sadatmalik (approved):

PR #3193, opened 2026-03-16
3041 add reference casi service

sadatmalik:
This pull request introduces a new minimal "reference" provider to the CASI platform, including its service code, task policies, default eligibility, resource allocation and importing logic, and API integration. The changes enable the platform to support a new reference provider for voucher distribution. The reference provider is visible in local dev environments only.

New Reference Provider Feature:

* Added the `REFERENCE` provider and `VOUCHER` service code to `ServiceProvider` and `ServiceCode` enums.
* Implemented the `ReferenceService` class, which provides the core logic for candidate assistance, resource allocation, and inventory importing for the reference provider.
* Added a minimal CSV importer (`ReferenceVoucherImporter`) for reference vouchers.
* Registered a resource allocator bean for reference vouchers in the CASI configuration.

API and Integration:

* Added the general `ServicesPortalController` with endpoints for eligibility, assignment, and resource status updates, scoped to the logged-in candidate.
* Updated the email notification listener to handle reference assignments with a no-op email policy.

Eligibility and Task Policy Infrastructure:

* Introduced general `EligibilityPolicy` interface and registry, with an example reference "is active" policy. New providers will implement EligibilityPolicy's to define their service eligibility criteria (defaults to always eligible)
* Added a specific reference task policy for task lifecycle hooks.


server/src/main/java/org/tctalent/server/casi/README.md
```diff
@@ -196,87 +223,336 @@ public class DuolingoTaskPolicy implements TaskPolicy {
...
+
+An example CSV for testing is at:
+`ui/admin-portal/src/app/components/casi-management/import-reference-vouchers/reference-vouchers-example.csv`
+
+```csv
+voucher_code,expires_at
+REF-TEST-0001,2026-12-31
+REF-TEST-0002,31/01/2027
+REF-TEST-0003,02/15/2027
+```
```
camerojo, line 558: This is so good! It is a model for all of us to (try to!) follow.

server/src/main/java/org/tctalent/server/casi/application/policy/EligibilityPolicyRegistry.java
```diff
@@ -0,0 +1,54 @@
...
+/**
+ * Registry of EligibilityPolicy implementations keyed by provider.
+ * Providers without an explicit policy default to eligible.
+ *
+ * @author sadatmalik
+ */
+@Component
+public class EligibilityPolicyRegistry {
+
+  private final Map<ServiceProvider, EligibilityPolicy> policies;
```
camerojo, line 34: It was a mystery to me how this is initialized.

Does Spring, when it sees this, automatically hoover up all components that implement EligibilityPolicy, notice that they all have a ServiceProvider, and automatically populate this Map???!!!

server/src/main/java/org/tctalent/server/casi/api/ServicesPortalController.java
```diff
@@ -0,0 +1,134 @@
...
+
+    ServiceAssignment assignment = assignments.stream()
+        .filter(a -> a.getResource() != null && a.getResource().getStatus() == ResourceStatus.RESERVED)
+        .findFirst()
+        .orElseGet(() -> assignments.stream()
+            .filter(a -> a.getResource() != null && a.getResource().getStatus() == ResourceStatus.REDEEMED)
+            .findFirst()
+            .orElse(null));
+
+    return ServiceAssignmentMapper.toDto(assignment);
```
camerojo, line 82: ServiceAssignmentMapper is another cool construct - decoupling the entity, the model and the dto.

We could do this a lot more in other places.

camerojo (approved):
I really just looked at the back end - which is where all the action happens I guess.

The overall functionally looks fine - and easily extensible.

There was also a bunch of Java and Spring coding patterns that I learned from.

Great stuff.

The README.md placed in the package was brilliant as well.

server/src/main/java/org/tctalent/server/casi/application/policy/EligibilityPolicyRegistry.java
sadatmalik: Yes that's pretty much it @camerojo 

The constructor takes List<EligibilityPolicy> policies. For that List parameter, Spring injects a list of every bean that implements EligibilityPolicy. 

So all our policy beans (e.g. ReferenceEligibilityPolicy) are collected into that list automatically. And the constructor then builds the Map

server/src/main/java/org/tctalent/server/casi/api/ServicesPortalController.java
sadatmalik: Agreed — we could use this pattern more. 

I went with a plain Java mapper here because it keeps the mapping steps explicit and easy to follow. Sometimes that’s clearer than MapStruct, where the “what’s happening” is generated. 

For this case, plain Java felt simpler and more obvious!

PR #3219, opened 2026-03-23
Update chatbot visibility logic for GRN instance type on local and staging only

### LINKED ISSUE #3218: Gate TC Chatbot by instance type and environment
Only show the TC Chatbot when both of these conditions are true:

- the TC instance type is GRN
- the application is running in staging

The chatbot must not be visible for TBB instances, and it must not be visible (yet) in production even for GRN.

Implement a combined feature gate covering both instance type gate and environment gate.

sadatmalik:
This PR -- 

- updates the logic for displaying the chatbot in the candidate portal
- replaces the previous partner-based check with GRN instance-type and local/staging environment check

### DIFF
--- a/ui/candidate-portal/src/app/components/app.component.html
+++ b/ui/candidate-portal/src/app/components/app.component.html
@@ -19,4 +19,4 @@
   Loading ...
 </ng-container>
 <router-outlet *ngIf="!loading"></router-outlet>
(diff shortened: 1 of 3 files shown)


EhsanEhrari (approved):
Good job!

hiba-machfej (approved):

camerojo (approved):

PR #3228, opened 2026-03-24
Propagate tab loading state & lazy-load Visa tab

### LINKED ISSUE #2990: Delayed when opening Mini Intake / Full Intake tabs in TC
hello, I'm facing an issue with TC, when I click a candidate's profile and wanna go through the tabs Education, Mini Intake, Full Intake, etc. by clicking on each, it doesn't trigger any action unless I refresh the page, which then will bring back the page/tab I clicked before the refresh. I already tried a different browser, cleared my browser's cache (I use Chrome), and changed WiFi network, same issue! Appreciate your feedback and help! <@U02D25E9TUP> <@U03TEBVBMUG> fyi

[Slack Message](https://refugeejobsmarket.slack.com/archives/C42S4AJ73/p1768300137266039?thread_ts=1768300137.266039&cid=C42S4AJ73)

EhsanEhrari:
Make intake tabs report loading state and lazily load data when activated. Added tabIsActive @Input and loadingChange @Output to IntakeComponentTabBase and CandidateVisaTabComponent; visa tab now uses ngOnChanges to load data only when the tab becomes active, emits loadingChange true/false around async calls, and improves error handling and form initialization. ViewCandidate now shows a page-level <tc-loading> for async tabs, passes tabIsActive to MiniIntake/FullIntake/Visa tabs and listens to loadingChange to control tabLoading. Misc: updated imports and minor formatting/refactorings.


ui/admin-portal/src/app/components/util/intake/IntakeComponentTabBase.ts
```diff
@@ -65,6 +74,9 @@ export abstract class IntakeComponentTabBase implements OnInit, OnChanges {
    */
   @Input() tabIsActive: boolean;
 
+  @Output() loadingChange = new EventEmitter<boolean>();
```
sadatmalik: All other properties in the base class have a doc comment - please can you update this property with a comment with a doc comment as well?

ui/admin-portal/src/app/components/candidates/view/view-candidate.component.ts
```diff
@@ -321,9 +331,26 @@ export class ViewCandidateComponent extends MainSidePanelBase implements OnInit,
...
+    const asyncTabs = ['MiniIntake', 'FullIntake', 'Visa'];
+
+    if (!asyncTabs.includes(activeTabId)) {
+      setTimeout(() => {
+        this.tabLoading = false;
+      });
+    }
+  }
+
+  onTabLoadingChange(loading: boolean) {
```
sadatmalik: Bit of a race condition here. . . ?
E.g. if a user switches from MiniIntake (which is still loading) to Visa, both tabs are alive. 

Then for instance MiniIntake's forkJoin completes and emits loadingChange(false), setting tabLoading = false in the parent even though Visa's data fetch may still be in progress. This can silently drop the spinner mid-load.

Not major but could misleading for users. Suggested including tabId in the event payload -- 
e.g., loadingChange emits { loading: boolean, tabId: string } -- and then the parent only responds if tabId === activeTabId

?

ui/admin-portal/src/app/components/candidates/view/view-candidate.component.html
```diff
@@ -223,6 +223,8 @@ <h2>{{candidate.user.firstName}} {{candidate.user.lastName}}</h2>
   <div class="row">
       <!-- Just have main panel taking up whole screen if they cannot view private data (case notes)-->
       <div class="col-sm-{{canViewPrivateInfo() ? mainPanelColWidth : totalPanelWidth}}">
+        <tc-loading type="page" [loading]="tabLoading"></tc-loading>
```
sadatmalik, line 226: This is good but I think the intention is to have it for initial tab loads?

createRecord, doDelete and reloadAndSelectViasCheck all call this.loadingChange.emot(true/false) which activate this same spinner every time a user creates or deletes a visa check record -- probably unintentional

Suggest a private _initialLoad flag -- set it on first load ad only emit loadingChanges when initialLoad is true? (Or restrict loadingChange.emit to loadVisaTabData () only?

Same concern for IntakeComponentTabBase.refreshIntakeData().  . . parent page level spinner is triggered from multiple places, not just initial load

EhsanEhrari: You’re right that the parent-level spinner should only represent the initial load when switching into a tab, not every later action inside the tab like create/delete/refresh.
I’ll tighten that up so loadingChange is only emitted for the initial tab activation load, and local tab actions will continue using the tab’s own internal loading state.

ui/admin-portal/src/app/components/candidates/view/view-candidate.component.ts
EhsanEhrari: Thanks, these are good catches.
I agree about the race condition. I’ll update the event payload to include the tabId, so the parent only responds to loading events from the currently active tab.

ui/admin-portal/src/app/components/util/intake/IntakeComponentTabBase.ts
EhsanEhrari: I’ll add a doc comment for the new loadingChange property in the intake base class to match the surrounding style.

hiba-machfej:
@EhsanEhrari there’s a merge conflict in this PR that needs to be resolved.

EhsanEhrari:
> @EhsanEhrari there’s a merge conflict in this PR that needs to be resolved.

@hiba-machfej thanks! I just fixed the conflicts.

sadatmalik (approved):

hiba-machfej (approved):

PR #3271, opened 2026-03-31
2529 migrate file storage to aws s3

### LINKED ISSUE #2529: Migrate file storage to AWS S3
Includes #3216

See also #3255 and #312

### Overview on using Postgres + S3 for storage.

As we do now, we store text extracted from, for example, CV files in the Postgres database so that we can efficiently do text searches (like we used to do with Elasticsearch).

The difference is that we will use S3 for storing the original files - instead of Google Drive.

We can continue to support exporting lists to Google Sheets where partners (such as TBB) have a Google Workspace account. This can be changed to be configurable by partner. TBB's Google Drive connection parameters will move to their partner record.

It seems like the best option is code directly to S3 for storing files but managing our own links (using TBB or GRN domains - eg https://files.tctalent.org/d/{doc_public_id} to those files. Cloudfront, which we are also already using, has good support for serving up files with our domain's urls.

See note below with suggestions from ChatGPT on how to do this.

#### Off the shelf file systems based on S3

- [S3 NIO FileSystem (by AWS Labs)](https://github.com/awslabs/aws-java-nio-spi-for-s3) -  Allows use of java.nio.file.Path, Files.read, Files.write, etc., but backed by S3.
- [Mountpoint for Amazon S3](https://docs.aws.amazon.com/AmazonS3/latest/userguide/mountpoint.html): Doesn't work with Fargate (requires FUSE which Fargate does not support)

There are others but they are only recommended if we already are managing files using a file system or Java NIO (Path, Files, and FileSystem). But we aren't. (Our existing code, which interfaces to Google Drive, sits behind a basic "home grown" abstraction of org/tctalent/server/util/filesystem/FileSystemBaseEntity.java)

camerojo:


server/src/main/java/org/tctalent/server/files/AttachmentAuthorizationService.java
```diff
@@ -0,0 +1,44 @@
...
+@Service
+public class AttachmentAuthorizationService {
+
+    /**
+     * Checks that the current user has permission to access the given attachment.
+     *
+     * @param attachment the attachment to check
+     * @throws AccessDeniedException if the current user does not have permission to access the attachment
+     */
+    public void assertCurrentUserCanAccess(CandidateAttachment attachment) {
```
samschlicht, line 36: @camerojo just checking this was deliberately left as TODO for this PR?

samschlicht: I see that the only caller will just proceed but just checking!

samschlicht (dismissed):
Approved with one query

server/src/main/java/org/tctalent/server/files/AttachmentAuthorizationService.java
camerojo: Nice catch. Thanks. This is only used for files that are being access through a logged in user.

Currently, attachments are all potentially accessible to anyone who has the link - and come in through the public API CandidateFilesController and so don't have a logged in user associated.

So this is reserved for the future. But the TODOs are misleading. I will fix that.

sadatmalik (approved):

PR #3365, opened 2026-04-21
Closes TC-1212 - 3364 build a parallel docx export path

### LINKED ISSUE #3364: Build a parallel DOCX export path
Build a parallel **DOCX** export path beside the existing PDF path. Right now, the admin API exposes POST /api/admin/candidate/{id}/cv.pdf, fetches the candidate, and calls candidateService.generateCv(candidate, showName, showContact). That service currently just delegates to **pdfHelper.generatePdf(...)**.

EhsanEhrari:


sadatmalik (dismissed):
Hi @EhsanEhrari I see that you factored out a couple of cleaner methods from PdfHelper to CvExportDataPreparer but in doing so I don't understand why the HtmlSanitization has been dropped.

The HtmlSanitization is there for a reason, prevent XSS attacks for example. Can you confirm why it was removed?

(Also DocxHelper is missing a license header).

EhsanEhrari:
> Hi @EhsanEhrari I see that you factored out a couple of cleaner methods from PdfHelper to CvExportDataPreparer but in doing so I don't understand why the HtmlSanitization has been dropped.
> 
> The HtmlSanitization is there for a reason, prevent XSS attacks for example. Can you confirm why it was removed?
> 
> (Also DocxHelper is missing a license header).

Hi @sadatmalik , thanks for catching this. There’s no intended reason for removing `HtmlSanitization` I mistakenly used `StringSanitizer` there.

You’re absolutely right that the HTML sanitization is needed to help prevent XSS issues. I’ll fix that and also add the missing license header to `DocxHelper`.

camerojo:
Thinking more about this, I don't understand why we need a parallel DOCX path for CVs.

We deliberately created CVs as PDF's because

- CVs are not normally intended to writeable - they are normally read only PDFs
- We want to avoid relying on a proprietary format (Microsoft's DOCX).

I just checked and PDFs can be uploaded to Google Drive, then opened as Google Docs and then translated. I believe that was the plan for the DOCX files - but if you can do the same with PDF's, why duplicate our CV generation to include DOCX?

camerojo (changes requested):
See my comment. I don't understand why we need a parallel docx path.

EhsanEhrari:
> Thinking more about this, I don't understand why we need a parallel DOCX path for CVs.
> 
> We deliberately created CVs as PDF's because
> 
> * CVs are not normally intended to writeable - they are normally read only PDFs
> * We want to avoid relying on a proprietary format (Microsoft's DOCX).
> 
> I just checked and PDFs can be uploaded to Google Drive, then opened as Google Docs and then translated. I believe that was the plan for the DOCX files - but if you can do the same with PDF's, why duplicate our CV generation to include DOCX?

Thanks for the comment @camerojo . I understand your point, and I agree that PDF should stay as the main/final format for CVs.

I added DOCX not to replace PDF, and not to make the CV editable for employers. The reason is the request from Mathie:

> Some employers in France want to receive the anonymised CV samples that we send them in French.
>
> At the moment, we translate them ourselves manually because if we use software, it changes the layout and makes the CV look less attractive.
>
> Would it be possible to have the opportunity to download the CV in Word or Google Form format rather than PDF? That way, I think it would be easier for us to translate and would save us time.

So the main problem is translation. The team needs to translate CVs into French, but when they use software with PDF files, the layout can change and the CV does not look good.

I also checked the PDF → Google Docs option. It works, because Google Drive can open a PDF as a Google Doc. But the layout is not always safe. Things like horizontal lines or section separators. columns, spacing, lists, and tables can change. CVs depend a lot on layout, so the team may still need to fix the CV manually after translation.

So my thinking was:

* PDF stays as the default and final read-only CV format.
* DOCX is only an optional file for internal translation.
* This helps the France team save time with translation and layout fixing.
* We are not changing the current PDF behaviour.
* We are not asking employers to edit CVs.

I understand your concern about having a second DOCX generation path. We can also make the button clear, for example: “Download Word for translation”, so it is not shown as an equal replacement for PDF.
Also, the idea is to use the DOCX file to create a Google Doc, then translate/edit it there more easily.

camerojo:
For translation purposes I don't think it matters if there are some formatting differences. You can see those formatting differences also when converting from DocX to Google Docs as well. It just depends on the doc you are converting.

The main thing is the translation. They can tidy up the formatting. I think they will have to do that for both PDFs and DOCX's

EhsanEhrari:
@camerojo If that’s the case, I can close this PR and create a new PR under #3398 to upload the generated PDF into Google Drive and open it as a Google Doc for translation purposes. What do you think?

sadatmalik (commented):
@EhsanEhrari before reviewing further, can we confirm and document whether we actually want to support and maintain a parallel DOCX export path rather than solving translation via the existing PDF --> Google Docs workflow.

This has been aired and discussed in engineering calls but the decision is unclear to me.

EhsanEhrari:
> @EhsanEhrari before reviewing further, can we confirm and document whether we actually want to support and maintain a parallel DOCX export path rather than solving translation via the existing PDF --> Google Docs workflow.
> 
> This has been aired and discussed in engineering calls but the decision is unclear to me.

@sadatmalik The decision is unclear to me as well. I think John wants to take one more look at this before we decide. We can discuss it tomorrow and then document the final decision.

EhsanEhrari:
@camerojo waiting for your review and thoughts on this one!

server/src/main/java/org/tctalent/server/service/db/CandidateService.java
```diff
@@ -436,7 +436,15 @@ List<DataRow> computeReferrerStats(Gender gender, String country, LocalDate date
     List<DataRow> computeStatusStats(Gender gender, String country, LocalDate dateFrom, LocalDate dateTo, Set<Long> candidateIds, List<Long> sourceCountryIds);
 
     Resource generateCv(Candidate candidate, Boolean showName, Boolean showContact);
```
camerojo: Can you add similar JavaDoc here

camerojo (commented):
This code is nicely structured and easy to understand - so well done on that.

However, it is more complex than PdfHelper because PDF helper used a different approach, where the basic format of the CV is defined in a Thymeleaf "template" - which is an approach we use through the code for other purposes as well - including formatting emails.

The PdfHelper just runs the template, with the candidate data as input. That generates html/xhtml. Then we use a standard off the shelf utility which converts that xhtml to a PDF.

There are other standard tools for converting xhtml to DOCX. For example, see this from a Google search: https://share.google/aimode/S5ZmpdLmVIwlZ74xd

Pandoc seems to be the most popular option. https://pandoc.org/

The advantage of this approach of using a Thymeleaf template and then off the shelf converters, apart from being simpler is that we have a standard single template for CVs rather than having to maintain two separate formats.

EhsanEhrari:
> This code is nicely structured and easy to understand - so well done on that.
> 
> However, it is more complex than PdfHelper because PDF helper used a different approach, where the basic format of the CV is defined in a Thymeleaf "template" - which is an approach we use through the code for other purposes as well - including formatting emails.
> 
> The PdfHelper just runs the template, with the candidate data as input. That generates html/xhtml. Then we use a standard off the shelf utility which converts that xhtml to a PDF.
> 
> There are other standard tools for converting xhtml to DOCX. For example, see this from a Google search: https://share.google/aimode/S5ZmpdLmVIwlZ74xd
> 
> Pandoc seems to be the most popular option. https://pandoc.org/
> 
> The advantage of this approach of using a Thymeleaf template and then off the shelf converters, apart from being simpler is that we have a standard single template for CVs rather than having to maintain two separate formats.

Thanks for the comment @camerojo, Because TC deploys the Spring Boot server through Jib to ECR, using Pandoc would require adding and maintaining the Pandoc executable inside the runtime container image. There is another option in your  Google search: https://share.google/aimode/S5ZmpdLmVIwlZ74xd which is the **docx4j** implementation that keeps the same Thymeleaf-template approach but avoids deployment changes, because it is just a Java dependency. So I think **docx4j** is the simpler option for this ticket, while Pandoc would be a good future improvement if we decide to standardize external document conversion tooling in our container image.

camerojo:
> Thanks for the comment @camerojo, Because TC deploys the Spring Boot server through Jib to ECR, using Pandoc would require adding and maintaining the Pandoc executable inside the runtime container image. There is another option in your Google search: https://share.google/aimode/S5ZmpdLmVIwlZ74xd which is the **docx4j** implementation that keeps the same Thymeleaf-template approach but avoids deployment changes, because it is just a Java dependency. So I think **docx4j** is the simpler option for this ticket, while Pandoc would be a good future improvement if we decide to standardize external document conversion tooling in our container image.


That sounds a good approach @EhsanEhrari

server/src/main/java/org/tctalent/server/exception/DocxGenerationException.java
```diff
@@ -0,0 +1,25 @@
...
+ * FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License
+ * for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see https://www.gnu.org/licenses/.
+ */
+
+package org.tctalent.server.exception;
+
+public class DocxGenerationException extends RuntimeException {
```
camerojo: This is different to PdfGenerationException which subclasses ServiceException, which is our standard and which we have explicit support for.

Perhaps it would be better anyway to have just one CvGenerationException used by both.

server/src/main/java/org/tctalent/server/service/db/impl/CandidateServiceImpl.java
```diff
@@ -2133,6 +2135,11 @@ public Resource generateCv(Candidate candidate, Boolean showName, Boolean showCo
        return pdfHelper.generatePdf(candidate, showName, showContact);
     }
 
+    @Override
+    public Resource generateCvDocx(Candidate candidate, Boolean showName, Boolean showContact) {
+        return docxHelper.generateDocx(candidate, showName, showContact);
```
camerojo: Instead of adding a whole new duplicate method to the candidate interface, perhaps it would be better to keep the existing generateCv method, passing in an indicator whether a pdf or docx is required.

server/src/main/java/org/tctalent/server/service/db/CandidateService.java
camerojo: See other comment above. These effectively identical methods should probably be replace by one generic one.

Also, if they can throw an exception that should be documented.

server/src/main/java/org/tctalent/server/service/db/util/DocxHelper.java
```diff
@@ -0,0 +1,108 @@
...
+ */
+@Service
+@RequiredArgsConstructor
+@Slf4j
+public class DocxHelper {
+
+  private static final String PDF_RESOURCE_BASE_PATH = "/pdf/";
+
+  private final PdfHelper pdfHelper;
+
```
camerojo, line 50: Instead of including Pdf code inside the Docx helper, just so that you can retrieve the template and xhtml that they both use, it would be better to do a bit of simple refactoring, moving the common template xhtml processing code that they both require to a different package.

ui/admin-portal/src/app/services/candidate.service.ts
```diff
@@ -176,6 +176,23 @@ export class CandidateService implements IntakeService {
       );
   }
 
+  downloadCvDocx(request: DownloadCVRequest) {
+    return this.http.post(
```
camerojo, line 175: Could the code be simplified if we did not have to open up a whole new API for the docx, and just add a new field to DownloadCVRequest indicating whether the user is requesting a PDF or Docx?

camerojo (changes requested):
See comments. I think that a bit of refactoring could mean fewer changes and a simpler architecture.

ui/admin-portal/src/app/components/util/download-cv/download-cv.component.ts
```diff
@@ -44,20 +44,26 @@ export class DownloadCvComponent implements OnInit {
     this.form = this.fb.group({
       name: [false],
       contact: [false],
+      format: ['PDF'],
     });
   }
 
   onSave() {
+    const format = this.form.value.format;
```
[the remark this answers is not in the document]
EhsanEhrari: Fixed it by using the format inside request.

server/src/main/java/org/tctalent/server/service/db/CandidateService.java
```diff
@@ -438,7 +440,16 @@ List<DataRow> computeReferrerStats(Gender gender, String country, LocalDate date
...
+    /**
+     * Generates a candidate CV in the requested format.
+     *
+     * @param candidate the candidate whose CV will be generated
+     * @param showName whether the candidate's name should be included in the generated CV
+     * @param showContact whether the candidate's contact details should be included in the generated CV
+     * @throws CvGenerationException if the CV cannot be generated
+     * @return a {@link Resource} containing the generated CV as a DOCX document
+     */
+    Resource generateCv(Candidate candidate, Boolean showName, Boolean showContact, CvFormat format) throws CvGenerationException;
```
camerojo, line 453: Missing JavaDoc for CvFormat parameter

camerojo (dismissed):
Looks good - approved - but just one missing JavaDoc

server/src/main/java/org/tctalent/server/service/db/CandidateService.java
EhsanEhrari: @camerojo Just pushed the missing JavaDoc, could u please reapprove it thanks!

camerojo (approved):

sadatmalik (approved):

hiba-machfej (approved):

PR #3371, opened 2026-04-23
Remove hardcoded secrets from configs

EhsanEhrari:
Replace hardcoded default secrets in application.yml with environment-variable placeholders for datasource password, JWT secret, and translation password to avoid shipping sensitive values. Add a test-only dummy Base64 JWT secret in application-test.yml to keep tests working without exposing production secrets.

### DIFF
--- a/server/src/main/resources/application.yml
+++ b/server/src/main/resources/application.yml
@@ -107,7 +107,7 @@ spring:
     driverClassName: org.postgresql.Driver
     url: ${SPRING_DATASOURCE_URL:jdbc:postgresql://localhost:5432/tctalent}
     username: tctalent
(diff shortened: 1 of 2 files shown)


hiba-machfej (approved):

camerojo (changes requested):
Do this changes come with an updated Secrets file? If they don't it will break everyone's dev .

EhsanEhrari:
> Do this changes come with an updated Secrets file? If they don't it will break everyone's dev .

Yes @camerojo after approval and before merging I will send the updated secret file to the channel.

camerojo (approved):
Thanks Ehsan. Approving this so that you can get the new secrets out.

PR #3381, opened 2026-04-27
3374 bug where candidates on grn are marked as ineligible at the end of the registration process if they are already located in a destination country

EhsanEhrari:


server/src/main/java/org/tctalent/server/service/db/impl/CandidateServiceImpl.java
```diff
@@ -3041,6 +3057,11 @@ private CandidateStatus checkForAutomaticStatusChanges(
                         //that can also be changed to pending.
                         newStatus = CandidateStatus.pending;
                     }
+                    else if (!tcInstanceService.isTBB()) {
+                        // GRN should not keep candidates ineligible due to the TBB-only
```
camerojo: A bit clearer to use "isGRN()"

I don't think that this logic is correct. I don't think that a change of nationality or country should have ANY effect on a GRN candidate. The way it is now, if a GRN candidate was ineligible for some reason it is possible that they become eligible just because their nationality or country changes. It shouldn't make any difference.

This code could be a lot simpler. It could be exactly as it was - ie no changes. Except that right at the beginning there is the line

If (isGrn())
  return null

camerojo (changes requested):

server/src/test/java/org/tctalent/server/service/db/impl/CandidateServiceImplTest.java
```diff
@@ -327,6 +328,8 @@ private void stubUpdatePersonalToReachReassignPartnerIfNeeded(CandidateStatus ca
     updateCandidatePersonalRequest.setCountryId(1L);
     updateCandidatePersonalRequest.setNationalityId(2L);
     updateCandidatePersonalRequest.setOtherNationalityIds(new Long[0]);
+    
+    given(tcInstanceService.isTBB()).willReturn(true);
```
camerojo, line 331: Can you also create a test which verifies the changes that you have added for the GRN case?

Currently the test only tests that the logic is working correctly for the TBB case.

camerojo (changes requested):

server/src/test/java/org/tctalent/server/service/db/impl/CandidateServiceImplTest.java
EhsanEhrari: Done and pushed the changes.

sadatmalik (approved):

camerojo (approved):

PR #3391, opened 2026-04-30
3348 tsquery injection via simplequerystring ds 002

EhsanEhrari:


camerojo (changes requested):
I am not sure about this whole approach.

It seems like a whole lot of complexity has been added by moving to parameterized queries which I deliberately avoided when initially designing this.

Also, the move to plainto_tsquery removes users' existing ability to use boolean expressions in their text searches. That was deliberately kept in so that users did not lose existing functionality - which they do use.
The buildTsQuerySQL method in CandidateSearchUtils preserves Elasticsearch style boolean logic which our users are used to.  We don't want to just throw away that functionality by moving to plainto_tsquery.

I am not sure what DeepSearch thinks is the main security issue here. It is true that it is possible for users to enter characters which break the syntax - but that was known and accepted. 

https://www.postgresql.org/docs/current/textsearch-controls.html

If there is a syntax error, the user just receives an error message - which is hardly a security issue.

Perhaps all we need to do to make DeepSearch happy is to sanitize the input field.

sadatmalik (approved):

camerojo (approved):

PR #3394, opened 2026-04-30
fix: added @Order(1) annotation to SystemAdminConfiguration startup method

### LINKED ISSUE #3393: Bug: startup failure due to ServiceListSetupService issue
Caroline reported this one. The rest of us didn't catch it presumably because we weren't starting up with a brand new empty DB. I've added her screenshots below but will explain it first (it was quite an easy one for me to figure out, since it was caused by my own omission to check in a crucial code snippet for my [CASI list setup PR](https://github.com/Talent-Catalog/talentcatalog/pull/3261/changes/06d846d6bb7a76977395f97add3c30a2106149b3). 

- Service Lists are created at startup, utilising the `@EventListener(ApplicationReadyEvent.class)` annotation.
- Creating a List requires a user, so the method was first fetching the system admin user.
- For that reason, I also added this to the startup annotation: `@Order(2)` — to ensure that the system admin user (which is also created by a startup method) already exists.
- Somehow I omitted to check in the `@Order(1)` annotation to the system admin creation method, causing the entire sequence to fail for Caroline. (And not for the rest of us, since we already had a system admin user in our dev DBs.)

Apologies all, and particularly Caroline! I'll make a PR now to fix this.

<br><br>

samschlicht:


sadatmalik (approved):

cazcam34 (approved):
Thanks Sam sorry for the delay in approving!

samschlicht:
> Thanks Sam sorry for the delay in approving!

No probs, @cazcam34 ! I've merged it, so you can go ahead with getting your build going. 😅

PR #3396, opened 2026-05-01
Fixes TC-1044 - add audit fields to all candidate provided data tables

sadatmalik:
This PR --
- Implements Spring Data JPA auditing for all candidate-provided data tables

**What was built**

- Introduced `AbstractCandidateDataDomainObject`, a new `@MappedSuperclass` that wires `@EntityListeners(AuditingEntityListener.class)` which declares the four audit fields (`createdBy`, `createdDate`, `updatedBy`, `updatedDate`) as Spring Data `@CreatedBy` / `@CreatedDate` / `@LastModifiedBy` / `@LastModifiedDate` properties
- All candidate-data entities now extend this class: `Candidate`, `CandidateAttachment`, `CandidateCertification`, `CandidateCitizenship`, `CandidateDependant`, `CandidateDestination`, `CandidateEducation`, `CandidateExam`, `CandidateJobExperience`, `CandidateLanguage`, `CandidateNote`, `CandidateOccupation`, `CandidateReviewStatusItem`, and the `@MappedSuperclass` bases `CandidateVisaCheckBase` and `CandidateVisaJobCheckBase` (covering `CandidateVisaCheck`, and `CandidateVisaJobCheck`)
- Added `JpaAuditingConfig` to enable `@EnableJpaAuditing` with a custom `AuditorAware` bean and `OffsetDateTime` provider
- Added Flyway migration `V2_12` to add the four audit columns to the tables that previously had none

**AuditorAware**

- Implemented `SpringSecurityAuditorAware` to supply the current auditor: returns the logged-in user for authenticated requests; falls back to the system admin for anonymous or system-triggered writes
- The fallback avoids loading the full system-admin `User` entity (which would trigger eager association fetches and cause a `JpaSystemException` during flush). Instead, `SystemAdminConfiguration` caches the system admin's ID at startup and the auditor returns a lightweight shell `User` carrying only that ID — sufficient for Hibernate to write the FK

**Service cleanup**

- Removed all manual `setAuditFields(...)` calls from the candidate service implementations now that the listener handles this automatically

**Tests**

- `CandidateDataAuditingIntegrationTest` covers create and update auditing for all entities, for both admin and candidate-user flows


server/src/main/java/org/tctalent/server/security/SpringSecurityAuditorAware.java
```diff
@@ -0,0 +1,85 @@
...
+     */
+    public void setSystemAdminId(Long systemAdminId) {
+        this.systemAdminId = systemAdminId;
+    }
+
+    @NotNull
+    @Override
+    public Optional<User> getCurrentAuditor() {
+        Optional<User> loggedInUser = authService.getLoggedInUser()
+            .filter(user -> user.getId() != null);
```
camerojo: I think it might be better to use UserService.getLoggedInUser(). That will guarantee that the returned User is a live JPA User object. Sometimes that might be important.
JPA can fail sometimes if you add an entity that is not in the current persistence context.

AuthService.getLoggedInUser() has comments suggesting that we should phase it out.

camerojo (changes requested):
Very nice! Just one comment about where you are getting User from.

server/src/main/java/org/tctalent/server/security/SpringSecurityAuditorAware.java
sadatmalik: Good catch John - fixed

EhsanEhrari (dismissed):

camerojo (dismissed):

camerojo:
Has some conflicts @sadatmalik

EhsanEhrari (approved):

camerojo (approved):

PR #3397, opened 2026-05-02
3331 look at and fix version warnings in gradle build file

EhsanEhrari:


server/build.gradle
```diff
@@ -131,41 +131,42 @@ dependencies {
...
-    implementation group: 'org.apache.pdfbox', name: 'pdfbox', version: '2.0.1'
-    implementation group: 'org.apache.poi', name: 'poi', version: '4.1.2'
-    implementation group: 'org.apache.poi', name: 'poi-ooxml', version: '4.1.2'
-    implementation group: 'org.apache.poi', name: 'poi-scratchpad', version: '4.1.2'
-    implementation group: 'org.bouncycastle', name: 'bcprov-jdk16', version: '1.45'
+    implementation group: 'org.apache.pdfbox', name: 'pdfbox', version: '2.0.35'
+    implementation group: 'org.apache.poi', name: 'poi', version: '5.5.1'
+    implementation group: 'org.apache.poi', name: 'poi-ooxml', version: '5.5.1'
+    implementation group: 'org.apache.poi', name: 'poi-scratchpad', version: '5.5.1'
+    implementation group: 'org.bouncycastle', name: 'bcprov-jdk18on', version: '1.84'
```
camerojo, line 144: We are on Java 17 (see our README.md) - so jdk18 might be a problem.

server/build.gradle
```diff
@@ -131,41 +131,42 @@ dependencies {
...
     implementation 'com.google.cloud:google-cloud-storage'
     implementation 'com.google.apis:google-api-services-sheets:v4-rev20210629-1.32.1'
 
-    implementation group: 'org.apache.pdfbox', name: 'pdfbox', version: '2.0.1'
-    implementation group: 'org.apache.poi', name: 'poi', version: '4.1.2'
-    implementation group: 'org.apache.poi', name: 'poi-ooxml', version: '4.1.2'
-    implementation group: 'org.apache.poi', name: 'poi-scratchpad', version: '4.1.2'
-    implementation group: 'org.bouncycastle', name: 'bcprov-jdk16', version: '1.45'
+    implementation group: 'org.apache.pdfbox', name: 'pdfbox', version: '2.0.35'
+    implementation group: 'org.apache.poi', name: 'poi', version: '5.5.1'
```
camerojo, line 139: This is a major release change. Did you check that the code that uses these libraries still works?

See https://poi.apache.org/versioning.html

server/build.gradle
```diff
@@ -131,41 +131,42 @@ dependencies {
...
 // https://mvnrepository.com/artifact/com.slack.api/slack-api-client
-    implementation group: 'com.slack.api', name: 'slack-api-client', version: '1.9.0'
+    implementation group: 'com.slack.api', name: 'slack-api-client', version: '1.48.1'
     implementation("org.jdom:jdom2")
-// https://mvnrepository.com/artifact/mysql/mysql-connector-java
-    implementation group: 'mysql', name: 'mysql-connector-java', version: '5.1.47'
+
+//  Artifact Relocated: mysql » mysql-connector-java → com.mysql » mysql-connector-j
+//  https://mvnrepository.com/artifact/com.mysql/mysql-connector-j
+    implementation group: 'com.mysql', name: 'mysql-connector-j', version: '9.7.0'
```
camerojo: Are we still using MySQL? It would be good to remove this if we aren't.

camerojo (changes requested):

server/build.gradle
EhsanEhrari: Sorry I had to put a comment there, the artifact name is really confusing. `bcprov-jdk18on` does not mean “Java 18+”; Bouncy Castle’s official download page lists `bcprov-jdk18on-1.84.jar` under “**JDK 1.8 and later**” and says it is compiled for Java **1.8+**. So Java **17** is within range.

server/build.gradle
EhsanEhrari: For PDFBox 2.0.35, the tests pass using both PDFParser/PDFTextStripper and PDDocument.load(...) text extraction.

For POI 5.5.1, the DOCX extraction tests pass for both English and Arabic files using XWPFDocument/XWPFWordExtractor.

I agree POI 4.1.2 -> 5.5.1 is a major version upgrade with those tests passing, I think we can keep the upgrade.

server/build.gradle
```diff
@@ -131,41 +131,40 @@ dependencies {
...
+    implementation("com.opencsv:opencsv:5.8")
 
     // https://github.com/lukas-krecan/ShedLock
     implementation "net.javacrumbs.shedlock:shedlock-spring:5.10.0"
     implementation "net.javacrumbs.shedlock:shedlock-provider-jdbc-template:5.10.0"
 
-    implementation 'com.google.guava:guava:31.1-jre'
+    implementation 'com.google.guava:guava:33.5.0-jre'
     implementation("io.jsonwebtoken:jjwt-api:0.11.2")
     implementation 'com.google.api-client:google-api-client:1.32.1'
```
camerojo, line 160: This is a major version change.  I am not sure that we are still using Guava. Remove if we are not.

camerojo: @EhsanEhrari Can you check if we are still using Guava? I don't think we are.

camerojo (changes requested):

server/build.gradle
EhsanEhrari: You are right @camerojo I couldn't find any usage of Guava, I will remove it.

server/build.gradle
EhsanEhrari: @camerojo 
I tested the UI behavior for the PDFBox and Apache POI upgrade. These libraries are used when uploading CV attachments, because the system extracts text from PDF, DOCX, and DOC files for keyword search.

I tested uploading PDF, DOCX, and DOC CV files from the Admin Portal candidate Attachments section. The files uploaded successfully, appeared in All Attachments, showed as CV attachments, and downloaded correctly.

I also tested the Candidate Portal upload flow from Profile → Upload → Edit. The files uploaded and appeared correctly.

I checked the important behavior by searching unique words from the uploaded CV files. The candidate appeared in search results, so text extraction is working after the dependency upgrade.

server/build.gradle
EhsanEhrari: I also checked the Bouncy Castle upgrade by testing the security-related UI flows. I tested admin login, candidate login, wrong password behavior, password reset, and new candidate registration/login. These flows worked correctly after the upgrade. I also checked page refresh and navigation after login to make sure JWT/session behavior still works. I did not see any UI error or backend crypto/authentication error. @sadatmalik @camerojo

sadatmalik (approved):
Well done Ehsan!

camerojo (approved):

PR #3402, opened 2026-05-04
Add aspirations field to candidate profiles

EhsanEhrari:
Introduce an aspirations field across backend and UI. Added aspirations to CandidateReadDto and Candidate model, and fixed CandidateBuilderSelector to include "aspirations" instead of a duplicate additionalInfo. Updated admin UI to display/edit aspirations (templates + form control) and gated visibility with AuthenticationService.isGrnInstance(); injected the service where needed. Also exposed aspirations in the candidate portal profile view.


server/src/main/java/org/tctalent/server/api/dto/CandidateBuilderSelector.java
```diff
@@ -241,7 +241,7 @@ private DtoBuilder candidateDto(
                     .add("shareableNotes")
                     .add("additionalInfo")
                     .add("candidateMessage")
-                    .add("additionalInfo")
+                    .add("aspirations")
                 ;
```
camerojo, line 245: Why remove additionalInfo?

ui/admin-portal/src/app/components/candidates/view/additional-info/view-candidate-additional-info.component.ts
```diff
@@ -52,4 +54,7 @@ export class ViewCandidateAdditionalInfoComponent implements OnInit {
 
   }
 
+  showAspirations(): boolean {
+     return this.authenticationService.isGrnInstance();
+  }
 }
```
camerojo, line 55: I think it might be better to pass showInspirations in as an @Input(). That removes the dependency on AuthenticationService resulting in a more loosely coupled component.

ui/candidate-portal/src/app/components/profile/view/tab/profile/candidate-profile.component.ts
```diff
@@ -189,6 +193,10 @@ export class CandidateProfileComponent implements OnInit {
     return this.countries?.find(c => c.id === country?.id)?.name;
   }
 
+  showAspirations(): boolean {
+    return this.authenticationService.isGrnInstance();
+  }
+
```
camerojo: See above comment - probably better to make this more loosely coupled by having a @Input for showAspirations

camerojo (changes requested):

server/src/main/java/org/tctalent/server/api/dto/CandidateBuilderSelector.java
EhsanEhrari: We already have ("additionalInfo") in the dto just above candidateMessage, I don't know maybe the second additionalInfo was added mistakenly so I changed it to aspirations

ui/admin-portal/src/app/components/candidates/view/additional-info/edit/edit-candidate-additional-info.component.ts
```diff
@@ -73,4 +76,8 @@ export class EditCandidateAdditionalInfoComponent implements OnInit {
   dismiss() {
     this.activeModal.dismiss(false);
   }
+
+  showAspirations(): boolean {
+    return this.authenticationService.isGrnInstance();
```
sadatmalik: Small consistency thought: since ViewCandidateAdditionalInfoComponent opens this modal and already has showAspirations, could you pass it through as an @Input() here as well instead of calling auth directly?

That would match the pattern you've used elsewhere in this PR.

EhsanEhrari: Nice point, thanks! Pushed the new changes. @sadatmalik

sadatmalik (changes requested):
Requesting changes — there’s a gap in the admin-portal flow.

End-to-end the candidate portal works correctly for aspirations: a candidate can both view and edit their aspirations, and the value is persisted to the DB.

On the admin portal, however, only the read path is in place. Admin users can see a candidate’s aspirations on the Additional Info tab, but editing the value in the Additional Info modal does not persist to the DB.

Could you please close the write path on the admin side so that admins can view but also update the candidate aspirations.

EhsanEhrari:
> Requesting changes — there’s a gap in the admin-portal flow.
> 
> End-to-end the candidate portal works correctly for aspirations: a candidate can both view and edit their aspirations, and the value is persisted to the DB.
> 
> On the admin portal, however, only the read path is in place. Admin users can see a candidate’s aspirations on the Additional Info tab, but editing the value in the Additional Info modal does not persist to the DB.
> 
> Could you please close the write path on the admin side so that admins can view but also update the candidate aspirations.

Thanks @sadatmalik 
I’m not sure why I didn’t notice this earlier, even though I tested the flow multiple times. I think I missed it because aspirations were displayed under the Additional Info component, and when testing the edit flow I was actually editing Additional Info, not Aspirations.

I’ve now separated this into dedicated view and edit Aspirations components, so admins can view and update candidate aspirations properly from the admin portal.

sadatmalik (approved):

camerojo (approved):

PR #3411, opened 2026-05-06
1093 when candidate data is enhanced by partners the original candidate entered data should be retained

### LINKED ISSUE #1093: When candidate data is enhanced by partners, the original candidate entered data should be retained
As a partner working to improve data about a candidate
I need to keep the original data entered by the candidate
to avoid making mistakes about what the candidate has actually reported to us.

Raised by Mariam in UX call

When doing searches and stats both the original and enhanced data should be considered.

Currently admin staff are hesitant to update a candidate's profile directly because they don't feel comfortable overwriting the candidate's words. Instead they add candidate notes - but that data is less structured than the data in the profile. For example, our existing automated CV generation works from the profile data - not the candidate notes.

camerojo:


sadatmalik (dismissed):
Neat implementation - admins will love this!

PR #3416, opened 2026-05-07
Added upper-bound and lower-bound validation page size

cazcam34:
After looking at this fix I was curious to try Junie on Intellij, and it included my fix (just adding max size to the page size as per the issue) but also added lower bound validation with min sizes, and put the default size (25) as a variable. This has been tested and reviewed by myself and I think although it's a little more through than the Deep Scan suggested it is all useful validation that makes sense to add.


cazcam34:
For some reason this PR is opened but it isn't appearing in the list of pull requests (/pulls)... I have tested repushing a commit with a small change but it still isn't appearing. Could be github side issue?

camerojo (approved):

PR #3417, opened 2026-05-07
3351 pagination dos no maximum page size ds 008

cazcam34:
Second attempt at pull request as the first PR weirdly wasn't showing up in the list of PRs.
After looking at this fix I was curious to try Junie on Intellij, and it included my fix (just adding max size to the page size as per the issue) but also added lower bound validation with min sizes, and put the default size (25) as a variable. This has been tested and reviewed by myself and I think although it's a little more through than the Deep Scan suggested it is all useful validation that makes sense to add.

### DIFF
--- a/server/src/main/java/org/tctalent/server/request/PagedSearchRequest.java
+++ b/server/src/main/java/org/tctalent/server/request/PagedSearchRequest.java
@@ -16,6 +16,8 @@
 
 package org.tctalent.server.request;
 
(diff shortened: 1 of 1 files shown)


cazcam34:
Was a github issue why the PR wasn't appearing and resolved by itself, so this is good for review

camerojo (approved):

EhsanEhrari (approved):

PR #3446, opened 2026-05-14
Closes TC-1220 3385 linear migration

sadatmalik:
This PR --

- documents a record of system decisions made during the migration from ZenHub to Linear
- it is a living record - future decisions about Linear use should be documented here for clarity and ease of reference

### DIFF
--- a/docs/decisions/zenhub-to-linear-migration.md
+++ b/docs/decisions/zenhub-to-linear-migration.md
@@ -236,3 +236,34 @@ Adopt both layers immediately. Use Initiatives to express longer-term strategic 
 _To be agreed._
 
 **Recommendation:** Option B. Improved roadmap capability is a primary driver for the migration, and Initiatives are the layer that gives Linear's roadmap its strategic structure. Deferring them would result in a flat roadmap that fails to deliver one of the main benefits of moving away from ZenHub. Starting with a small, well-chosen set of Initiatives gives product management a meaningful roadmap from day one.
+
+---
+
+## Decision 9: Access Control and Role Assignment
+
+**Status:** Pending
+
+### Context
+
+Linear has three role levels: **Admin**, **Member**, and **Guest**. Roles must be assigned to everyone who needs access to the workspace. Two questions of principle need to be agreed:
+
+1. How permissively or strictly to assign roles within the core team.
+2. Whether to invite external open-source contributors to Linear (e.g. as Guests).
+
+### Decision
+
+**Role assignment principles**
+
+| Role | Assigned to | Rationale |
+|---|---|---|
+| Admin | Core maintainers (a small subset of the team) | Manage workspace settings, integrations, members, and teams. |
+| Member | All core developers with regular commit activity | Full ability to create and manage issues, participate in cycles, update states. |
+| Guest | Product or non-technical stakeholders requiring visibility but not full participation | Read + comment access on specific Projects or teams. |
+
+Named role assignments are operational and will be maintained outside this record.
+
+**External open-source contributors**
+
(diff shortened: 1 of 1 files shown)


docs/decisions/zenhub-to-linear-migration.md
```diff
@@ -0,0 +1,329 @@
...
+One workspace, one team (`Talent Catalog`). All issues share a single prefix. Simpler day-to-day; product planning is scoped across the whole project.
+
+**Option B — Multiple teams**  
+One workspace, multiple teams (e.g. `Backend`, `Frontend`, `Platform`). Each team has its own prefix, cycles, and roadmap. Allows independent planning per stream at the cost of added structural complexity.
+
+### Decision
+
+_To be agreed._
+
+**Recommendation:** Start with Option A. Split into multiple teams later if product management identifies a clear need to plan streams independently.
```
camerojo, line 111: This sound very sensible to me

docs/decisions/zenhub-to-linear-migration.md
```diff
@@ -0,0 +1,329 @@
...
+
+Priorities are not in scope here. Linear's built-in priority system (`Urgent` / `High` / `Medium` / `Low` / `No Priority`) is adopted as-is; there is no meaningful alternative.
+
+### Options
+
+**Option A — Adopt the existing labels as-is**  
+Enable sync without modification. The current GitHub label set becomes the Linear label set. No clean-up effort, but inherits existing label drift and duplication.
+
+**Option B — Tidy up before sync**  
+Audit and consolidate the existing GitHub labels before enabling Linear sync: retire stale labels, merge overlapping ones, rename inconsistent ones. The cleaner set then propagates into Linear.
```
camerojo, line 167: As long as it is not too much work, this looks like the way to go.

docs/decisions/zenhub-to-linear-migration.md
```diff
@@ -0,0 +1,329 @@
...
+- _Service partnership ecosystem_
+- _Global scale and resilience_
+- _Trust, privacy and consent_
+- _Candidate matching intelligence_
+
+### Decision
+
+_To be agreed._
+
+**Recommendation:** Option B. Improved roadmap capability is a primary driver for the migration, and Initiatives are the layer that gives Linear's roadmap its strategic structure. Deferring them would result in a flat roadmap that fails to deliver one of the main benefits of moving away from ZenHub. Starting with a small, well-chosen set of Initiatives gives product management a meaningful roadmap from day one.
```
camerojo, line 242: Agree with this recommendation

docs/decisions/zenhub-to-linear-migration.md
```diff
@@ -0,0 +1,329 @@
...
+
+Plan-dependent availability is important:
+
+- **Guest role requires the Business plan ($16/user/month) or Enterprise.** It is not available on Free or Basic.
+- On Basic, everyone with access is a Member (or Admin) — there is no read-only / scoped option for stakeholders.
+
+Two questions of principle need to be agreed:
+
+1. How permissively or strictly to assign roles within the core team.
+2. Whether the need for Guest access for non-technical stakeholders is significant enough to require the Business plan.
```
camerojo, line 262: I think Guest access will be important - particularly for TBB and UNHCR. 

We are not anticipating a very large number of users - especially given that external contributors are not members.

For example, we may have 10 OPC users plus TBB, UNHCR and maybe a couple more guest users. That is less than US$3,000/year which I think is reasonable for what we are getting in return.

camerojo (approved):
A wonderful structured way to record these decisions - and potential future decisions.

A great model, as you suggest for decision making on other complex issues.

docs/decisions/zenhub-to-linear-migration.md
sadatmalik: Linear have agreed a 75% discount for OPC - so it will now be US$48 per user per year

PR #3522, opened 2026-06-12
Sadat/tc 1235 increase node version to 20

### LINKED ISSUE #3425: Increase Node version to 20
We are currently on Angular 17 - which supports Node 20.9+. See [https://angular.dev/reference/versions](<https://angular.dev/reference/versions>)

We currently specify Node version 18 (in the main README.md) but Node 18 is no longer supported - see [https://nodejs.org/en/about/previous-releases](<https://nodejs.org/en/about/previous-releases>)

Node 20 has now been chosen for LTS (Long Term Support).

When we move to Angular 18/19 we can go to Node 22 (LTS)<br>When we move to Angular 20+ we can go to Node 24 (LTS)

### Notes

* Note that .nvmrc file in talent catalog directory selects the correct Node version. That file is not checked into GitHub. Maybe it should be?
* README currently says we are on Angular 18 - but the package.json specifies 17

sadatmalik:
Small PR that updates the Node version from 18 to 20.


README.md
camerojo: I think that this also might need to be updated to @17?

camerojo (changes requested):

README.md
sadatmalik: Yes - nice catch - done!

camerojo (approved):

PR #3523, opened 2026-06-15
Fix bug where html stored as part of a json string was not being prop…

### LINKED ISSUE #3520: Fix job experience display and editing issues
@John Cameron said in Talent-Catalog/talentcatalog#3515:

> ## Job Experience Text Preserve:
>
> * The new Original title appears close to the job title, could be confusing and look like Original is part of the job title? Perhaps make slightly smaller, even muted colour a little?
>
> 
>
> * After editing the job experience and adding some tidied text, my edited job moves to the bottom of the card. It was at the top before it was edited. Can be confusing as when modal closes my job has moved and if there's a few jobs in the card it can look like it disappeared until you scroll down. Maybe need to keep the same ordering once saved?
> * I edited another job (after editing about three prior in the same occupation card) and added tidied text (Boo) and when saved it appeared like this…
>
> 
>
> And when I click edit on the same description the tidied text appears empty now.
>
> 
>
> Seems like the tidied text is getting saved to the original and applying some other text/formatting. This occurred to two job experiences on the same candidate. See candidate here:[https://tctalent-test.org/admin-portal/candidate/187641](<https://tctalent-test.org/admin-portal/candidate/187641>). I tried doing the same to other candidates (adding Boo to tidied text) and it worked fine, have not been able to replicate!!
>
> Went back to the candidate 187641 and added tidied text (boo) again to one of the failed ones and it's something about these particular job experiences that is causing the tidied text to just get added to the original and not the tidied text.
>
> 
>
> ##

camerojo:
…erly processed by our HtmlSantizer - messing up the json encoding.

The HtmlSanitizer now behaves differently if it detects a TextParts json string. It now extracts the html parts of the TextParts object, sanitizes them, and returns a new TextParts object containing the sanitized html.

TextPartsCodec has been converted to use static methods which allows it be used from the static HtmlSanitizer methods.

### DIFF
--- a/server/src/main/java/org/tctalent/server/util/text/TextPartsCodec.java
+++ b/server/src/main/java/org/tctalent/server/util/text/TextPartsCodec.java
@@ -22,8 +22,8 @@ import java.util.ArrayList;
 import java.util.List;
 import lombok.Data;
 import lombok.extern.slf4j.Slf4j;
-import org.springframework.stereotype.Component;
-
+import org.springframework.lang.NonNull;
+import org.springframework.lang.Nullable;
 /**
  * Codec (Code/Decode) for TextParts.
  * <p>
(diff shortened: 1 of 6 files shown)


EhsanEhrari (dismissed):

hiba-machfej (dismissed):

server/src/main/java/org/tctalent/server/util/text/TextPartsCodec.java
```diff
@@ -32,54 +32,88 @@
...
+ * </p>
+ *
  * @author John Cameron
  */
-@Component
 @Slf4j
 public class TextPartsCodec {
 
-    private final ObjectMapper objectMapper;
+    private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
```
samschlicht, line 44: @camerojo a note from Claude that I pursued long enough to see that it's only an issue in a very hypothetical edge case: 'the hand-rolled mapper is stricter than the injected one; consider disable(FAIL_ON_UNKNOWN_PROPERTIES) for parity'. 

More pertinent because the callers of `readJson` now swallow any exception and would fail silently, returning the raw JSON blob as the field value.

samschlicht (dismissed):
Approved with 1 non-blocking comment

server/src/main/java/org/tctalent/server/util/text/TextPartsCodec.java
samschlicht: Another option might be to log the failure in both cases.

camerojo: @samschlicht  Thanks for the thoughtful feedback but that is deliberate. I don't want the injected one. The mapper only has one job to do and that is to detect a well formed TextParts. If it is not well formed in any way, it should be ignored - which is what happens - and the text is processed just as text. That is the correct behaviour. 
Because TextParts are automatically created by the Angular code (not by the user), I can't see any way that anything but a correct TextParts will be encountered. But if a incorrectly formatted TextParts was received, it should be treated just as text. For example, if the json included an unknown property, that would be very strange and I would not want to process it - it should fail.

Note also that throwing a "Not Json" Exception is also a normal case - where there is no TextParts - just the original HTML text and nothing else. We wouldn't want to be logging an Exception in that case.

sadatmalik (dismissed):

server/src/main/java/org/tctalent/server/util/text/TextPartsCodec.java
samschlicht: Hey @camerojo yep I suppose where I saw a valid concern was that a dev could edit the frontend to cause the TextParts to be improperly formatted, and this code would fail silently because the `readJson()` exception is swallowed by the callers `TextPartsCodec#read()` and `HtmlSanitizer#sanitize()`. It's an unlikely scenario, I agree.

PR #3540, opened 2026-06-18
Ehsan/tc 1317 cv export formatting

EhsanEhrari:


camerojo (approved):

camerojo:
@sadatmalik This is quite a big change - a lot of lines of code. But it all only affects the creation of the DOCX CV's so if there any problems it will only affect that, nothing else. And the current DOCX formatting is not great.

So I am inclined to put it in. What do you think?

fyi @EhsanEhrari

sadatmalik:
Agreed @camerojo - I have also approved.

@hiba-machfej — this would be a good area for you to test today

sadatmalik (approved):

PR #3568, opened 2026-06-24
TC-1339: Align google doc download cv question mark icon with text

### LINKED ISSUE #3563: Align question mark with text
@Hiba said in Talent-Catalog/talentcatalog#3515:

> 
>
> Very small UI issue the question mark should be aligned with the text.

hiba-machfej:
Screenshot 2026-06-24 at 3 53 15 PM

### DIFF
--- a/ui/admin-portal/src/app/components/util/download-cv/download-cv.component.html
+++ b/ui/admin-portal/src/app/components/util/download-cv/download-cv.component.html
@@ -61,7 +61,6 @@
 
         <button type="button"
                 class="btn btn-link p-0 ms-2 google-doc-help"
-                style="position: relative; top: -0.25rem;"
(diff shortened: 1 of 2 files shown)


EhsanEhrari (changes requested):
Non-blocking: I think it would be better to move this inline style into the existing `google-doc-help` class, since this is a static style.

Inline style is okay for small quick changes, but keeping this in the component stylesheet is easier to maintain.

Also, [Angular Docs](https://angular.dev/guide/components/styling) usually recommends keeping component-specific styles in the component stylesheet. For this fixed style, using the CSS class looks cleaner.

EhsanEhrari (approved):

sadatmalik (approved):

PR #3570, opened 2026-06-24
TC-1336 Add Mark All as Yes button for destinations step in candidate portal registration

### LINKED ISSUE #3560: Add button the marks all inputs yes
@Hiba said in Talent-Catalog/talentcatalog#3515:

> When creating a new candidate profile, there are around 11 countries to select from, and the candidate needs to go through each one individually. Since most candidates use mobile, this can mean a long scroll and many clicks, as they need to select “Yes” or “No” for every country.
>
> We already have a ticket for improving this Talent-Catalog/talentcatalog#3373 , but as a starting point, maybe all countries could have “Yes” selected by default. Then the candidate would only need to scroll and change the specific countries where the answer is “No.” This would reduce the number of clicks and make the process easier for mobile users.
>
>

hiba-machfej:
Adds a bulk action to the candidate registration Destinations step so candidates can mark all destination interest fields as `Yes` with one click. 
Screenshot 2026-06-24 at 7 22 07 PM

### DIFF
--- a/ui/candidate-portal/src/app/components/register/destinations/registration-destinations.component.html
+++ b/ui/candidate-portal/src/app/components/register/destinations/registration-destinations.component.html
@@ -22,6 +22,14 @@
 
     <app-error [error]="error"></app-error>
 
+    <div class="destination-actions">
+      <tc-button type="outline"
+                 [disabled]="saving"
+                 (onClick)="setAllDestinationsToInterested()">
+        Mark all as Yes
+      </tc-button>
+    </div>
+
     <div class="destination-card" *ngFor="let destination of destinations; let i = index;">
       <div class="card">
         <app-destination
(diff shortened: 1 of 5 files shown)


ui/candidate-portal/src/app/components/register/destinations/registration-destinations.component.html
```diff
@@ -22,6 +22,14 @@
 
     <app-error [error]="error"></app-error>
 
+    <div class="destination-actions">
+      <tc-button type="outline"
+                 [disabled]="saving"
+                 (onClick)="setAllDestinationsToInterested()">
+        Mark all as Yes
```
EhsanEhrari: should "Mark all as Yes" use a translation key here, same as other candidate portal labels?

camerojo (approved):
Looks good

EhsanEhrari (approved):

PR #3571, opened 2026-06-24
Hiba/tc 1338 remove empty description title original in job experience

### LINKED ISSUE #3562: Fix empty description validation and UI display issues
@Hiba said in :

> 
>
> If the tidied description is empty, the tidied title disappears. However, if the original description is empty, the title still appears and only says “**Original**” without any text underneath. This might be a bit confusing.
>
> [Talent-Catalog/talentcatalog#3631](<https://github.com/Talent-Catalog/talentcatalog/issues/3631>)
>
> The same thing happens when I create a candidate from the registration page. It says that the description is mandatory, but after I **write something and then delete it**, I can still save the form with an empty field. (I am wondering, now that we have the Tidied field, do admin portal users still need to edit the candidate’s original text? Or should only the Tidied field be editable? )
>
> possible solution:
>
> * Only show the “Tidied” or “Original” title when there is text in that section, so empty titles do not appear on their own.
> * Make the Original description field required in both the edit and registration forms. The star sign should be next to the Original field, and users should not be able to save if it is empty.
>
> 
>
>

hiba-machfej:
This PR fixes the Candidate Portal job experience registration form so the description field is truly mandatory. 

Previously, the field was marked as required, but a user could type a description, delete it, and still save the form. 

he validation now checks the actual visible description text inside `parts.original`, after removing HTML tags and non-breaking spaces. If no real text remains, the field is treated as empty and the user cannot save the form.
Legacy plain-text descriptions are still supported for older existing data created before the TextParts format was introduced.

### DIFF
--- a/ui/candidate-portal/src/app/components/common/candidate-job-experience-form/candidate-job-experience-form.component.ts
+++ b/ui/candidate-portal/src/app/components/common/candidate-job-experience-form/candidate-job-experience-form.component.ts
@@ -25,7 +25,13 @@ import {
   ViewChild
 } from '@angular/core';
 import {CandidateJobExperience} from "../../../model/candidate-job-experience";
-import {UntypedFormBuilder, UntypedFormGroup, Validators} from "@angular/forms";
+import {
+  AbstractControl,
+  UntypedFormBuilder,
+  UntypedFormGroup,
+  ValidationErrors,
+  Validators
+} from "@angular/forms";
 import {Country} from "../../../model/country";
 import {CountryService} from "../../../services/country.service";
 import {CandidateOccupation} from "../../../model/candidate-occupation";
@@ -98,7 +104,10 @@ export class CandidateJobExperienceFormComponent implements OnInit, AfterViewIni
       endDate: [this.candidateJobExperience ? this.candidateJobExperience.endDate : null],
       fullTime: [this.candidateJobExperience ? this.candidateJobExperience.fullTime : null, Validators.required],
       paid: [this.candidateJobExperience ? this.candidateJobExperience.paid : null, Validators.required],
-      description: [this.candidateJobExperience ? this.candidateJobExperience.description : '', Validators.required]
+      description: [
+        this.candidateJobExperience ? this.candidateJobExperience.description : '',
+        this.textPartsDescriptionRequired
+      ]
     }, {validator: this.startDateBeforeEndDate('startDate', 'endDate')});
 
     this.form.controls['paid'].valueChanges.subscribe(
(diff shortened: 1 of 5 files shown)


EhsanEhrari (dismissed):

sadatmalik (commented):
I'll just pause on approving this - it does the job of not displaying the title if the corresponding field is empty.

However before approving/merging, the original issue raised some product questions (copied below) - let's clarify in the engineering meeting if removing the titles is the preferred approach; or making the fields required; or possibly both.

From the issue:

> (I am wondering, now that we have the Tidied field, do admin portal users still need to edit the candidate’s original text? Or should only the Tidied field be editable? )
>
> possible solution:
>
> * Only show the “Tidied” or “Original” title when there is text in that section, so empty titles do not appear on their own.
> * Make the Original description field required in both the edit and registration forms. The star sign should be next to the Original field, and users should not be able to save if it is empty.

camerojo:
As Sadat says, let's chat about this. I think that Original can be blank. What if admin adds text (maybe after talking with candidate). Then it is valid that Original is blank (and I would argue that it is clearer if it has a label, even if it is blank. And Tidied is present. I think that they way the code was, was probably correct. I would be inclined not leave the Original title there, even if the content is blank.

ui/candidate-portal/src/app/components/common/candidate-job-experience-form/candidate-job-experience-form.component.ts
```diff
@@ -169,4 +178,28 @@ export class CandidateJobExperienceFormComponent implements OnInit, AfterViewIni
...
+    try {
+      const parsed = JSON.parse(value);
+      if (typeof parsed?.parts?.original === 'string') {
+        description = parsed.parts.original;
+      }
+    } catch {
+      // Legacy plain text descriptions are valid input, so validate the raw value.
+    }
+
+    const visibleText = description
```
camerojo, line 197:
```suggestion
    const visibleText = description
    // Replace HTML tags such as <p>, </div>, <br> with a normal space.
    .replace(/<[^>]*>/g, ' ')

    // Replace HTML non-breaking spaces with a normal space.
    // Handles both &nbsp; and &#160;, case-insensitively.
    .replace(/&nbsp;|&#160;/gi, ' ')

    // Remove whitespace from the start and end.
    .trim();
```

This is a bit cryptic. It could use some comments - like the above.

camerojo (changes requested):

ui/candidate-portal/src/app/components/common/candidate-job-experience-form/candidate-job-experience-form.component.ts
camerojo: Nice work though Hiba!

camerojo (approved):

EhsanEhrari (approved):

Issue #3578, opened 2026-06-30
Epic 1: Candidate-portal QR scanning


sadatmalik:
**Description:** Add client-side QR scanning to the candidate portal so a refugee can scan their Verify+ card on their own device. Decoding happens locally; only the resulting payload string is sent to the backend.

- [X] Add `@zxing/ngx-scanner@^17` + `@zxing/library` and register the module (no Angular upgrade — portal is Angular 17)
- [X] Build scan component + route (back camera default, manual switch, permission/error states)
- [X] Confirm-before-submit screen showing the parsed UNHCR ID
- [X] `verify-plus.service.ts` to POST the raw, untouched payload to the portal endpoint
- [X] i18n keys (ngx-translate) + accessibility
- [X] Component & service unit specs

**Dependencies:** Epic 0 (payload contract)  <br>  <br>**Done when:** a card can be scanned on a mobile device and its payload posted to the backend.

PR #3592, opened 2026-07-05
Sadat/tc 1355 feature verify 11 candidate portal qr scanner component

### LINKED ISSUE #3588: [Feature] Verify+ 1.1 — Candidate-portal QR scanner component + Services card
## Overview

Introduce QR scanning into the candidate portal for UNHCR Verify+ cards. This first  <br>sub-issue delivers a reusable scanner component and exposes it as a flag-gated card in the  <br>CASI "Services" tab. Decoding happens on-device; no backend integration in this PR.

## Problem / Opportunity

Refugees issued a UNHCR Verify+ card have their UNHCR ID encoded in a QR code. We want to  <br>capture that data in GRN to link GRN accounts to UNHCR account IDs. The foundational step is a  <br>reliable, reusable in-browser scanner. Building it as a shared component now lets both the  <br>Services tab and (later) the registration flow consume the same scanning code.

## Proposed Solution

* Add `@zxing/ngx-scanner@^17` + `@zxing/library` and register the scanner module in  <br>`app.module.ts` (candidate-portal is Angular 17 — no framework upgrade required).
* Build a reusable, presentational scanner component (e.g. `app-verify-plus-scanner`) that  <br>opens the camera (back camera by default), decodes `QR_CODE`, and emits the decoded  <br>payload via `@Output()`. Handles permission-denied, no-camera, and invalid-scan states.  <br>No backend calls.
* Add `ServiceProvider.VERIFY_PLUS` to `model/services.ts`; add a Services-tab card and a  <br>container component `app-verify-plus` (mirroring `app-unhcr`: `@Input() candidate`,  <br>`@Output() backButtonClicked`).
* Gate the card behind `isLocalEnv()` / a feature flag, following the existing Reference  <br>service pattern, so it can merge before backend and approvals are ready.
* For this PR, the container simply displays the decoded string.

## Acceptance Criteria

- [X] `@zxing/ngx-scanner` + `@zxing/library` added and the module registered
- [X] Reusable scanner component decodes a QR code and emits the payload via `@Output()`
- [X] Camera permission-denied / no-camera / invalid-scan states handled gracefully
- [X] Verify+ card appears in the Services tab only when the flag / `isLocalEnv()` is on
- [X] Selecting the card opens the camera and displays the decoded string
- [X] Component unit specs pass (scanner mocked)

## Sign-off Criteria

- [X] Release notes updated
- [X] Release planning updated
- [X] Users notified
- [X] Documentation updated

sadatmalik:
This PR --

- Introduces a reusable in-browser QR scanner using @zxing/ngx-scanner
- Adds a VERIFY_PLUS CASI Services card and app-verify-plus container that opens the scanner and displays the decoded payload (on-device only; no backend in this PR)
- Gates the Verify+ service to GRN instances (hidden on TBB)
- Adds/updates unit specs for the scanner, container, and services components.

### DIFF
--- a/ui/candidate-portal/src/app/components/common/verify-plus-scanner/verify-plus-scanner.component.html
+++ b/ui/candidate-portal/src/app/components/common/verify-plus-scanner/verify-plus-scanner.component.html
@@ -1,0 +1,54 @@
+<!--
+  ~ Copyright (c) 2026 Talent Catalog.
+  ~
+  ~ This program is free software: you can redistribute it and/or modify it under
+  ~ the terms of the GNU General Public License as published by the Free
+  ~ Software Foundation, either version 3 of the License, or any later version.
+  ~
+  ~ This program is distributed in the hope that it will be useful, but WITHOUT
+  ~ ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+  ~ FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License
+  ~ for more details.
+  ~
+  ~ You should have received a copy of the GNU General Public License
+  ~ along with this program. If not, see https://www.gnu.org/licenses/.
+  -->
+
+<div class="verify-plus-scanner">
+  <div class="scanner-actions" *ngIf="!scanning && cameraPermission !== false && hasDevices !== false">
+    <tc-button [type]="'outline'" (onClick)="startScanning()">
+      Enable camera
+    </tc-button>
+  </div>
+
+  <div class="scanner-actions" *ngIf="cameraPermission === false && hasDevices !== false">
+    <p class="scanner-message scanner-message-error">
+    Camera access was denied. Please enable camera permission and try again.
+    </p>
+    <tc-button [type]="'outline'" (onClick)="startScanning()">
+      Try again
+    </tc-button>
+  </div>
+
+  <p class="scanner-message scanner-message-error" *ngIf="hasDevices === false">
+    No camera was detected on this device.
+  </p>
+
+  <zxing-scanner
+    *ngIf="hasDevices !== false"
+    [autostart]="false"
+    [enable]="scanning"
+    [formats]="formats"
+    [device]="selectedDevice"
+    (camerasFound)="onCamerasFound($event)"
+    (camerasNotFound)="onCamerasNotFound()"
+    (permissionResponse)="onPermissionResponse($event)"
+    (scanSuccess)="onScanSuccess($event)"
+    (scanFailure)="onScanFailure()"
+    (scanError)="onScanError($event)"
+  ></zxing-scanner>
+
+  <p class="scanner-message" *ngIf="scanning && invalidScan && cameraPermission !== false && hasDevices !== false">
+    A QR code has not been detected yet. Try moving closer and improving lighting.
+  </p>
+</div>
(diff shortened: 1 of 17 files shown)


ui/candidate-portal/src/app/components/common/verify-plus-scanner/verify-plus-scanner.component.ts
```diff
@@ -0,0 +1,83 @@
...
+      this.scanning = false;
+    }
+  }
+
+  onScanSuccess(decodedValue: string) {
+    if (!decodedValue) {
+      this.invalidScan = true;
+      return;
+    }
+
```
EhsanEhrari, line 75: Right now `onScanSuccess(`) emits the decoded value but keeps `scanning = true`, so the camera can remain active and may repeatedly emit the same QR payload while the code stays in view.
For better for UX, privacy, battery usage, and avoiding duplicate scan events, isn't it better to make `scanning=false` inside `onScanSuccess()`?

ui/candidate-portal/src/app/components/common/verify-plus-scanner/verify-plus-scanner.component.html
```diff
@@ -0,0 +1,51 @@
...
+  ~ for more details.
+  ~
+  ~ You should have received a copy of the GNU General Public License
+  ~ along with this program. If not, see https://www.gnu.org/licenses/.
+  -->
+
+<div class="verify-plus-scanner">
+  <div class="scanner-actions" *ngIf="!scanning && cameraPermission !== false && hasDevices !== false">
+    <tc-button [type]="'outline'" (onClick)="startScanning()">
+      Enable camera
```
EhsanEhrari: I understand this is the first PR for **Verify+** and that more changes or follow-up PRs may come later. One small point for this PR: since the candidate portal supports **translation**, these new user-facing strings should ideally be moved into the translation files instead of being hardcoded in the HTML. That will keep the UI consistent for non-English users and make the Verify+ flow ready for localization.

camerojo (changes requested):
I think Ehsan has picked up a couple of useful issues

ui/candidate-portal/src/app/components/common/verify-plus-scanner/verify-plus-scanner.component.html
sadatmalik: Thanks — that’s a helpful callout. The plan for this early PR is to defer translation work until the flow is a bit more settled, since the UX and copy are still likely to evolve as the next pieces land. It was already noted in Talent-Catalog/talentcatalog#3578, and I’ve now also opened Talent-Catalog/talentcatalog#3618 so it’s tracked explicitly and doesn’t get missed.

ui/candidate-portal/src/app/components/common/verify-plus-scanner/verify-plus-scanner.component.ts
sadatmalik: Good catch — you’re right. I’d left it open in this first cut mainly because it made repeated local testing easier without having to keep re-enabling the scanner, but re-enabling via the button is a better approach, so I’ll switch it to stop after a successful scan and just bring the button back with updated text, something like “Re-scan code” — untranslated for now :-).

sadatmalik:
Hi John, Ehsan — thanks both for the reviews. I’ve pushed a few changes, please take a look when you can.

Keen to get this into staging so I can continue testing on mobile.

camerojo (approved):

EhsanEhrari (approved):

PR #3595, opened 2026-07-07
John/tc 1058 need an easy way of adding our own skills

### LINKED ISSUE #2907: Need an easy way of adding our own skills
The ESCO and ONet skills are good to have but they are not enough.

We should be able to easily add our own skills to an extra TC Skills table - based on what we see coming in from Job Descriptions.

For example we have job descriptions containing acronyms for skills that exist in ONET - eg PLC and SCADA in the Scottish Power JDs - but they appear there with their full names in ONET rather than as the acronymns commonly used in job descriptions.
PLC = Programmable Logic Controller
SCADA = Supervisory control and data acquisition

camerojo:
Screenshot 2026-07-07 at 5 25 29 pm


sadatmalik (approved):
Great functionality here!

EhsanEhrari (approved):

PR #3606, opened 2026-07-08
Sadat/tc 1363 feature verify 21 backend ingest endpoint capture unhcr

sadatmalik:
This PR delivers the backend for processing Verify+ QR payloads --

- API and services to ingest a raw Verify+ QR payload.
- Raw payload is handled byte-exact - so a future signature check stays valid.
- UNHCR ID is parsed out from a mocked QR payload and stored on Candidate.unhcrNumber.
- Signals duplicates in parsed response

### DIFF
--- a/server/src/main/java/org/tctalent/server/api/portal/VerifyPlusPortalApi.java
+++ b/server/src/main/java/org/tctalent/server/api/portal/VerifyPlusPortalApi.java
@@ -1,0 +1,56 @@
+/*
+ * Copyright (c) 2026 Talent Catalog.
+ *
+ * This program is free software: you can redistribute it and/or modify it under
+ * the terms of the GNU General Public License as published by the Free
+ * Software Foundation, either version 3 of the License, or any later version.
+ *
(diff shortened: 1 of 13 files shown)


EhsanEhrari (approved):
Looks good to me overall! 👌
Minor non-blocking suggestion: since we only need a boolean duplicate check, the repository could use an **existsOtherByUnhcrNumber**... query instead of loading matching candidates into a list. Also test files do not seem to have the copyright header.

camerojo (approved):
Nice!

sadatmalik:
Thanks, Ehsan — good suggestions.

I’ll leave the repository method as-is for now, but I agree an existsOtherByUnhcrNumber query would be a good alternative for the current use case. My thinking is that retrieving the matching candidates could still be useful in a follow-up PR for the admin portal work, where we may want to surface and tag candidates with duplicate UNHCR numbers for admins to review.

And good catch on the missing copyright headers in the test files. Since this PR is approved, I’ll merge it and push those header updates in a separate PR.

PR #3608, opened 2026-07-08
Sadat/tc 1368 bug redis cache failure breaks jwt authentication durable

sadatmalik:
This PR fixes two Redis cache related problems:

- Implements a custom CacheErrorHandler to log cache operation failures and fall back to database
- Add custom candidate cache properties with configurable TTL for Redis entries on CandidateRedisCache

See the issue comments for further details.

### DIFF
--- a/server/src/main/java/org/tctalent/server/configuration/CacheConfig.java
+++ b/server/src/main/java/org/tctalent/server/configuration/CacheConfig.java
@@ -16,11 +16,54 @@
 
 package org.tctalent.server.configuration;
 
+import lombok.extern.slf4j.Slf4j;
+import org.springframework.cache.Cache;
+import org.springframework.cache.annotation.CachingConfigurer;
 import org.springframework.cache.annotation.EnableCaching;
+import org.springframework.cache.interceptor.CacheErrorHandler;
 import org.springframework.context.annotation.Configuration;
+import org.tctalent.server.logging.LogBuilder;
 
+@Slf4j
 @Configuration
 @EnableCaching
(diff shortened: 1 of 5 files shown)


EhsanEhrari (commented):
I think it looks good for fixing the login/authentication issue. If Redis cache fails, don’t crash the request. Just log the error and get the data from database.

My only concern is CandidateRedisCache. It uses Redis directly with StringRedisTemplate, so the new cache error handler may not catch errors there. This means some candidate/list/import code uses Redis directly, not through Spring cache. So, it may not protect those parts and CSV import or candidate list could still fail if Redis is down or full.

Could we also add try/catch inside CandidateRedisCache.multiGet() and putAll() so Redis failures are logged but the request continues using Postgres/recompute?

EhsanEhrari (approved):

camerojo (approved):

sadatmalik:
Thanks — I did note in the issue description that we currently have two different caching implementations in the repo. That’s why one was already picking up TTL from the generic config, while the other needed the custom config added in this PR.

We’re also aware the same distinction applies to cache error handling, as discussed in the team meeting. For this hotfix, though, we’re trying to keep the change as small and low-risk as possible: cache fall-through, plus TTL coverage for both caching implementations. TTL on the candidate cache should also help keep memory use under control.

Rather than adding more divergence in this PR, I’ll open a separate follow-up issue to converge on a single caching approach and remove the duplication. I’ll capture that, along with a few other Redis/cache-hardening items, in a small follow-up epic for prioritisation.

PR #3614, opened 2026-07-10
John/tc 1323 review cv creation xhtml template remove unnecessary

### LINKED ISSUE #3546: Review CV creation: xhtml template, remove unnecessary sanitization
## Summary

The CV template should be natively xhtml (rather than html). This eliminates the need for a html to xhtml conversion. Unnecessary sanitization increases complexity and also adds unnecessary updating to job descriptions on the database when, in fact, there have been no changes to the job description. 

During testing we also noted that generated PDF's do not support Arabic. They probably never have. But this seems to be fixable by using a different PDF creation library.

## Background

We have recently rewritten a lot of the CV generation code - this finishes off that work.

## Proposed Approach

See above

We should also 

## Risks

Low risk

## Sign-off Criteria

- [ ] Release notes updated (if applicable)
- [ ] Release planning updated (if applicable)
- [ ] Users notified (if applicable)
- [ ] Documentation updated

camerojo:
Simplify code by making template pure xhtml

Do some renaming to clarify cv generation from template, as distinct from generating pdfs from xhtml.


server/src/main/java/org/tctalent/server/service/db/util/PdfHelper.java
```diff
@@ -25,7 +26,6 @@
 import org.tctalent.server.exception.CvGenerationException;
 import org.tctalent.server.logging.LogBuilder;
 import org.tctalent.server.model.db.Candidate;
-import org.xhtmlrenderer.pdf.ITextRenderer;
 
 /**
  * Service for generating PDFs using Flying Saucer and Thymeleaf templates.
```
EhsanEhrari: The PdfHelper class documentation still mentions Flying Saucer, but the implementation now uses **OpenHTMLToPDF/PDFBox**.

server/build.gradle
```diff
@@ -162,20 +162,28 @@ dependencies {
     implementation "net.javacrumbs.shedlock:shedlock-spring:5.10.0"
     implementation "net.javacrumbs.shedlock:shedlock-provider-jdbc-template:5.10.0"
 
-    implementation("io.jsonwebtoken:jjwt-api:0.11.2")
+    //Note that there are breaking changes in jjwt 0.12 and 0.13. Fairly minor though.
+    //If you upgrade you will get some compile errors which should be easy to fix. JC
+    def jjwtVersion = "0.11.2"
+    dependencies {
```
EhsanEhrari: Nested dependencies {} block inside the existing one. Wouldn't' It be cleaner to remove the inner block?

EhsanEhrari (changes requested):

server/build.gradle
camerojo: I'm not sure. I liked that the version specification was close to the group of libraries that used that version - and that group of libraries who must share the same version are explicitly grouped. Before they were separate which caused a problem when they had different versions.
But maybe with version in a variable, the grouping is not necessary. 

What do you think @sadat

sadatmalik: My preference is to keep all versions (and any notes about them) in one place — i.e the ext {} block at the top of the file, alongside mapstructVersion, awsSdkVersion, etc. So I'd move jjwtVersion and the "breaking changes in 0.12/0.13" note up there:

```
ext {
  ...
  //Note there are breaking changes in jjwt 0.12 and 0.13 (minor - easy compile fixes). JC
  jjwtVersion = "0.11.2"
  ...
}
```

Then down in dependencies {} the three jjwt artifacts stay grouped as a plain block of adjacent lines (no nesting), which keeps the "these move together" intent:

```
// jjwt: api/impl/jackson must share the same version (jjwtVersion in ext)
implementation("io.jsonwebtoken:jjwt-api:$jjwtVersion")
runtimeOnly("io.jsonwebtoken:jjwt-impl:$jjwtVersion")
runtimeOnly("io.jsonwebtoken:jjwt-jackson:$jjwtVersion")
```

That way all versions and upgrade notes live in one spot, and dependent libraries are still visibly grouped.

EhsanEhrari (approved):

sadatmalik (approved):

Issue #3591, opened 2026-07-03
Remove Erase Candidate option from non sys admins, add delete option instead to change status to deleted


hiba-machfej:
<@U07M8NT2ASJ> Hanine should have received a better error message than "_*org.springframework.security.authorization.AuthorizationDeniedException: Access Denied*_."

Can you log that as an issue?

[Slack Message](https://refugeejobsmarket.slack.com/archives/C42S4AJ73/p1783033007093939?thread_ts=1781481878.770259&cid=C42S4AJ73)

cazcam34:
Spoke with John about this:

1. Bug in that only system admins should see this 'Erase candidate data' dropdown option. Currently admins (and maybe lower access roles?) can see the option. 
2. However the error message should be improved anyways, even though the dropdown option will be hidden when access is denied.
3. To avoid any issues with the team by removing this dropdown option add a new dropdown option 'Delete candidate' which just changes the status to deleted.

Hanine's concerns in the slack channel are that by not being able to do a full erasure herself will cause a backlog and more work needing to go to a system admin. She says it happens frequently that candidates request this, however full erasure is a new feature and until recently the status change to deleted was sufficient. So we want them to continue to use that method unless a candidate is very adamant and specific about wanting a full erasure, which we think is a more rare request. Replacing the erasure dropdown option with a delete dropdown option (doing the regular status = deleted) hopefully will keep everyone happy.

cazcam34:
How does the CandidateRepository.delete() method fit in with our delete strategies (status deleted and full erasure)? We have the code to delete there (candidate API delete and an unused delete modal from the front end) but the delete modal doesn't get used.

cazcam34:
Discussed the above with John - safe to delete this old code where a candidate is deleted via repository delete method. This kind of delete probably won't work anyways due to foreign keys etc. And this is no longer used as a delete method as we have the two other delete methods: status deleted and full erase.

cazcam34:
Bit of a bug found in the candidate status selector component. Basically sometimes we are passing in a candidate status key (e.g. from the candidate object, the candidate status is stored as a key in the databse) and sometimes we are passing in the candidate status string value (e.g. CandidateStatus.active). The ng select binds to the key value. It works because we are currently only passing in string values that match the key value (e.g. active = "active") but if we try to pass in deleted for example CandidateStatus.deleted key does not equal the string value exactly. So the select option appears blank instead of showing the deleted status option.

[...]

PR #3650, opened 2026-07-23
Caroline/tc 1398 fix candidate status key mapping in ng select

### LINKED ISSUE #3648: Fix candidate status key mapping in ng select
@Caroline Ritchie said in Talent-Catalog/talentcatalog#3591:

> Bit of a bug found in the candidate status selector component. Basically sometimes we are passing in a candidate status key (e.g. from the candidate object, the candidate status is stored as a key in the databse) and sometimes we are passing in the candidate status string value (e.g. CandidateStatus.active). The ng select binds to the key value. It works because we are currently only passing in string values that match the key value (e.g. active = "active") but if we try to pass in deleted for example CandidateStatus.deleted key does not equal the string value exactly. So the select option appears blank instead of showing the deleted status option.
>
> To solve this I have to make sure that the candidate status KEY is being used, so need to convert any string values to their associated key before passing into the ng select.

cazcam34:
Fixes bug with the candidate-status-selector where if an string value was passed in (e.g. CandidateStatus.deleted) it wouldn't work as the key and the string value are different. Previously has worked in the code where we pass in CandidateStatus.active, but thats only been because the key and string value of active status was both 'active'. Added a function where pass in an enum and a string value or a key, and return the appropriate key of an enum. This then can be used in the ng-select and returned to the DB.<br>Fixing this will allow me to implement issue Talent-Catalog/talentcatalog#3591 where I want to be able to add a dropdown called 'Delete Candidate' which opens the edit candidate status modal and prepopulates the status from the select dropdown with CandidateStatus.deleted. This wouldn't work with the way the code was previously done.


ui/admin-portal/src/app/components/util/candidate-status-selector/candidate-status-selector.component.spec.ts
```diff
@@ -54,7 +54,7 @@ describe('CandidateStatusSelectorComponent', () => {
 
   it('should filter out draft option from candidateStatusOptions', () => {
     const options: EnumOption[] = enumOptions(CandidateStatus);
-    const filteredOptions = options.filter(option => option.key !== CandidateStatus.draft);
+    const filteredOptions = options.filter(option => option.stringValue !== CandidateStatus.draft);
     expect(component.candidateStatusOptions).toEqual(filteredOptions);
```
camerojo, line 59: Why change to check stringValue? Surely the key is more reliable.

ui/admin-portal/src/app/components/util/candidate-status-selector/candidate-status-selector.component.ts
```diff
@@ -48,18 +40,18 @@ export class CandidateStatusSelectorComponent implements OnInit, OnChanges {
 
     //Filter out the draft option. Users should not be able to set a candidate's status to draft.
     this.candidateStatusOptions =
-      enumOptions(CandidateStatus).filter(option => option.key !== CandidateStatus.draft);
+      enumOptions(CandidateStatus).filter(option => option.stringValue !== CandidateStatus.draft);
```
camerojo: See above

camerojo (changes requested):

camerojo:
The build is also failing

ui/admin-portal/src/app/components/util/candidate-status-selector/candidate-status-selector.component.spec.ts
cazcam34: @camerojo because CandidateStatus.draft returns the string value, not the key. So it actually wasn't working, if you look in staging/prod the draft status still appears and doesn't get filtered out because CandidateStatus.draft returns "draft (inactive)" and the key is just draft.

camerojo (dismissed):

camerojo (approved):

PR #3664, opened 2026-07-29
John/tc 1380 define postgres database layout and support for hybrid

### LINKED ISSUE #3623: Define Postgres database layout and support for hybrid matching using text search and vector search

camerojo:


server/src/main/java/org/tctalent/server/service/db/impl/CandidateJobExperienceServiceImpl.java
```diff
@@ -170,6 +173,47 @@ public CandidateJobExperience updateCandidateJobExperience(Long id, UpdateJobExp
         return candidateJobExperience;
     }
 
+    @Override
+    public void updateCandidateJobExperienceEmbeddings(List<CandidateJobExperience> experiences) {
+
+        final String tableName = vectorEmbeddingModelProperties.getAlternateEmbeddingTable();
+        final String modelKey = vectorEmbeddingModelProperties.getEmbeddingModelKey();
```
sadatmalik, line 180: This writes to alternateEmbeddingTable but loads/generates with embeddingModelKey. For cutover those can diverge -- should this use alternateEmbeddingModelKey?

server/src/main/java/org/tctalent/server/service/embedding/dto/TcVectorEmbeddingServiceClientConfig.java
```diff
@@ -0,0 +1,29 @@
...
+@Configuration
+public class TcVectorEmbeddingServiceClientConfig {
+
+    @Bean
+    public TcVectorEmbeddingServiceClient tcVectorEmbeddingServiceClient(
+        RestClient.Builder restClientBuilder,
+        @Value("${tc-vector-embedding-service.apiUrl}") String apiUrl
+    ) {
+        RestClient restClient = restClientBuilder
+            .baseUrl(apiUrl)
```
sadatmalik, line 20: apiKey is configured in application.yml but isn't attached to this client. Skills extraction uses x-api-key on RestClient — should we be wiring the same here?

server/src/main/resources/db/migration/V2_20__create_embedding_tables.sql
```diff
@@ -0,0 +1,48 @@
+create table if not exists embedding_model (
+    id bigint GENERATED BY DEFAULT AS IDENTITY PRIMARY KEY,
```
sadatmalik: These tables use IDENTITY, but EmbeddingModel / JobExperienceEmbedding declare Hibernate SEQUENCE generators (i.e \*\_id_seq).

I don't know if there's a reason for doing it this way. But if not, can we align with the repo’s usual sequence/bigserial pattern so Flyway and JPA don’t diverge on sequence generators?

sadatmalik (changes requested):
Looks great John.

Left a few comments - I think the Flyways sequence generation may need aligning

server/src/main/java/org/tctalent/server/service/db/impl/CandidateJobExperienceServiceImpl.java
camerojo: I changed the approach including vectorEmbeddingModelProperties in the subsequent PR.
Basically I no longer have a primary table/model and alternate table/model. The revised approach simply has multiple tables/models.

You are right that this was a bug - nice catch! But it is addressed by the subsequent PR.

server/src/main/java/org/tctalent/server/service/embedding/dto/TcVectorEmbeddingServiceClientConfig.java
camerojo: Yes. I think you are right. There is work to be done on the security side of things.

server/src/main/resources/db/migration/V2_20__create_embedding_tables.sql
camerojo: This wasn't intentional - the AI generated it and it slipped in without me noticing it. Nice catch! 

However, it seems like IDENTITY is the way to go. BIGSERIAL is currently seen a tech debt. I will update this PR to be consistent and use BIGSERIAL but I will also raise a tech debt issue for looking moving to IDENTITY.

Here is the techdebt issue I created https://github.com/Talent-Catalog/talentcatalog/issues/3685

sadatmalik (approved):

PR #3690, opened 2026-08-09
Hiba/tc 723 Fix pasted email validation during candidate registration

### LINKED ISSUE #578: Cut-and-paste email address in candidate registraion can fail validation
From Hanine:

Please note that we're encountering an issue with multiple candidates attempting to register. They all receive the same message: _"This field must contain a valid email."_ I've personally verified one candidate's email, and it is indeed correct. However, despite the validity of the email, she is unable to complete the registration process using it.

[Slack Message](https://refugeejobsmarket.slack.com/archives/C42S4AJ73/p1705925815288099?thread_ts=1705925815.288099&cid=C42S4AJ73)

On further investigation the issue is caused by cut-and-paste of the email address into the registration field. Enhancing the regex validation would be useful to detect for such invisible characters, and to remove them.

hiba-machfej:
The main issue occurred when candidates copied and pasted their email addresses into the registration form. In some cases, the pasted value contained invisible Unicode characters, causing an otherwise valid email address to be rejected.

To address this properly, email sanitization was implemented in both the frontend and backend.
## Frontend

Frontend sanitization improves the user experience and directly addresses the reported copy-and-paste issue.

1- A reusable email-sanitization function is added to remove leading and trailing whitespace and invisible Unicode characters from pasted email addresses.
2-A regular expression is necessary because some invisible characters are not removed by standard whitespace-handling methods. The following character pattern is used:
```
const INVISIBLE_OR_WHITESPACE_CHAR =
  '[\\s\\p{Zs}\\p{Cf}\\u115F\\u1160\\u17B4\\u17B5\\u3164\\uFFA0]';
```

`\s —` whitespace characters, including spaces, tabs, and line breaks.
`\p{Zs}` — Unicode space-separator characters, including non-breaking, thin, and ideographic spaces.
`\p{Cf} `— Unicode formatting characters, including zero-width and bidirectional-control characters.
`\u115F and \u1160 `— Hangul filler characters that can appear as invisible blanks.
`\u17B4 and \u17B5` — invisible Khmer vowel characters.
`\u3164 and \uFFA0 `— additional Hangul filler characters that can appear as invisible blanks.

The sanitizer removes these characters only from the beginning and end of the email address, preserving its internal content so that genuinely invalid email addresses still fail validation.

## Backend
The same sanitization is implemented on the backend to protect data integrity and ensure consistent validation regardless of where a request originates. This is important because email addresses may also be submitted through future API integrations, mobile applications, or other clients that do not use the frontend sanitizer.

Neither Java’s `trim()` nor `strip()` is sufficient without the additional regular expression. 
Java’s `trim()` only removes leading and trailing characters with Unicode values less than or equal to `U+0020`. Therefore, it does not handle many Unicode spaces or invisible formatting characters.

Java’s `strip() `provides broader Unicode support by using `Character.isWhitespace()`, but it still does not recognize every invisible or blank-looking character covered by the pattern.

The additional regular expression is therefore necessary to remove the characters responsible for the reported copy-and-paste issue.

### DIFF
--- a/ui/candidate-portal/src/app/model/base.ts
+++ b/ui/candidate-portal/src/app/model/base.ts
@@ -281,27 +281,10 @@ export const EMAIL_REGEX: string =
   '(?!.*[@.]{2})[a-zA-Z0-9!#$%&\'*+-/=?^_`{|}~]+[a-zA-Z0-9.!#$%&\'*+-/=?^_`{|}~]*@(?!-)[a-zA-Z0-9-]+(?<!-)(\\.(?!-)[a-zA-Z0-9-]+(?<!-))*$';
 
 /**
- * Matches one invisible/blank Unicode character: whitespace, plus the Hangul/Khmer filler
- * characters that render blank but aren't classified as whitespace.
- *
- * Mirrors INVISIBLE_OR_WHITESPACE_CHAR in
- * server/.../request/candidate/BaseCandidateContactRequest.java - keep both in sync.
- */
-const INVISIBLE_OR_WHITESPACE_CHAR = '[\\s\\p{Zs}\\p{Cf}\\u115F\\u1160\\u17B4\\u17B5\\u3164\\uFFA0]';
-
-/**
- * Strips invisible/blank characters from the start and end of a pasted email address - see
- * TC-723. Leaves anything in the middle untouched, since that would mean the email is
- * genuinely malformed rather than just messily pasted.
+ * Trims leading/trailing whitespace from a pasted email address - see TC-723.
  */
 export function sanitizeEmailInput(value: string): string {
-  if (!value) {
-    return value;
-  }
-  const edgeRegex = new RegExp(
-    `^${INVISIBLE_OR_WHITESPACE_CHAR}+|${INVISIBLE_OR_WHITESPACE_CHAR}+$`, 'gu'
-  );
-  return value.replace(edgeRegex, '');
+  return value == null ? value : value.trim();
 }
 
 /**
--- a/ui/candidate-portal/src/app/model/base.ts
+++ b/ui/candidate-portal/src/app/model/base.ts
@@ -281,27 +281,25 @@ export const EMAIL_REGEX: string =
   '(?!.*[@.]{2})[a-zA-Z0-9!#$%&\'*+-/=?^_`{|}~]+[a-zA-Z0-9.!#$%&\'*+-/=?^_`{|}~]*@(?!-)[a-zA-Z0-9-]+(?<!-)(\\.(?!-)[a-zA-Z0-9-]+(?<!-))*$';
 
 /**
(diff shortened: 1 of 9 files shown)


camerojo (changes requested):
Did you consider whether this sanitization would be better on the Java server? What are the arguments for doing it on the front end?

ui/candidate-portal/src/app/model/base.ts
```diff
@@ -280,6 +280,32 @@ export class LoginRequest {
 export const EMAIL_REGEX: string =
   '(?!.*[@.]{2})[a-zA-Z0-9!#$%&\'*+-/=?^_`{|}~]+[a-zA-Z0-9.!#$%&\'*+-/=?^_`{|}~]*@(?!-)[a-zA-Z0-9-]+(?<!-)(\\.(?!-)[a-zA-Z0-9-]+(?<!-))*$';
 
+/**
+ * Matches a single invisible/blank Unicode character (eg zero-width space, non-breaking space,
+ * BOM, Hangul filler character) or ordinary whitespace character.
+ */
+const INVISIBLE_OR_WHITESPACE_CHAR =
```
camerojo, line 287: Java has good standard built in support for this kind of sanitiztion as we discussed in  out call - eg with the String .strip method.

Arguably that would be one argument for doing the sanitization on the server rather than adding this fairly complex code on the front end.

camerojo (changes requested):
Consider whether this functionality is best on the back end or front end.

ui/candidate-portal/src/app/model/base.ts
hiba-machfej: Yes, I actually left a comment in the issue that the white space in the example we had is HANGUL FILLER/U+3164 won't be removed by strip() because this U+3164 is classified as a letter, not whitespace. 

According to the Java String.strip() removes only characters recognized by [Character.isWhitespace(int).](https://docs.oracle.com/en/java/javase/25/docs/api/java.base/java/lang/Character.html#isWhitespace(int))
 Although it handles Java-defined whitespace, it does not handle the relevant invisible characters — which may be mistakenly included when candidates copy and paste their email addresses, as described in the issue.  Like, hangul filler, zero-width space, BOM, and non-breaking space  Implementing this on the backend would still require explicit sanitization similar to the frontend utility; it would not be a one-line call to strip().
 
I considered whether this belongs on the frontend or backend. The frontend is necessary for this specific issue because pattern validation prevented the candidates from continuing when they copy pasted one of these characters submitted. Backend-only would not resolve the candidate’s immediate registration experience. So, for this focused issue, I placed the complexity in the frontend side. But backend sanitization ensures data integrity and can be useful for future mobile application.

EhsanEhrari (approved):
Good job @hiba-machfej  approved ✅ One small question: Do we still need the comments about stripping invisible characters?

camerojo (approved):

PR #3697, opened 2026-08-12
Add "Search candidates in list" icon to List displays - allowing lists to be searched.

### LINKED ISSUE #3044: Submission List filtering and field visibility
The below requirement is noted verbatim from Hanaa's Recruitment Challenges definition doc for further refinement.

In the Submission List, for instance with a hiring commitment of 20, the list comprises at least 100 candidates; in recent case there is an example of 700 candidates in a list; there are no filters for fields. We cannot filter by job stage, partner, or any other field (for example: status, IELTS score, etc.).

Not all fields are available in the Submission List, which reduces visibility and makes decision-making in TC not practical for larger cohorts.

The recruitment team needs all candidate application fields to be available, and they should be able to choose which fields are visible.

### Sign-off Criteria

- [x] Release notes updated
- [x] Release planning updated
- [x] Users notified
- [x] Documentation updated

camerojo:
This PR also removed some legacy code that no longer serves any purpose - plus a check related to checking sorting fields that would never work (my bad code!).

### DIFF
--- a/ui/admin-portal/src/app/components/candidates/candidates-search/candidates-search.component.html
+++ b/ui/admin-portal/src/app/components/candidates/candidates-search/candidates-search.component.html
@@ -22,6 +22,7 @@
 
 <app-define-search *ngIf="!error && !loading"
                      [jobId]="jobId"
+                     [listId]="listId"
                      [savedSearch]="savedSearch"
                      [pageNumber]="pageNumber"
                      [pageSize]="pageSize"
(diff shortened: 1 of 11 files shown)


sadatmalik:
Looks great - just the code-cov report to look at above.

camerojo:
> Looks great - just the code-cov report to look at above.

Wow - nice! Extra unit tests included.

PR #3717, opened 2026-08-25
Build embeddings should skip records which already have an embedding

### LINKED ISSUE #3715: Build embeddings should skip records which already have an embedding.
## Overview

When building embeddings for a new model, there is not need to rebuild an embedding that is alreday there.

## Problem / Opportunity

This allows us to restart a build embedding batch which terminated for some reason, without needing to start right from the beginning again. We can continue from where we left off.

## Proposed Solution

The search that drives the batch should exclude records which already have embeddings.

In Batch can just page through experiences with no embeddings. See ChatGPT|Spring for pageable code not using repository (because we need to construct the search SQL using the computed embedding table name).

## Acceptance Criteria

- [ ] Should be able to restart a server in the middle of batch run. When the server restarts it can continue from where it left off instead of starting all over from the beginning

## Sign-off Criteria

- [ ] Release notes updated
- [ ] Release planning updated
- [ ] Users notified
- [ ] Documentation updated

camerojo:


server/src/main/java/org/tctalent/server/service/db/impl/CandidateJobExperienceServiceImpl.java
```diff
@@ -213,47 +219,76 @@ private void batchUpdateCandidateJobExperienceEmbeddings(
...
+        List<EmbeddingInput> embeddingInputs = new ArrayList<>();
         experiences.forEach(experience -> {
-            final String id = experience.getId().toString();
-            EmbeddingInput input = EmbeddingInput.builder()
-                .id(id)
-                .context(computeExperienceContext(experience))
-                .text(experience.getDescription())
-                .build();
-            descriptions.add(input);
+            //Skip computing embeddings if they already have them.
```
EhsanEhrari, line 221: @camerojo , I think there may be an **N+1** query issue here.
We loop through every experience and call `hasEmbeddedVector(experience, model)` for each one. That method then runs a separate `SELECT EXISTS` query against the embedding table.

Since `buildEmbeddings()` processes 100 experiences per page, this could mean around 100 extra database queries for every page just to check which records already have embeddings.

Could we instead fetch all existing embedding IDs for the whole page in one query, then filter them in Java? That would reduce it from roughly 100 DB queries per page to 1.

camerojo: Thanks for this @EhsanEhrari . I did realize that it was a little inefficient but didn't worry too much because it is part of a long batch anyway. But you are right - a simple change makes it better. See changes.

EhsanEhrari (dismissed):

hiba-machfej (dismissed):

server/src/main/java/org/tctalent/server/service/db/impl/CandidateJobExperienceServiceImpl.java
```diff
@@ -61,6 +66,9 @@
 @Slf4j
 public class CandidateJobExperienceServiceImpl implements CandidateJobExperienceService {
 
+    @PersistenceContext
+    private EntityManager entityManager;
```
sadatmalik: > Prefer not injecting EntityManager here? 
>
> Think we can use JobExperienceEmbeddingRepository which can support the native SQL and dynamic tables - see below. . .

server/src/main/java/org/tctalent/server/service/db/impl/CandidateJobExperienceServiceImpl.java
```diff
@@ -213,47 +221,100 @@ private void batchUpdateCandidateJobExperienceEmbeddings(
...
         }
     }
 
+    /**
+     * Checks which of the given job experiences have embeddings for the given model.
+     * @param experiences Job experiences
+     * @param model Embedding model
+     * @return Set of ids of experiences that have embeddings for the given model.
+     */
+    private Set<Long> checkForEmbeddedVectors(List<CandidateJobExperience> experiences, EmbeddingModel model) {
```
sadatmalik, line 272: I think this belongs on JobExperienceEmbeddingRepository like upsert (same JdbcTemplate + validateTableName). Then this service can drop EntityManager and the duplicated table-name check?

sadatmalik:
Just left a couple of minor review comments @camerojo

server/src/main/java/org/tctalent/server/service/db/impl/CandidateJobExperienceServiceImpl.java
camerojo: @sadatmalik Good catch - thanks.

PR #3781, opened 2026-09-08
Sadat/tc 1475 perf smoke fails on candidate search due to invalid

sadatmalik:


sadatmalik:
Merging this because it fixes an issue in staging - all other pushes will fail until this is pushed.

PR #3811, opened 2026-09-17
ehsan/tc-1498-saved-search-csv-export-ignores-unsaved-filter-changes

EhsanEhrari:
## Summary

Fixes saved search CSV exports so that they use the filters currently applied on screen instead of always using the previously saved search criteria.

Previously, a user could modify filters such as Country, Status, Gender, Occupation, or Nationality and click **Search/Apply** without updating the saved search. The displayed candidate results would correctly reflect those changes, but exporting the CSV would still use the persisted saved search filters.

## Expected Behaviour

When a user:

1. Opens a saved search.
2. Changes one or more filters.
3. Clicks **Search/Apply** without clicking **Update Search**.
4. Exports the candidates.

The generated CSV now matches the candidate results currently displayed on screen.


camerojo (commented):
This is a good fix - but there is a more radical fix that I have been thinking about for a long time. This might be the time to do that more radical fix.

Currently this is acting as designed. Saved searches only update when someone clicks on the Update Search button. So a user can put all sorts of changes into a saved search, but they will all be lost if they forget to click on Update Search.

Back end operations - like creating a CSV file - are designed to run from the search as saved on the back end.

Another model which I think is simpler - and also makes the code simpler - is that a saved search should automatically update when a search is run. In fact this is how an "unsaved" search operates. It actually auto saves to the user's default search whenever it is run. This allows users to come back to where they left off when working with unsaved searches.

The fact that unsaved searches and saved searches work differently can be confusing - and it is behind this bug.

A simple way to change this behaviour without losing the capacity to reinstate it if we need to would be to introduce an "autoUpdateOnSearch" flag which is associated with a search and defaults to true.

PR #3731, opened 2026-08-27
Make terms and mail templates well-formed XHTML

devamchakrabarty:
## Summary

- add the XHTML namespace and XML declarations required by the terms template standard
- self-close void elements and repair invalid paragraph nesting in terms and mail templates
- add a regression test that parses every terms/mail template and verifies its XHTML namespace

## Modified components

- `server/src/main/resources/terms`
- `server/src/main/resources/mail`
- `HtmlTemplateWellFormednessTest`

## Testing

- `./gradlew server:build -x test` — passed
- focused template, email, agreement, and services controller tests — passed
- `xmllint --noout` across all 16 terms/mail templates — passed
- `./gradlew :performance-tests:compileGatlingJava --rerun-tasks` — passed
- full `./gradlew clean :server:test --stacktrace --rerun-tasks` — 3,192 of 3,203 tests passed; the remaining 11 database integration classes could not initialize in the local environment. The CI-only database dump is private and unavailable locally, so those classes remain for the repository CI build to verify.

## PR check notes

- AWS Amplify preview build — passed on rerun
- Gatling smoke check — the performance sources compile and the simulation starts, but fork PRs receive an empty `USERS_CSV`; the workflow skips creating `users.csv` and Gatling then reports that the feeder file is missing
- preview-link check — the workflow found stale broken Amplify comments and failed with HTTP 403 while attempting to delete a bot-owned comment

No release notes, release-planning changes, or user notification are needed because this is a markup-authoring consistency change with no intended runtime behavior change.

Closes #3627

PR #3817, opened 2026-09-17
2342 Enforce ownership check when deleting candidate certification

### LINKED ISSUE #2342: Bug: Candidate Certification delete method not verifying as intended
Discovered while testing `CandidateCertificationServiceImpl.deleteCandidateCertification()`. Not a showstopper given that candidates don't have a UI option to delete certifications belonging to others. Nevertheless, the method contains this tautological condition check that should probably be fixed:

```
        Candidate candidate = candidateCertification.getCandidate();

        // Check that the user is deleting their own candidate certification
        if (!candidate.getId().equals(candidateCertification.getCandidate().getId())) {
            throw new InvalidCredentialsException("You do not have permission to perform that action");
        }
```

devamchakrabarty:
## Summary

Closes #2342.

`CandidateCertificationServiceImpl.deleteCandidateCertification()` compared the certification's candidate id against itself, so the ownership check could never fail. This replaces it with the admin versus owner check already used by the sibling services.

## What changed

- Replaced the tautological `candidate.getId().equals(candidateCertification.getCandidate().getId())` condition with a real check: users with admin privileges may delete any candidate's certification, while candidate portal users may only delete their own.
- Followed the same structure as `CandidateOccupationServiceImpl.deleteCandidateOccupation()` and `CandidateJobExperienceServiceImpl.deleteCandidateJobExperience()`, so the three delete paths now behave consistently.
- Reused the candidate already loaded by `findByIdLoadCandidate` for the id comparison and the subsequent `candidateService.save()`, rather than re-fetching it through `candidateRepository`, so the fix adds no extra query.
- Added Javadoc to `CandidateCertificationService.deleteCandidateCertification()` recording the permission rule and the exceptions it can throw.

## Why

The issue notes that candidates have no UI option to delete another candidate's certification. However `DELETE /api/portal/candidate-certification/{id}` takes the certification id straight from the path and the service performed no effective check, so any logged in candidate could delete another candidate's certification by iterating ids. Both `CandidateCertificationPortalApi` and `CandidateCertificationAdminApi` call the same service method, which is why the fix distinguishes admin users from candidates rather than simply rejecting mismatches.

Admin behaviour is unchanged. Candidates deleting their own certification are unchanged.

## Validation

- `JAVA_HOME=/opt/homebrew/opt/openjdk@17 ./gradlew :server:test --tests org.tctalent.server.service.db.impl.CandidateCertificationServiceImplTest --tests org.tctalent.server.api.portal.CandidateCertificationPortalApiTest --tests org.tctalent.server.api.admin.CandidateCertificationAdminApiTest`
  - Passed: 13 tests in `CandidateCertificationServiceImplTest`, 0 failures, plus both API test classes.
- Confirmed the new tests genuinely cover the bug: with only `CandidateCertificationServiceImpl` reverted to the previous version, 4 of the 13 tests fail, including "should throw when candidate deletes another candidate's certification".
- Added lines are within the 100 column Google Java Style limit.

## Tests added or updated

Added to `CandidateCertificationServiceImplTest`:

- `deleteCandidateCertification_shouldDeleteOwnCertification_whenCandidateNotAdmin`
- `deleteCandidateCertification_shouldThrow_whenDeletingAnotherCandidatesCertification`
- `deleteCandidateCertification_shouldThrow_whenNonAdminUserHasNoCandidate`

Updated:

- `deleteCandidateCertification_shouldDeleteCandidateCertification` now stubs `authService.hasAdminPrivileges()` and is named for the admin path. No existing tests were removed.

## Modified components

- `server/src/main/java/org/tctalent/server/service/db/impl/CandidateCertificationServiceImpl.java`
- `server/src/main/java/org/tctalent/server/service/db/CandidateCertificationService.java`
- `server/src/test/java/org/tctalent/server/service/db/impl/CandidateCertificationServiceImplTest.java`

## Out of scope

Two related things were left alone to keep this change small, and are worth separate issues if the team agrees:

1. `updateCandidateCertification()` performs no ownership check at all, so `PUT /api/portal/candidate-certification` has the same exposure. The sibling update methods (`updateCandidateJobExperience()` and others) share that gap, so it looks systemic rather than specific to certifications.
2. `candidateRepository` is injected into `CandidateCertificationServiceImpl` but is unused. That predates this change and removing it would alter the constructor signature.

## Contribution guideline alignment

- Branched from current upstream `staging` and uses the fork and pull workflow.
- Branch name is issue linked and descriptive.
- Scope limited to issue #2342, with no unrelated changes.
- No database or Flyway changes.
- Follows Google Java Style and adds Javadoc for the changed interface method.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


devamchakrabarty:
The two failing checks on this PR come from the CI environment, not from this change. Neither one builds or runs the server code.

- **`smoke / gatling`** failed after 13 seconds while pulling its base image, because Docker Hub returned a `502 Bad Gateway` from `auth.docker.io`. A re-run would probably still fail: the job needs the `STAGING_PERF_USERS_CSV` secret, and GitHub does not pass secrets to pull requests from forks. On my earlier PR #3709 the image pull worked, and the job then failed with `USERS_CSV not provided; skipping users.csv`.
- **`check-preview-link`** fails with `403 Resource not accessible by integration` when it tries to delete the Amplify preview comments, because workflows triggered by fork PRs get a read-only `GITHUB_TOKEN`.

Both checks fail the same way on my earlier fork PRs, #3709 and #3731.

The server unit tests have not run in CI for this PR either. `tc-build.yml` only triggers on `push`, so it ran on my fork instead, and stopped at "Configure AWS credentials" before reaching the test step. Locally, `CandidateCertificationServiceImplTest` passes 13 of 13, and both certification API test classes pass. Reverting only `CandidateCertificationServiceImpl` makes 4 of the 13 fail. The exact command is in the Validation section of the PR description.

PR #3821, opened 2026-09-21
Ehsan/tc 1492 feature flush redis caches on aws after build and deploy

EhsanEhrari:
This PR updates the staging deployment workflow so Redis/ElastiCache is automatically rebooted after deployments, preventing stale cached data from surviving a staging release.

The implementation keeps the workflow small by moving the shared Redis reboot logic into a reusable local GitHub composite action.

## Changes

- Added reusable GitHub Action:
  - `.github/actions/reboot-elasticache/action.yml`

- Added automatic Redis reboot for:
  - TBB staging
  - OPC staging
  - GRN staging

- OPC and GRN now wait for ECS deployment to become stable before rebooting Redis.

- TBB Redis is rebooted after the Jib deployment, but the unused TBB ECS service is not force-restarted.

- Redis primary cluster/node are discovered dynamically from the replication group.

- Workflow fails if:
  - the Redis primary cannot be found
  - the reboot command fails
  - ElastiCache does not return to an available state

- Production deployment workflow remains unchanged; production Redis continues to be handled manually.

### TBB

```text
Jib deploy
→ Reboot Redis
→ Wait for Redis available
```

### OPC

```text
Jib deploy
→ Force ECS deployment
→ Wait for ECS stable
→ Reboot Redis
→ Wait for Redis available
```

### GRN

```text
Jib deploy
→ Force ECS deployment
→ Wait for ECS stable
→ Reboot Redis
→ Wait for Redis available
```

## Configuration

TBB, OPC and GRN use their known staging replication groups:

```text
TBB: tc-test-cache
OPC: tc-test-cache
GRN: grn-staging-cache
```

The existing GitHub Actions AWS credentials are reused; no Redis credentials are required.

Required AWS permissions for the deployment identities are:

```text
elasticache:DescribeReplicationGroups
elasticache:DescribeCacheClusters
elasticache:RebootCacheCluster
```


sadatmalik (approved):
Nice use of reusable actions Ehsan - good that you introduced that.

I'm approving as it's staging only — but I think the workflow will fail because the IAM user used by the action may not have the required Redis permissions. If so, I can take a look, or you can.

